PR #3, opened 2019-08-27
Update StorageImpl.java

Sasha192:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,15 +2,116 @@
...
+            Entry entry = table[index];
+            if (entry.getKey().equals(key)) {
+                entry.setValue(value);
+            }
+        }
+        table[size] = new Entry<K, V>(key, value);
+        size++;
+    }
+
+    public boolean isExit(K key) {
```
boroda4436: @Sasha192 is it correct method name?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,15 +2,116 @@
...
+            Entry entry = table[i];
+            if (entry.getKey().equals(key)) {
+                return (V) entry.getValue();
+            }
+        }
         return null;
     }
-}
 
+    public V getForNullKey() {
```
boroda4436: @Sasha192 do you need this method? Why we can't use `public V get(K key) ` ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,15 +2,116 @@
...
+        for (int i = 0; i < size; i++) {
+            Entry entry = table[i];
+            if (entry.getKey() == null) {
+                return (V) entry.getValue();
+            }
+        }
+        return null;
+    }
+
+    private static class Entry<K, V> {
```
boroda4436: @Sasha192 should this class be static?

PR #5, opened 2019-08-27
added solution for storage homework

Tarabarnik:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,16 +1,53 @@
...
 public class StorageImpl<K, V> implements Storage<K,V> {
 
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+    private int capacity;
+
+    public StorageImpl() {
+        size = 0;
+        capacity = 8;
```
Sofasmile: maybe better make it as constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,16 +1,53 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (capacity == size) {
```
Sofasmile: better move this logic in the separate method
and read about Arrays.copyOf()

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,16 +1,56 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key) {
+                return (V) values[i];
+            }
+        }
+        return  null;
+    }
+
+    private void increaseCapacity() {
+        capacity += BASIC_CAPACITY;
```
Sofasmile: maybe better capacity *2 ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,16 +1,56 @@
...
+        size = BASIC_SIZE;
+        capacity = BASIC_CAPACITY;
+        keys = new Object[capacity];
+        values = new Object[capacity];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key) {
```
Sofasmile: how will work this method if key is object?
what is the difference between equals and ==?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,16 +1,56 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        if (size == 0) {
+            return null;
+        }
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key) {
```
Sofasmile: the same

src/main/java/core/basesyntax/impl/StorageImpl.java
Tarabarnik: == compares references, equals compares instances.

Tarabarnik: If key is Object, values in array should be Objects, too, otherwise equals will be 100% false, isn't it?

Tarabarnik: And equals doesn't like 'null's, of course  : -)

Sofasmile: cool :)

PR #11, opened 2019-08-27
Solution added

kmety:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,54 @@
...
+                    return;
+                }
+            } else {
+                if (key.equals((K) keyArray[i])) {
+                    valueArray[i] = value;
+                    return;
+                }
+            }
+        }
+        if (size == keyArray.length) {
```
Sofasmile: maybe better move this checking in the separate private method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,58 @@
...
-    @Override
     public void put(K key, V value) {
+        if (size == 0) {
+            keyArray[size] = key;
+            valueArray[size] = value;
+            size++;
+            return;
+        }
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
```
Sofasmile: we can simplify:
if (key == null || key.equals((K) keyArray[i])){
valueArray[i] = value;
                    return;
}

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,58 @@
...
+        if (size == keyArray.length) {
+            keyArray = Arrays.copyOf(keyArray, keyArray.length * 2);
+            valueArray = Arrays.copyOf(valueArray, valueArray.length * 2);
+        }
     }
 
-    @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
```
Sofasmile: the same)

PR #47, opened 2019-12-10
added StorageImpl

YarikJ:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,31 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keyArr = new Object[10];
```
Tarabarnik: create a constant 10 as a default capacity

Tarabarnik: put the initialization of arrays to constructor, please

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,31 @@
...
+    private Object[] valueArr = new Object[10];
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArr.length; i++) {
```
Tarabarnik: suggestion: you may add a field called `size` which stores amount of objects, so you won't need to search all array for key

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,31 @@
...
+            if (key == keyArr[i]) {
```
Tarabarnik: what's the difference between == and equals?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,31 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArr.length; i++) {
+            if (key != null && key.equals((keyArr[i]))) {
+                return (V) valueArr[i];
+            } else if (keyArr[i] == null) {
```
Tarabarnik: I think, this check is redundant. try to solve without it

PR #52, opened 2019-12-10
6.2 is done

dmytro-hryhoruk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,45 @@
...
+    private Object[] keys;
+    private Object[] values;
+
+    public StorageImpl() {
+        this.keys = new Object[ARRAY_CAPACITY];
+        this.values = new Object[ARRAY_CAPACITY];
+    }
 
     @Override
     public void put(K key, V value) {
```
Tarabarnik: too complicated. some checks are repeating. try to simplify

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,45 @@
...
+            if (key == null && keys[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+            }
+            System.out.println(values[i]);
+        }
     }
 
     @Override
     public V get(K key) {
```
Tarabarnik: same about this one

PR #53, opened 2019-12-10
Added solution for hw 6.2

ElvenNurse:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int CAPACITY = 10;
```
```suggestion
    private static final int CAPACITY = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+        keys = new Object[CAPACITY];
+        values = new Object[CAPACITY];
+        itemsCount = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < itemsCount; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
```
Tarabarnik: should we add break; here?

PR #54, opened 2019-12-10
hw6.2

Mykhaylo12:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,18 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    HashMap<K, V> storage = new HashMap<>();
```
Tarabarnik: we shouldn't use hashmap for this task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,28 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_SIZE = 10;
+    private Object[] keyStorage = new Object[ARRAY_SIZE];
+    private Object[] valueStorage = new Object[ARRAY_SIZE];
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < valueStorage.length; i++) {
+            if (valueStorage[i] == null & keyStorage[i] == null | keyStorage[i] == key) {
```
Tarabarnik: values == null are allowed and keys should be compared by equals so this if should be fixed :)

src/main/java/core/basesyntax/impl/StorageImpl.java
Mykhaylo12: ok

src/main/java/core/basesyntax/impl/StorageImpl.java
Mykhaylo12: ok

PR #56, opened 2019-12-10
added hw 6.2 solution

programmer28:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private static Object[] arrayKeys = new Object[CAPACITY];
```
Tarabarnik: fields should be initialized at constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+    private static Object[] arrayValues = new Object[CAPACITY];
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < arrayKeys.length; i++) {
```
Tarabarnik: this method contains too much checks and returns. try to create a field called size to limit your loop iterations and try to remove redundant checks

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < arrayKeys.length; i++) {
+            if (key == null && arrayKeys[i] == null) {
+                return (V)arrayValues[i];
+            } else if (key != null && key.equals(arrayKeys[i])) {
```
Tarabarnik: there is a check for key == null and key != null, so one of them is useless :)

programmer28: I`ve changed it

src/main/java/core/basesyntax/impl/StorageImpl.java
programmer28: Okay, I`ve made it

src/main/java/core/basesyntax/impl/StorageImpl.java
programmer28: I`ve changed that by this way

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private int size;
+    private Object[] arrayKeys;
+    private Object[] arrayValues;
+
+    public StorageImpl() {
+        this.size = 0;
```
Tarabarnik: lets remove all `these` keyws

programmer28: Sorry, but I don`t understand for why I must remove that. You`ve recommend me to create a counter named size and initialized all field in constructor. What have I made wrong?

programmer28: Okay, I`ve fixed that

PR #57, opened 2019-12-10
hw:6.2 - Storage solved

Danhov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys = (K[]) new Object[10];
```
Tarabarnik: any magic number should be a constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+    private V[] values = (V[]) new Object[10];
+    private int index = 0;
```
Tarabarnik: fields should be initialized at constructor

PR #58, opened 2019-12-10
added storage solution

TarabanDmitry:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,30 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private final int arrayCapacity = 10;
+    private Object[] arrayOfKeys = new Object[arrayCapacity];
+    private Object[] arrayOfValues = new Object[arrayCapacity];
```
Tarabarnik: fields initialization should be provided at constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,30 @@
...
+
+    int size = 1;
```
Tarabarnik: add access modifier, and why is it 1? you don't have any objects at arrays from start, do you?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,30 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private final int arrayCapacity = 10;
```
Tarabarnik: what are the rules of proper constant initialization and naming?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,30 @@
...
+    private Object[] arrayOfKeys = new Object[arrayCapacity];
+    private Object[] arrayOfValues = new Object[arrayCapacity];
+
+    int size = 1;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < arrayCapacity - size; i++) {
```
Tarabarnik: how much objects should add this method? how much it does now?

PR #60, opened 2019-12-10
hm6.2 v1.

Levkosvd:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxSize = 10;
```
Tarabarnik: is it a constant?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
...
+    Object[] keys =  new Object[maxSize];
```
```suggestion
    Object[] keys = new Object[maxSize];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
...
+    Object[] values = new Object[maxSize];
+    private int indexCounter = 0;
+    private int nullIndex;
```
Tarabarnik: this field is redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,39 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    Object[] keys = new Object[MAX_SIZE];
+    Object[] values = new Object[MAX_SIZE];
+    private int indexCounter = 0;
 
     @Override
     public void put(K key, V value) {
+        if (key == null) {
```
Tarabarnik: try to add two values with key == null. they should be written to the same spot

PR #61, opened 2019-12-11
HW 6.2 - solution

Julianevinska:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,38 @@
...
+
+    public StorageImpl() {
+        keyArr = new Object[MAX_SIZE];
+        valueArr = new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArr.length; i++) {
+            if ((key == null && keyArr[i] == null)
```
Tarabarnik: this method may be simplified :)

PR #63, opened 2019-12-11
home work 6.2

graveman13:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+            for (int i = 0; i < size; i++) {
+                if (arrayOfKey[i] == key || arrayOfKey[i].equals(key)) {
+                    return (V) arrayOfValue[i];
+                }
+            }
+        }
         return null;
     }
+
+    private int getIdxTheSameKey(K key) {
```
Tarabarnik: suggested name: indexOf
this method should return -1, if key does not present

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+        if (size > 0) {
+            for (int i = 0; i < size; i++) {
+                if (arrayOfKey[i] == key) {
+                    return i;
+                }
+            }
+        }
+        return size++;
```
Tarabarnik: size++ is a job for put method :)

PR #68, opened 2019-12-11
added hw-6-2-solution

AlinaPodoliaka:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,38 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private Object[] arrayOfKeys;
+    private Object[] arrayOfValues;
+    private int size = 10;
```
Tarabarnik: make this field constant, please

PR #70, opened 2019-12-11
Done

IegorSoboliev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
```
Tarabarnik: this key and value are never used in your code. please, delete them :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+    private static final int DEFAULT_ARRAY_LENGTH = 10;
+    private Object[] keyArray = new Object[DEFAULT_ARRAY_LENGTH];
+    private Object[] valueArray = new Object[DEFAULT_ARRAY_LENGTH];
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < indexCounter; i++) {
+            if (key == null || key.equals(keyArray[i])) {
+                valueArray[i] = value;
+                //I was thinking it should be - valueArray[indexCounter] = value;. Please, help:)!
```
Tarabarnik: this is a good solution, but I wanted you to create a storage, that may contain values with null key. it's not necessary, so good job! :) and please, remove this comment

PR #72, opened 2019-12-12
Added solution homework_6.2

blakh706:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static int MAX_SIZE = 10;
+    private int counter = 0;
+    private K key;
```
Tarabarnik: this key and value are never used. remove them, please

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
...
+        }
+        keyStorage[counter] = key;
+        valueStorage[counter] = value;
+        counter++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < counter; i++) {
+            if (key == null & key == keyStorage[i] || (key != null & keyStorage[i].equals(key))) {
```
```suggestion
            if (key == null & key == keyStorage[i] || keyStorage[i].equals(key)) {
```

PR #73, opened 2019-12-15
Finished projet

MrNopi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,52 @@
 
 import core.basesyntax.Storage;
 
+import java.util.Objects;
+
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private int size = 0;
+    private static String[][] storage = new String[10][2];
```
Tarabarnik: try to solve without two-dimensional array

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,52 @@
...
+    private static final int INDEX_OF_KEY = 0;
+    private static final int INDEX_OF_VALUE = 1;
```
Tarabarnik: redundant constant

src/main/java/core/basesyntax/impl/StorageImpl.java
Tarabarnik: also, 10 us a magic number so should be constant :)

PR #81, opened 2020-04-06
hw6.2.2

Andrewmazyar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,51 @@
...
+
+    public StorageImpl() {
+        count = 0;
+        keys = (K[]) new Object[SIZE];
+        values = (V[]) new Object[SIZE];
+    }
+
+    private boolean ifExist(K key) {
+        for (K k : keys) {
+            if (Objects.equals(key, k)) {
```
RomanDubovskyi, line 20: let's try to solve it without Objects class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,51 @@
...
 
     @Override
     public void put(K key, V value) {
+        if (!ifExist(key)) {
+            this.keys[count] = key;
+            this.values[count] = value;
+            count++;
+        } else {
+            for (int i = 0; i < keys.length; i++) {
+                if (keys[i] == key) {
```
RomanDubovskyi, line 35: do we compare objects with == ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,51 @@
...
+                    values[i] = value;
+                }
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(key, keys[i])) {
```
RomanDubovskyi, line 45: same with Objects class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
+    private V[] values;
+    private int count;
+
+    public StorageImpl() {
+        count = 0;
+        keys = (K[]) new Object[SIZE];
+        values = (V[]) new Object[SIZE];
+    }
+
+    private boolean ifExist(K key) {
```
RomanDubovskyi, line 17: you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,45 @@
...
```
RomanDubovskyi, line 17: i don't see you using this method anywhere, so why is it here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,45 @@
...
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                values[i] = value;
+            }
+        }
+        this.keys[count] = key;
```
RomanDubovskyi, line 33: should we execute this code every time we call put() method ?

RomanDubovskyi:
i don't see this comment being fixed:
 "i don't see you using this method anywhere, so why is it here?"

Andrewmazyar:
> i don't see this comment being fixed:
> "i don't see you using this method anywhere, so why is it here?"

i don`t understand what do you mean

PR #84, opened 2020-04-06
storage

Vinglish:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keyStorage = new Object[10];
+    private Object[] valueStorage = new Object[10];
```
RomanDubovskyi, line 7: check this part according to checklist 

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,46 @@
...
     public V get(K key) {
+        for (int i = 0; i < keyStorage.length; i++) {
+            if (compare(keyStorage[i], key)) {
+                return (V) valueStorage[i];
+            }
+        }
         return null;
     }
+
+    private boolean find(K key) {
```
RomanDubovskyi, line 35: btw you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

src/main/java/core/basesyntax/impl/StorageImpl.java
Vinglish: done

src/main/java/core/basesyntax/impl/StorageImpl.java
Vinglish: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,51 @@
...
+    public StorageImpl() {
+        index = 0;
+        keyStorage = (K[]) new Object[LENGTH];
+        valueStorage = (V[]) new Object[LENGTH];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (find(key) >= 0) {
+            for (int i = 0; i < keyStorage.length; i++) {
```
RomanDubovskyi, line 21: why do you need execute for loop if you already have index where to put from method find() ?

PR #87, opened 2020-04-06
add solution for storage

fruts:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,55 @@
...
+        if (key != null && value != null) {
+            for (int i = 0; i < keys.length; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                }
+            }
+        }
+    }
+
+    public boolean isExist(K[] keys, K key) {
```
RomanDubovskyi, line 35: btw you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

And you don't need to pass keys here, you have acces to it anyway + you only one array keys

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,55 @@
...
+                }
+            }
+        }
+        return false;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(key,keys[i])) {
```
RomanDubovskyi, line 49: let's solve it without Objects class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,55 @@
...
 
     @Override
     public void put(K key, V value) {
+        if (!isExist(keys, key)) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        }
+
+        if (key != null && value != null) {
```
RomanDubovskyi, line 26: you should be able to pass null as a value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,51 @@
...
+    }
 
     @Override
     public void put(K key, V value) {
+        if (indexOf(key) == index) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        }
+        if (key != null) {
```
RomanDubovskyi, line 24: you should be able to store only one pair with unique key (null, object), but in your current impl is you keep adding objects with null as key and different values it will add them all to your storage.
Replace your check in indexOf() method to the one, you use in get() and you'll be able to work with nulls and you won't need to execute for loop inside put()

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
     public void put(K key, V value) {
+        if (indexOf(key) == index) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        }
+        values[indexOf(key)] = value;
+    }
+
+    public int indexOf(K key) {
```
RomanDubovskyi, line 27: if you use this method as util method only inside this class make it private

PR #89, opened 2020-04-06
added solution of hw 6-2-2

KseniiaMakarova:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+    public StorageImpl() {
+        keys = new Object[ARRAY_SIZE];
+        values = new Object[ARRAY_SIZE];
+        size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null ? key == null : keys[i].equals(key)) {
```
RomanDubovskyi, line 20: Optional: you can change this check with `(key == keys[i] || (key != null && key.equals(keys[i]))` for better readability, rest is ok

KseniiaMakarova: thank you! exactly, we can compare two nulls with a ==

PR #90, opened 2020-04-06
Added solution

Allduiin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static Object[][] o = new Object[10][2];
```
RomanDubovskyi, line 6: why static? 
and check it with checklist

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
...
+    private static int i = 0;
 
     @Override
     public void put(K key, V value) {
+        for (int j = 0; j < i; j++) {
+            if ((key == null && o[j][0] == null) || (o[j][0] != null && o[j][0].equals(key))) {
```
RomanDubovskyi, line 12:
```suggestion
            if ((key == o[j][0]) || (o[j][0] != null && o[j][0].equals(key))) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
...
+        }
+        o[i][0] = key;
+        o[i][1] = value;
+        i++;
     }
 
     @Override
     public V get(K key) {
+        for (int j = 0; j < i; j++) {
+            if ((key == null && o[j][0] == null) || (o[j][0] != null && o[j][0].equals(key))) {
```
RomanDubovskyi, line 25: same

src/main/java/core/basesyntax/impl/StorageImpl.java
Allduiin: static because its class won't have any objects and this array for class not have any objects

Allduiin: So I need to do this by two arrays?

Allduiin: @RomanDubovskyi

Allduiin: Static because we need it for working class not objects of the class, so I need to do this by two arrays? @RomanDubovskyi

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final byte arrayLength = 10;
+    private static int i = 0;
```
RomanDubovskyi, line 7:
```suggestion
    private int i = 0;
```

PR #92, opened 2020-04-06
added solution

6ALLIKA:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
 
     @Override
     public void put(K key, V value) {
+        if (!contains(key)) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        } else {
+            for (int i = 0; i < LENGTH; i++) {
+                if (keys[i] == key) {
```
RomanDubovskyi, line 25: should we compare objects with == ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
     public V get(K key) {
+        for (int i = 0; i < LENGTH; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
+
+    private boolean contains(K key) {
```
RomanDubovskyi, line 42: btw you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

PR #94, opened 2020-04-06
Solution hw622

woododoCode:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
 
     @Override
     public void put(K key, V value) {
+        if (!exists(key)) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        }
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == key) {
```
RomanDubovskyi, line 26: what does == check when comparing two objects?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
+    }
 
     @Override
     public void put(K key, V value) {
+        if (!exists(key)) {
+            keys[index] = key;
+            values[index] = value;
+            index++;
+        }
+        for (int i = 0; i < keys.length; i++) {
```
RomanDubovskyi, line 25: do we need this for loop to be executed every time we call put method ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
+            if (keys[i] == key) {
+                values[i] = value;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(key, keys[i])) {
```
RomanDubovskyi, line 35: let's not use Objects class here, ok?

PR #97, opened 2020-04-06
Hm

MaksymMatuliak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_LENGTH = 10;
+    K[] keys = (K[]) new Object[MAX_LENGTH];
```
RomanDubovskyi, line 7: check this part according to 

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
...
+    V[] values = (V[]) new Object[MAX_LENGTH];
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null || keys[i] == key) {
```
RomanDubovskyi, line 13: what does == check when comparing two objects ? 
and you probably want to divide your logic into two situation:
 - when key.equals(keys[i]) so you just set new value in position "i"
 - when you have new key and value and you setting them both

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
...
+                values[i] = value;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key == null) {
```
RomanDubovskyi, line 24: try to make this check use only one if statement, it will have better readability

MaksymMatuliak: done

src/main/java/core/basesyntax/impl/StorageImpl.java
MaksymMatuliak: done

src/main/java/core/basesyntax/impl/StorageImpl.java
MaksymMatuliak: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,33 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_LENGTH = 10;
+    K[] keys;
+    V[] values;
```
RomanDubovskyi, line 8: should there be default access modifiers ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,33 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_LENGTH];
+        this.values = (V[]) new Object[MAX_LENGTH];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null || (key != null && key.equals(keys[i]))) {
```
RomanDubovskyi, line 18: with this impl it's impossible to store multiple values null with different keys, it will override first one it meets all the time, fix it please

src/main/java/core/basesyntax/impl/StorageImpl.java
MaksymMatuliak: done

PR #100, opened 2020-04-06
added solution 1

tanchak123:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,53 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys;
+    private V[] values;
+    private int limit;
+
+    public StorageImpl() {
+        limit = 0;
+        keys = (K[]) new Object[10];
```
RomanDubovskyi, line 13: make 10 cosntant according to checklist 

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,53 @@
...
+                    values[i] = value;
+                }
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(key, keys[i])) {
```
RomanDubovskyi, line 35: let's try not using Objects class here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,53 @@
...
+                return values[i];
+            }
+        }
         return null;
     }
+
+    private boolean duplicateKey(K key) {
```
RomanDubovskyi, line 42: btw you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

tanchak123: Да.Неплохая идея.Спасибо.

PR #101, opened 2020-04-06
added solution

popovychyb:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,87 @@
...
+            }
+        }
         return null;
     }
+
+    private static class Entry<K, V> {
+        private K key;
+        private V value;
+
+        public Entry(K key, V value) {
```
RomanDubovskyi, line 44: do we need public constructor for Entry here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,87 @@
...
+            this.key = key;
+            this.value = value;
+        }
+
+        public K getKey() {
```
RomanDubovskyi, line 49: same here with access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,87 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        if (entries[0] == null) {
+            return null;
+        }
+        for (int i = 0; i < entries.length; i++) {
+            if (key == entries[i].getKey() || (key != null && key.equals(entries[i].getKey()))) {
```
RomanDubovskyi, line 33: what happens with entries[i].getKey() if it's not initialized ? 
for example we have initialize only entries[0] and entries[1] == null

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,87 @@
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+
+            Entry<?, ?> entry = (Entry<?, ?>) o;
```
RomanDubovskyi, line 74: make as in previous task

src/main/java/core/basesyntax/impl/StorageImpl.java
popovychyb: Done. Yes, private is better here. Thank you!

src/main/java/core/basesyntax/impl/StorageImpl.java
popovychyb: Done. Thank you!

src/main/java/core/basesyntax/impl/StorageImpl.java
popovychyb: Done. Fixed checking for not initialized element of the array. Thank you!

src/main/java/core/basesyntax/impl/StorageImpl.java
popovychyb: Done. Thank you!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,87 @@
...
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+
+            Entry<K, V> entry = (Entry) o;
+
+            if (key != null ? !key.equals(entry.key) : entry.key != null) {
```
RomanDubovskyi, line 76: optimize this check to smt like this 
```
return (entry.getKey() == key
                || (entry.getKey != null && entry.getKey().equals(key)))
                && (entry.getValue() == value
                || (entry.getValue() != null && entry.getValue().equals(value)));
```

popovychyb: Done. Thank you!

PR #106, opened 2020-04-06
hw 6.2.2

Djeckkenni:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
     public V get(K key) {
+        for (int i = 0; i < index; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                return values[i];
+            }
+        }
         return null;
     }
+
+    public boolean containKey(K key) {
```
RomanDubovskyi, line 42: btw you can return from this method not boolean but index of element if it exists, so you won't need to run extra loop in put method

PR #113, opened 2020-04-07
added solution

matrix880:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,38 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    public static final int ARRAY_MAX_LENGTH = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int count = 0;
```
RomanDubovskyi, line 10: where should we initalize non-constant field?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,38 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    public static final int ARRAY_MAX_LENGTH = 10;
```
RomanDubovskyi, line 7: it better to make it private if we use this variable only in this class

src/main/java/core/basesyntax/impl/StorageImpl.java
matrix880: in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
matrix880: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_LENGTH = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int count = 0;
```
RomanDubovskyi, line 9: and what about int count?

matrix880: done

PR #115, opened 2020-04-07
solution to hw-6.2.2

SpringWay:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,37 @@
...
     @Override
-    public void put(K key, V value) {
+    public void put(K keyPut, V valuePut) {
+        for (int i = 0; i < size; i++) {
+            if (keyPut == keys[i] || (keyPut != null && keyPut.equals(keys[i]))) {
+                values[i] = valuePut;
+            }
+        }
+        keys[size] = keyPut;
+        values[size] = valuePut;
```
RomanDubovskyi, line 25: do we want this code to be executed every time we call method put() ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,14 +2,38 @@
...
+        keys = (K[]) new Object[STORAGE_LENGTH];
+        values = (V[]) new Object[STORAGE_LENGTH];
+        size = 0;
+    }
 
     @Override
-    public void put(K key, V value) {
+    public void put(K keyPut, V valuePut) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null ? keyPut == null : keys[i].equals(keyPut)) {
```
RomanDubovskyi, line 20: make here same check as in get() method, it will be easier to understand logic of code

PR #117, opened 2020-04-07
added solution

MaksymVakuliuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH_OF_STORAGE = 10;
+    private StorageCell<K, V>[] cell;
+    private int firstEmptyCell;
```
RomanDubovskyi, line 8: this naming isn't logical according to work of the program

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
+
+    public StorageImpl() {
+        cell = new StorageCell[10];
```
RomanDubovskyi, line 11: why do we need constant if we aren't using it ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (StorageCell<K, V> tempCell : cell) {
+            if (tempCell == null) {
+                return null;
+            }
+            if (tempCell.getKey() == null ? key == null : tempCell.getKey().equals(key)) {
```
```suggestion
            if (tempCell.getKey() ==  key || key != null && key.equals(tempCell.getKey())) {
```
make same changes to other checks

src/main/java/core/basesyntax/impl/StorageCell.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.impl;
+
+public class StorageCell<K, V> {
```
RomanDubovskyi, line 3: if you use this class only as a helper for your StorageImpl it will be better if you put it inside of StorageImpl and make it private

PR #118, opened 2020-04-07
Added solution hw-6.2.2  jv-storage

nevskyy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
...
     public V get(K key) {
+        for (int i = 0; i < index; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                return values[i];
+            }
+        }
         return null;
     }
+
+    private boolean containsKey(K key) {
```
RomanDubovskyi, line 42: you can change this method to indexOf() and return index of element it it exists, if not return -1 or smt like this

nevskyy: Changed as recommended. Thanks.

PR #120, opened 2020-04-07
Hw 6.2.2

Kaijou88:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,36 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    public static final int LENGTH = 10;
+
+    K[] arrayKey;
+    V[] arrayValue;
+    int next;
```
RomanDubovskyi, line 11: should these access modifiers be default ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,36 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < next; i++) {
+            if ((arrayKey[i] == key) || (arrayKey[i] != null && arrayKey[i].equals(key))) {
+                arrayValue[i] = value;
+            }
+        }
+        arrayKey[next] = key;
+        arrayValue[next++] = value;
```
RomanDubovskyi, line 27: should we execute this code even if there is already element with such key and we updated value for it ?

Kaijou88: No. you're right. Added return.

src/main/java/core/basesyntax/impl/StorageImpl.java
Kaijou88: added

PR #128, opened 2020-08-04
ready solution hw-7.2

RomanHreits:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
...
+                if ((keyStorage[i] == key) || (((keyStorage[i] != null)
+                        && keyStorage[i].equals(key)))) {
+                    return valueStorage[i];
+                }
+            }
+        }
         return null;
     }
+
+    private <E> boolean checkKeys(E[] keysArray, E key) {
```
dmytro-hryhoruk, line 42: why do you need the third type parameter?

dmytro-hryhoruk, line 42: can we modify this method to use this in get too?

RomanHreits: done

RomanHreits: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+
+    public StorageImpl() {
+        valueStorage = (V[]) new Object[CAPACITY];
+        keyStorage = (K[]) new Object[CAPACITY];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (key != null & checkKeys(keyStorage, key) != -1) {
+            keyStorage[index] = key;
```
dmytro-hryhoruk, line 20: do we need to do this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private V[] valueStorage;
+    private K[] keyStorage;
+    private int size = 0;
+    private int index = 0;
```
dmytro-hryhoruk, line 10: it's not the state of the object. You are using this as temp variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+            return valueStorage[index];
+        }
         return null;
     }
+
+    private int checkKeys(K[] keysArray, K key) {
+        for (int i = 0; i < keysArray.length; i++) {
+            if (keysArray[i] == key || (keysArray[i] != null && keysArray[i].equals(key))) {
+                index = i;
+                return index;
```
dmytro-hryhoruk, line 41:
```suggestion
                return i;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+    private int index = 0;
+
+    public StorageImpl() {
+        valueStorage = (V[]) new Object[CAPACITY];
+        keyStorage = (K[]) new Object[CAPACITY];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (key != null & checkKeys(keyStorage, key) != -1) {
```
dmytro-hryhoruk, line 19: You can save the result of checkKeys to local variable and use this instead of field index

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+        } else {
+            keyStorage[size] = key;
+            valueStorage[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (keyStorage != null && valueStorage != null & checkKeys(keyStorage, key) != -1) {
```
dmytro-hryhoruk, line 31: same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+            return valueStorage[index];
+        }
         return null;
     }
+
+    private int checkKeys(K[] keysArray, K key) {
```
dmytro-hryhoruk, line 37:
```suggestion
    private int getElementIndex(K[] keysArray, K key) {
```

dmytro-hryhoruk: keysArray it's a Storage field. We have access from this from any place inside the class

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: done

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: done

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: done

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: done

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: yes

dmytro-hryhoruk: is it?

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-hryhoruk: ?

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: key != null - is unnecesary - i removed it

src/main/java/core/basesyntax/impl/StorageImpl.java
RomanHreits: i removed (K[] keysArray)

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-hryhoruk: if index != -1 it's mean that you are overriding a value for already existed key, right ?so in that case do you need this operation?

RomanHreits: yes, when such a key is already in the collection returns 1, if not returned -1

RomanHreits: Я не зрозумів відразу що ти мав на увазі  keyStorage[index] = key;
Я все виправив.

PR #132, opened 2020-08-04
Added solution 7.2

DmGrt:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
+    private int size;
+
+    public StorageImpl() {
+        this.storage = new Pair[MASS_LENGTH];
+        this.size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MASS_LENGTH; i++) {
```
dmytro-hryhoruk, line 17: we are not changing the array capacity, but it's would be better to use the current array.length

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,48 @@
...
+            if (storage[i] != null && storage[i].equals(Pair.of(key, value))
+                    || i == size || storage[i].key == key) {
```
dmytro-hryhoruk, line 19: 1) You don't have overridedn equals for Pair class;
2) Do we need to compare Pairs objects or something else?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,68 @@
...
+            return key == null ? key == pair.key : key.equals(pair.key)
+                    && value == null ? value == pair.value : value.equals(pair.value);
+        }
+
+        @Override
+        public int hashCode() {
+            int primaryNumber = 17;
+            return ((primaryNumber * 7 + (key == null ? 0 : key.hashCode()))
+                    * 7 + (value == null ? 0 : value.hashCode()));
+        }
```
dmytro-hryhoruk, line 68: are you using the pari object as Keys?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,68 @@
...
+
+    public StorageImpl() {
+        this.storage = new Pair[MASS_LENGTH];
+        this.size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < storage.length; i++) {
+            if (storage[i] != null && storage[i].equals(key)
```
dmytro-hryhoruk, line 18: you are comparing object of type K and with the object of type Pair<K, V>

PR #133, opened 2020-08-04
homework7.2

Alkey:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,39 @@
...
+    public StorageImpl() {
+        this.key = (K[]) new Object[SIZE];
+        this.value = (V[]) new Object[SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (iterator > 0) {
+            for (int i = 0; i < this.key.length; i++) {
+                if (((this.key[i] != null) && this.key[i].equals(key))) {
```
KseniiaMakarova, line 21: what if both `this.key[i]` and `key` are null?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,40 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private static final int SIZE = 10;
+    private K[] key;
+    private V[] value;
+    private int iterator;
+
+    public StorageImpl() {
+        this.key = (K[]) new Object[SIZE];
+        this.value = (V[]) new Object[SIZE];
```
KseniiaMakarova, line 14: do we need to use `this` all over this class?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,40 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (iterator > 0) {
```
KseniiaMakarova, line 19: this check is redundant
if `iterator` is 0, you will skip the loop

PR #134, opened 2020-08-04
Add solution

OleksiiTromsa:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_SIZE = 10;
```
KseniiaMakarova, line 6: what is wrong here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+
+    public int getSize() {
+        return size;
+    }
 
     @Override
     public void put(K key, V value) {
+
+        for (int i = 0; i < size; i++) {
+            if (key == null ? keys[i] == null : key.equals((K) keys[i])) {
```
KseniiaMakarova, line 26:
```suggestion
            if (key == null ? keys[i] == null : key.equals(keys[i])) {
```
same for the other method

src/main/java/core/basesyntax/impl/StorageImpl.java
OleksiiTromsa: Forgot about access modificator. Fixed.

PR #135, opened 2020-08-04
added solution

MaliukDaria:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STORAGESIZE = 10;
```
dmytro-hryhoruk, line 6:
```suggestion
    public static final int STORAGE_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,39 @@
...
+    private int counter = 0;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGESIZE];
+        values = (V[]) new Object[STORAGESIZE];
+    }
 
     @Override
     public void put(K key, V value) {
+        boolean isKeyUnique = true;
```
dmytro-hryhoruk, line 18: that redundant variable. try to use return; in your code and you will be able to get rid of this

src/main/java/core/basesyntax/impl/StorageImpl.java
MaliukDaria: done

src/main/java/core/basesyntax/impl/StorageImpl.java
MaliukDaria: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < counter; i++) {
+            if (keys[i] == null ? keys[i] == key : keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        this.keys[counter] = key;
+        this.values[counter] = value;
```
dmytro-hryhoruk, line 25:
```suggestion
        keys[counter] = key;
        values[counter] = value;
```

MaliukDaria: done

PR #136, opened 2020-08-04
added solution

tarkaiev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,89 @@
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+
+            Pair<?, ?> pair = (Pair<?, ?>) o;
```
KseniiaMakarova, line 73: why using wildcard?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,89 @@
...
+            return (this.getFirst() == null || pair.getFirst() == null)
+                    ? this.getFirst() == pair.getFirst()
+                    : this.getFirst().equals(pair.getFirst())
+                    && (this.getSecond() == null || pair.getSecond() == null)
+                    ? this.getSecond() == pair.getSecond()
+                    : this.getSecond().equals(pair.getSecond());
```
KseniiaMakarova, line 79: what's the point of using `this` and getters?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,89 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
```
KseniiaMakarova, line 7: don't separate fields with blank lines

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,89 @@
...
+    public StorageImpl() {
+        storageArray = new Pair[MAX_SIZE];
+        size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        Pair<K, V> pair = Pair.of(key, value);
+        for (int i = 0; i < MAX_SIZE; i++) {
```
KseniiaMakarova, line 20: do we need to iterate over the whole array, or only the filled part?
please refactor this method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,89 @@
...
+                    size++;
+                }
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (Pair<K, V> pair : storageArray) {
```
KseniiaMakarova, line 34: same: do we need to iterate over the whole array, or only the filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: mistake that follows me from previous task
fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: tried to avoid unchecked cast but made it with mistake
fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: done

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: understood, done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,88 @@
...
+        size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        Pair<K, V> pair = Pair.of(key, value);
+        for (int i = 0; i <= size; i++) {
+            if (i == size || storageArray[i].getFirst() == key
+                    || storageArray[i] != null && storageArray[i].equals(key)) {
+                storageArray[i] = pair;
```
KseniiaMakarova, line 21: the problem is we are creating new Pair object even if we need just to change value

let's create a new object only if `i == size`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,88 @@
...
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            Pair<K,V> pair = storageArray[i];
+            if (pair != null && (pair.getFirst() == key
```
KseniiaMakarova, line 34: `if (pair != null` - are there cases where pair can be null?

tarkaiev: there are no such cases, fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
tarkaiev: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,91 @@
...
+
+        for (int i = 0; i <= size; i++) {
+            if (i == size) {
+                storageArray[i] = Pair.of(key, value);
+                size++;
+                return;
+            }
+            if (storageArray[i].firstVariable == key
+                    || storageArray[i].equals(key)
+                    && storageArray[i].firstVariable != null) {
```
KseniiaMakarova, line 26:
```suggestion
                    || storageArray[i].firstVariable != null && storageArray[i].firstVariable.equals(key)) {
```
can you explain why this fix is needed?

tarkaiev: this fix is needed to not create new Pair object when key is existing. This fix will just change second pair value of existing key
Also, understood that null checking is not necessary because key of existing pair cannot be null
fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,90 @@
...
     public void put(K key, V value) {
+
+        for (int i = 0; i <= size; i++) {
+            if (i == size) {
+                storageArray[i] = Pair.of(key, value);
+                size++;
+                return;
+            }
+            if (storageArray[i].firstVariable == key
+                    || storageArray[i].equals(key)) {
```
KseniiaMakarova, line 25:
```suggestion
            if (storageArray[i].firstVariable == key
                    || storageArray[i].firstVariable != null 
                    && storageArray[i].firstVariable.equals(key)) {
```

PR #137, opened 2020-08-04
Homework 7.2 completed

Borisko130:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,42 @@
...
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private static final int ARRAY_SIZE = 10;
+
+    private final K[] keysArray;
+    private final V[] valuesArray;
```
dmytro-hryhoruk, line 11: why final?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,42 @@
...
+        Arrays.fill(keysArray, null);
+        Arrays.fill(valuesArray, null);
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if ((key == keysArray[i]
+                    || (key != null && key.equals(keysArray[i])))
+                    || valuesArray[i] == null) {
```
dmytro-hryhoruk, line 25: Probably somebody put pair key, value with value == null on purpose. That not the reason to override this value.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,42 @@
...
+    private static final int ARRAY_SIZE = 10;
+
+    private final K[] keysArray;
+    private final V[] valuesArray;
+
+    public StorageImpl() {
+        keysArray = (K[]) new Object[ARRAY_SIZE];
+        valuesArray = (V[]) new Object[ARRAY_SIZE];
+        Arrays.fill(keysArray, null);
+        Arrays.fill(valuesArray, null);
```
dmytro-hryhoruk, line 17: if yo won't do that. Our arrays will be filled with what?

PR #140, opened 2020-08-04
added solution

glones:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private static final int arrayLength = 10;
```
KseniiaMakarova, line 7: what is wrong here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
+
+    public StorageImpl() {
+        keyArray = (K[]) new Object[arrayLength];
+        valueArray = (V[]) new Object[arrayLength];
+        counter = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean checkKay = true;
```
KseniiaMakarova, line 20: let's remove this and just exit the method if we need

src/main/java/core/basesyntax/impl/StorageImpl.java
glones: check style maybe? corrected

src/main/java/core/basesyntax/impl/StorageImpl.java
glones: done

PR #142, opened 2020-08-04
Hw 7.2

pavlokhomiak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private static final Object[] KEY_ARRAY = new Object[STORAGE_SIZE];
+    private static final Object[] VALUE_ARRAY = new Object[STORAGE_SIZE];
```
KseniiaMakarova, line 8: these should not be constants, but should be class fields initialised in constructor
also, let's create one more field that will store `currentSize` of the arrays, and refactor your solution using this field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < KEY_ARRAY.length; i++) {
+            if (key == null && KEY_ARRAY[i] == null) {
+                return (V) VALUE_ARRAY[i];
+            }
+            if (KEY_ARRAY[i] != null && KEY_ARRAY[i].equals(key)) {
+                return (V) VALUE_ARRAY[i];
+            }
```
KseniiaMakarova, line 36: are two checks necessary?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,47 @@
...
+            if (keyArray[i] == null) {
+                keyArray[i] = key;
+                valueArray[i] = value;
+                storageSize++;
+                break;
+            }
+            if (keyArray[i].equals(key)) {
+                valueArray[i] = value;
+                storageSize++;
+                break;
```
KseniiaMakarova, line 34: let's refactor this
you get `K key` in this method
you should iterate over the **filled** part of your `keyArray` array in a `for` loop and check if this key is already present in your array
if you find a match at any step of the iteration, you should change value and `return` from the method

if no match is found and you **go out** of the loop, add key and value to your arrays (currently this is implemented on lines 26-28)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,47 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
```
KseniiaMakarova, line 41: do we need to iterate over the whole array, or only the filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,47 @@
...
+            if ((key == null && keyArray[i] == null)
+                    || (keyArray[i] != null && keyArray[i].equals(key))) {
+                return (V) valueArray[i];
```
KseniiaMakarova, line 44: 👍

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private int storageSize = 0;
```
```suggestion
    private int storageSize;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
+        keyArray = new Object[STORAGE_CAPACITY];
+        valueArray = new Object[STORAGE_CAPACITY];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < storageSize; i++) {
+            if (keyArray[i] == null && valueArray[i] != null) {
+                continue;
+            }
```
KseniiaMakarova, line 21: remove this - this is redundant
does it matter if valueArray[i] is null?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
+            if (keyArray[i].equals(key)) {
```
KseniiaMakarova, line 22: let's refactor this check
think about NullPointerException

hint: you are doing the same check in get() method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
+                valueArray[i] = value;
+                storageSize++;
```
KseniiaMakarova, line 24: why are we incrementing `storageSize` if we just replaced the value?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+
+    public StorageImpl() {
+        keyArray = new Object[STORAGE_CAPACITY];
+        valueArray = new Object[STORAGE_CAPACITY];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < storageSize; i++) {
+            if (keyArray[i] != null && keyArray[i].equals(key)) {
```
KseniiaMakarova, line 19: add check if both `keyArray[i]` and `key` are null

PR #143, opened 2020-08-04
added solution 7.2

TimoxaV:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,80 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Pair<K, V>[] resultArray;
```
dmytro-hryhoruk, line 7: why final?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,80 @@
...
+
+    public StorageImpl() {
+        resultArray = new Pair[MAX_SIZE];
+    }
 
     @Override
     public void put(K key, V value) {
+        Pair<K, V> pairToStore = Pair.of(key, value);
+        for (int i = 0; i < MAX_SIZE; i++) {
```
dmytro-hryhoruk, line 16: let's wire this loop to current array size, but not to constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,80 @@
...
+            if (resultArray[i] == null) {
+                resultArray[i] = pairToStore;
+                return;
+            } else if ((resultArray[i].getKey() == null
```
dmytro-hryhoruk, line 20: do we need else after return?

dmytro-hryhoruk, line 20: You already made that check upon

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,80 @@
...
+            if (this == obj) {
+                return true;
+            }
+            if (obj == null || !obj.getClass().equals(Pair.class)) {
+                return false;
+            }
+            Pair objCast = (Pair) obj;
+            return (key != null ? key.equals(objCast.key) : key == objCast.key)
+                    && (value != null ? value.equals(objCast.value) : value == objCast.value);
+        }
```
dmytro-hryhoruk, line 80: are you using the objects of this class as keys?

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: That's my mistake. It must not be final, because values inside array can be changed.

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: Done.

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: Done. It doesn't make any difference here. I have deleted it.

TimoxaV: No, I check if there is a Pair at “i” position upon. The second check is for the key inside this Pair. Null keys are occured in test cases.

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: I use Pair for keeping key and value inside it.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,83 @@
...
+        for (int i = 0; i <= size; i++) {
+            if (resultArray[i] == null) {
+                resultArray[i] = pairToStore;
+                size++;
+                return;
+            }
+            if ((resultArray[i].getKey() == null
+                    ? resultArray[i].getKey() == key : resultArray[i].getKey().equals(key))) {
+                resultArray[i].value = pairToStore.value;
+                return;
```
dmytro-hryhoruk, line 26: do we need that return?

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-hryhoruk: Yes, but do you need the equals and hashcode for Pair?

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: Actually, no. I've deleted it. Thanks.

src/main/java/core/basesyntax/impl/StorageImpl.java
TimoxaV: You are right, I don't need these methods. Fixed.

PR #145, opened 2020-08-04
add solution

voloshynamaryna11:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,45 @@
...
+
+    public StorageImpl() {
+        keys = new Object[10];
+        array = new Object[10];
+        score = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int q = 0;
```
dmytro-hryhoruk, line 19: what for do you need this variable? you can have more than one value for each key?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,45 @@
...
+
+        for (int a = 0; a < keys.length; a++) {
+            if ((key == null && keys[a] == null)
+                    || (key != null && key.equals((K) (keys[a])))) {
```
dmytro-hryhoruk, line 23: try to simplify  that

dmytro-hryhoruk: do you need that cast?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,37 @@
...
     public void put(K key, V value) {
+        for (int a = 0; a < keys.length; a++) {
+            if (key != null ? key.equals(keys[a])
+                    : keys[a] == null) {
+                array[a] = value;
+            }
+        }
+        keys[score] = key;
+        array[score] = value;
+        score++;
```
dmytro-hryhoruk, line 27: even if you find that key and override the value, you still gonna add a new element?

PR #151, opened 2020-08-05
added solution

matviichuk92:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX = 10;
+    private int size = 0;
+    private K[] keys = (K[]) new Object[MAX];
+    private V[] values = (V[]) new Object[MAX];
 
     @Override
     public void put(K key, V value) {
+        if (isKeyExist(key)) {
+            for (int i = 0; i < size; i++) {
```
dmytro-hryhoruk, line 14: you are iteration through the whole array twice. Let's optimize this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
+                if (isKeyExist(keys[i])) {
+                    values[i] = value;
+                }
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
```
dmytro-hryhoruk, line 22: in this case if the same key was found you will override it's value and add a new element to the end of array anyway

matviichuk92:
done

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-hryhoruk: still the same

matviichuk92:
done

PR #152, opened 2020-08-05
added solution

floriangeyer21:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_SIZE = 10;
```
dmytro-hryhoruk, line 6: access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,42 @@
...
+    }
+
+    public int getSize() {
+        return size;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == null ? keys[i] == null : key.equals((K) keys[i])) {
```
dmytro-hryhoruk, line 25: do we need that cast?

PR #154, opened 2020-08-05
solution added

Vosker:


src/main/java/core/basesyntax/impl/MyBox.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.impl;
+
+public class MyBox<K, V> {
```
dmytro-hryhoruk, line 3: do you use this class somewhere else except StorageImpl?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
     }
+
+    private boolean isEquals(K key, K keyFromBox) {
+        if (keyFromBox == null) {
+            return keyFromBox == key;
+        }
+        return keyFromBox.equals(key);
+    }
 }
+
```
dmytro-hryhoruk, line 44:
```suggestion
```

PR #155, opened 2020-08-05
hw7.2

OlessiaKharchenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
...
+        values = (V[]) new Object[SIZE];
+        count = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < SIZE; i++) {
+            if ((keys[i] == null && key == null)
+                    || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
```
KseniiaMakarova, line 22: if you set the value, do you need to continue working in this method?

PR #156, opened 2020-08-05
added solution

AnaLanda:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE = 10;
+    int size;
```
dmytro-hryhoruk, line 7: access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[SIZE];
+        this.values = (V[]) new Object[SIZE];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < SIZE; i++) {
+            if (key == keys[i]) {
```
dmytro-hryhoruk, line 19: String firstKey = new String("Key")
String secondKey = new String("Key");
System.out.println(firstKey == secondKey);
What will be printed?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < SIZE; i++) {
+            if (key == keys[i]
+                    || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                size++;
+                break;
+            } else if (keys[i] == null) {
```
dmytro-hryhoruk, line 24: do we need this check and to do that in loop?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < SIZE; i++) {
+            if (key == keys[i]
+                    || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                size++;
+                break;
+            } else {
```
dmytro-hryhoruk, line 24: do you need else after break?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,41 @@
...
+                keys[size] = key;
+                values[size] = value;
+                size++;
+                break;
```
dmytro-hryhoruk, line 28: do we need break in this place?

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-hryhoruk: so now if it won't find in index 0 the same key it will not check further but add a new element immediately ?

AnaLanda: Yeah, I guess. Will change this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
...
+        for (int i = 0; i < SIZE; i++) {
+            if (key == keys[i]
+                    || key != null && key.equals(keys[i])) {
+                values[i] = value;
+            }
+            if (keys[i] == null) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+                break;
```
dmytro-hryhoruk, line 27: let's move it after loop. If we are not gonna  find the same key in our storage only then we will add a new element

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+        this.values = (V[]) new Object[SIZE];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < SIZE; i++) {
+            if (key == keys[i]
+                    || key != null && key.equals(keys[i])) {
+                values[i] = value;
+            }
```
dmytro-hryhoruk, line 22:
```suggestion
                values[i] = value;
                return;
            }
```

dmytro-hryhoruk: we should not add a new value if override by the same key

PR #159, opened 2020-08-05
hw 7.2

mikhail771:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int ARRAY_LENGTH = 10;
+    int generalKey = 0;
```
dmytro-hryhoruk, line 7: access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,52 @@
...
+        int cellNumber = getCellNumber(key);
+        if (cellNumber == -1) {
+            return null;
+        }
+        return values[cellNumber];
+    }
+
+    private boolean isKeyExists(K key) {
+        for (int i = 0; i < generalKey; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
dmytro-hryhoruk, line 33: can we have key null in our Storage?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,52 @@
...
+        return false;
+    }
+
+    private int getCellNumber(K key) {
+        for (int i = 0; i < generalKey; i++) {
+            if (keys[i] == null) {
+                if (keys[i] == key) {
+                    return i;
+                }
+            } else {
```
dmytro-hryhoruk, line 46: do we need else after return?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,52 @@
...
+                if (keys[i].equals(key)) {
+                    return i;
+                }
```
dmytro-hryhoruk, line 49: can we create a one condition?

src/main/java/core/basesyntax/impl/StorageImpl.java
mikhail771: done

src/main/java/core/basesyntax/impl/StorageImpl.java
mikhail771: done

src/main/java/core/basesyntax/impl/StorageImpl.java
mikhail771: done

src/main/java/core/basesyntax/impl/StorageImpl.java
mikhail771: yes. Test passed

dmytro-hryhoruk: test may not always cover 100% of cases. So in your current solution you can put some value with key == null?

mikhail771: done

PR #160, opened 2020-08-05
Create class storage and method get is overwriten

fyarkiy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE_OF_STORAGE = 10;
+    private K[] tableOfKey;
+    private V[] tableOfValue;
+    private int arrayIndex = 0;
```
```suggestion
    private int arrayIndex;
```

it will get default value of 0
if you still want to initialise it, it is better be done in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
...
+        for (int j = 0; j < tableOfKey.length; j++) {
+            if (tableOfKey[j] == key || (key != null && key.equals(tableOfKey[j]))) {
+                return (V) tableOfValue[j];
+            }
+        }
         return null;
     }
+
+    @Override
+    public boolean equals(Object object) {
```
KseniiaMakarova, line 43: this method will work for comparing one StorageImpl object with another
you should better remove it, because the check of fields of type `Array` is slightly different and we don't really need it here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
...
+                    && storageImpl.tableOfKey.equals(tableOfKey)))
+                    && ((tableOfValue == storageImpl.tableOfValue)
+                    || (storageImpl.tableOfValue != null
+                    && storageImpl.tableOfValue.equals(tableOfValue)));
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
```
KseniiaMakarova, line 63: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
...
+        if (j == tableOfKey.length || key == null) {
+            tableOfKey[arrayIndex] = key;
+            tableOfValue[arrayIndex] = value;
+            arrayIndex++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int j = 0; j < tableOfKey.length; j++) {
```
KseniiaMakarova, line 34: do we need to iterate over the whole array, or only the filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
...
+            if (tableOfKey[j] == key || (key != null && key.equals(tableOfKey[j]))) {
+                return (V) tableOfValue[j];
+            }
```
KseniiaMakarova, line 37: 👍

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,71 @@
...
+            if (tableOfKey[j] == key) {
+                tableOfValue[j] = value;
+                j = tableOfKey.length + 1;
+            }
+        }
+        if (j == tableOfKey.length || key == null) {
+            tableOfKey[arrayIndex] = key;
+            tableOfValue[arrayIndex] = value;
+            arrayIndex++;
+        }
```
KseniiaMakarova, line 29: let's refactor this

you should iterate over the filled part of your `tableOfKey` array in a `for` loop and check if this `key` (that you receive as a method argument) is already present in your array
if you find a match at any step of the iteration, you should change value and `return` from the method (void methods do support return)

if no match is found and you go out of the loop, add key and value to your arrays (currently this is implemented on lines 26-28)

PR #163, opened 2020-08-05
task done

yutsevich:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,44 @@
 
 public class StorageImpl<K, V> implements Storage<K, V> {
 
+    private K[] keys = (K[]) new Object[10];
+    private V[] values = (V[]) new Object[10];
+    private int size = 0;
```
KseniiaMakarova, line 9: see checklist: 
Non-constant fields should be initialized at constructor.
Any magic numbers should be constants.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,44 @@
...
+            }
+        }
+
+        if (i == size) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[i] = value;
+        }
```
KseniiaMakarova, line 30: let's refactor this
you get `K key` in this method
you should iterate over the filled part of your `keys` array in a `for` loop and check if this key is already present in your array
if you find a match at any step of the iteration, you should change value and return from the method

if no match is found and you go out of the loop, add key and value to your arrays (currently this is implemented on lines 25-27)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,44 @@
...
+            while (i < size && null != (keys[i])) {
+                i++;
+            }
+        } else {
+            while (i < size && !key.equals(keys[i])) {
+                i++;
+            }
+        }
+        return i == size ? null : values[i];
     }
```
KseniiaMakarova, line 46: let's refactor this
you get K key in this method
you should iterate over the filled part of your keys array in a for loop and check if this key is already present in your array
if you find a match at any step of the iteration, you should return the value

if no match is found and you go out of the loop, return null

src/main/java/core/basesyntax/impl/StorageImpl.java
yutsevich: is an initial capacity a magic number?

PR #166, opened 2020-08-05
homeWork7.2

khOksana99:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,40 @@
...
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < LENGTH; i++) {
```
KseniiaMakarova, line 33: do we need to iterate over the whole array, or only the filled part?

khOksana99: only filled part

PR #168, opened 2020-08-05
solution

Vas35:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,35 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_ELEMENT = 10;
```
dmytro-hryhoruk, line 6: access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,35 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == arrKey[i] || key != null && key.equals(arrKey[i])) {
+                arrValue[i] = value;
+            }
+        }
+        arrKey[size] = key;
+        arrValue[size] = value;
+        size++;
```
dmytro-hryhoruk, line 25: you will override value and add this element to anyway

src/main/java/core/basesyntax/impl/StorageImpl.java
Vas35: You mean I need to put private like in other fields?

src/main/java/core/basesyntax/impl/StorageImpl.java
Vas35: Sorry, what do you mean I didn't understand?

src/main/java/core/basesyntax/impl/StorageImpl.java
Vas35: I added private field to constant.

src/main/java/core/basesyntax/impl/StorageImpl.java
Vas35: I got it.

PR #172, opened 2020-08-05
Added soulution

Nokafan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int arrSize = 10;
+    int size;
```
KseniiaMakarova, line 7: what is missing here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int arrSize = 10;
```
KseniiaMakarova, line 6: what is wrong here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+    int size;
+    private K[] keysArray;
+    private V[] valuesArray;
+
+    public StorageImpl() {
+        this.keysArray = (K[]) new Object[arrSize];
+        this.valuesArray = (V[]) new Object[arrSize];
+        size = 0;
```
KseniiaMakarova, line 14: why using `this`?

src/main/java/core/basesyntax/impl/StorageImpl.java
Nokafan: Modifier for size, should be private. Name for constant should be in uppercase. Fixed.

src/main/java/core/basesyntax/impl/StorageImpl.java
Nokafan: Uppercase. Fixed.

src/main/java/core/basesyntax/impl/StorageImpl.java
Nokafan: <<Stupid. 
Variable of the same class we have to use without this.

PR #181, opened 2020-12-05
Put and get value in a storage.

gigbat:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,39 @@
...
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < STORAGE_CAPACITY; i++) {
+            if (this.key[i] == null && this.value[i] == null) {
+                this.key[i] = key;
+                this.value[i] = value;
+                break;
+            }
+            if (this.key[i] == key) {
```
boroda4436, line 23: What is the better way to compare two objects?

boroda4436 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
gigbat: done

boroda4436 (approved):

PR #183, opened 2020-12-06
Tests completed

Artificial-Friend:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,56 @@
...
+            if ((this.key == null && key == null)
+                    || (this.key != null && this.key.equals(key))) {
+                return (V) items[i + 1];
+            }
+        }
         return null;
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
boroda4436, line 39: this is redundant, but OK

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER_WITH_KEYS = 20;
+    private final Object[] items = new Object[MAX_ITEMS_NUMBER_WITH_KEYS];
+    private int counter = 0;
+    private K key;
 
     @Override
     public void put(K key, V value) {
```
boroda4436, line 12: how this method will work if we will put a new value by the same key?

boroda4436 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER_WITH_KEYS = 20;
+    private final Object[] items = new Object[MAX_ITEMS_NUMBER_WITH_KEYS];
```
dmytro-hryhoruk, line 7: https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,61 @@
...
+    private int counter = 0;
+    private K key;
```
dmytro-hryhoruk, line 9: don't store one key as field of StorageImpl

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,61 @@
...
+        for (int i = 0; i < items.length; i += 2) {
+            this.key = (K) items[i];
+            if ((this.key == null && key == null)
+                    || (this.key != null && this.key.equals(key))) {
+                if (this.key == null && items[i + 1] == null) {
+                    this.items[i + 1] = value;
+                    counter += 2;
+                } else {
+                    this.items[i + 1] = value;
+                }
```
dmytro-hryhoruk, line 22: That's too complicated: We have two scenarios here:

!)There is already a value in Storage by the given key, Then we should replace it with the given(new value) and stop the method;
2)If given key is not presented in our array, We should add a new key with its value to the end;
In simple words you are iteration through the array, checking keys equality, And if it's found, Then replaced its value, if key has not been found you will add a new element.

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Artificial-Friend: Thank you.

dmytro-hryhoruk (approved):

PR #189, opened 2020-12-08
implement methods put() and get()

YuraMitryakhin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
+    public StorageImpl() {
+        keys = new Object[ARRAY_SIZE];
+        values = new Object[ARRAY_SIZE];
+        index = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < index; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
dmytro-hryhoruk, line 20:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
Let's accept one value by null key;

dmytro-hryhoruk (approved):
Good Job!

PR #191, opened 2020-12-08
first commit

RuslanIbIsh:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,46 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private int incrementedIndex = 0;
+    private Object[][] storage;
+
+    public StorageImpl() {
+        storage = new Object[INITIAL_CAPACITY][INITIAL_CAPACITY];
```
KseniiaMakarova, line 13: does it need to be 10x10 or 10x2 is enough?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,46 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        Object[] result = isContains(key);
+        return (V) result[2];
+    }
+
+    private Object[] isContains(K key) {
```
KseniiaMakarova, line 34:
```suggestion
    private int indexOf(K key) {
```
I would suggest you to refactor this method

let's iterate over the `keys` of your storage (you can use `incrementedIndex` for the actual length by the way)
if we find the matching key - we return the index
if not - return `-1`

KseniiaMakarova (changes requested):
overall nice solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: If you want I can rewrite my code and add new constant - give know YES or NOT ?

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: Greate idea, I understud you - But I will to beg you accept my code, cosa Im in ArraList implementation now

KseniiaMakarova (changes requested):
let's refactor `isContains()` method and add a new constant `2` to represent the `storage` dimension length

this wouldn't take much time, and we should try to write **every** solution on the best level that we can

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
-        return null;
+        if (indexOf(key) == -1) {
+            return null;
+        }
+        return (V) storage[indexOf(key)][VALUE_INDEX];
+    }
+
+    private int indexOf(K key) {
+        int index = -1;
+        for (int i = 0; i < storage.length; i++) {
```
KseniiaMakarova, line 38: should we iterate over the whole array or only the filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+    private Object[][] storage;
+
+    public StorageImpl() {
+        storage = new Object[INITIAL_CAPACITY_KEYS][INITIAL_CAPACITY_VALUES];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (indexOf(key) != -1) {
+            storage[indexOf(key)][VALUE_INDEX] = value;
```
KseniiaMakarova, line 20: let's save the result of `indexOf(key)` call into the local variable so that we would not call this method twice

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        if (indexOf(key) == -1) {
+            return null;
+        }
+        return (V) storage[indexOf(key)][VALUE_INDEX];
```
KseniiaMakarova, line 33: same here

KseniiaMakarova (changes requested):
overall great job!

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: Do you propese use brake?

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: done

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY_KEYS = 10;
+    private static final int INITIAL_CAPACITY_VALUES = 2;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private int size = 0;
+    private int incrementedIndex = 0;
```
KseniiaMakarova, line 11: these two variables are created for the same purpose - please remove one of them

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
RuslanIbIsh: done

KseniiaMakarova (approved):
good job!

PR #193, opened 2020-12-08
added file Pair; fix: Pair.equals() improved

Max-Pochepets:


src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || this.getClass() != obj.getClass()) {
+            return false;
+        }
+        Pair newPair = (Pair) obj;
```
dmytro-hryhoruk, line 32: let's cast to parametrized type

src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,46 @@
...
+    public static <T, Z> Pair<T, Z> of(T first, Z second) {
+        return new Pair<>(first, second);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || this.getClass() != obj.getClass()) {
```
dmytro-hryhoruk, line 29:
```suggestion
        if (obj == null || getClass() != obj.getClass()) {
```

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,35 @@
...
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < storage.length; i++) {
+            if (storage[i] == null) {
+                storage[i] = Pair.of(key, value);
+                break;
+            }
+            if (storage[i].getFirst() == key) {
```
dmytro-hryhoruk, line 19:
```suggestion
            if (storage[i].getFirst() == key) {
```
What if it's not the same object, but equal to it?

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #197, opened 2020-12-08
implemented list "key - value"

Ilya-brinchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                return values[i];
+            }
+
+        }
         return null;
     }
+
+    private boolean isDuplicate(K key) {
```
KseniiaMakarova, line 42: even when you created the private method, your code still contains repetitive parts

let's refactor this method a bit:
```suggestion
    private int getIndex(K key) {
```
return the index of the matching key, and if there is no match - return `-1`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
+        for (int i = 0; i < keys.length; i++) {
```
KseniiaMakarova, line 43: should we iterate from 0 to 9 here or can we optimise this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_LENGTH_ARRAY = 10;
+    private static int itemCounter = 0;
+    private K[] keys = (K[]) new Object[INITIAL_LENGTH_ARRAY];
+    private V[] values = (V[]) new Object[INITIAL_LENGTH_ARRAY];
```
KseniiaMakarova, line 10: let’s initialise all fields in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,50 @@
...
+                }
+            }
+        } else {
+            keys[itemCounter] = key;
+            values[itemCounter] = value;
+            itemCounter++;
+            K[] cloneKeys = Arrays.copyOf(keys, INITIAL_LENGTH_ARRAY + itemCounter);
+            keys = cloneKeys;
+            V[] cloneValues = Arrays.copyOf(values, INITIAL_LENGTH_ARRAY + itemCounter);
+            values = cloneValues;
```
KseniiaMakarova, line 27: what is the purpose of these 4 lines? we don't have to do this in the current task)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[INITIAL_LENGTH_ARRAY];
+        values = (V[]) new Object[INITIAL_LENGTH_ARRAY];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (getIndex(key) >= 0) {
+            values[getIndex(key)] = value;
```
KseniiaMakarova, line 19: we can save the result of `getIndex(key)` into a local variable, so that we are not calling it twice

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                return values[i];
+            }
+
+        }
```
KseniiaMakarova, line 34: can we use `getIndex()` here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
         return null;
     }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
KseniiaMakarova, line 39: let's iterate over the **filled** part of the array only

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
great job!

PR #200, opened 2020-12-08
storage created, overridden methods put / get

AndriiBardin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_LENGTH];
+        values = new Object[MAX_LENGTH];
+        currentFill = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_LENGTH; i++) {
```
KseniiaMakarova, line 19: should we iterate from 0 to 9 here or can we optimise this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+            if ((keys[i] != null && keys[i].equals(key))) {
```
KseniiaMakarova, line 20: what if both `keys[i]` and `key` are null? we should replace the value as well

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+                values[i] = value;
+            }
+        }
+        keys[currentFill] = key;
+        values[currentFill] = value;
+        currentFill++;
```
KseniiaMakarova, line 26: these  lines are currently executed in any case, no matter if you replaced the value earlier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_LENGTH; i++) {
```
KseniiaMakarova, line 31: should we iterate from 0 to 9 here or can we optimise this?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
AndriiBardin: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
AndriiBardin: done

src/main/java/core/basesyntax/impl/StorageImpl.java
AndriiBardin: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
AndriiBardin: fixed

KseniiaMakarova (approved):
great job!

PR #201, opened 2020-12-08
Added implementation methods put(), get()

sanek98cv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 10;
+    private final K[] keyArray = (K[]) new Object[ARRAY_LENGTH];
+    private final V[] valueArray = (V[]) new Object[ARRAY_LENGTH];
+    private int index = 0;
```
KseniiaMakarova, line 9: let’s initialise all fields in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
...
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
```
KseniiaMakarova, line 13: should we iterate from 0 to 9 here or can we optimise this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,29 @@
...
+            if (keyArray[i] != null && keyArray[i].equals(key) || keyArray[i] == key) {
+                valueArray[i] = value;
+            }
+        }
+        keyArray[index] = key;
+        valueArray[index++] = value;
```
KseniiaMakarova, line 19: these lines are executed always at the moment, no matter if you replaced the value earlier

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,37 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < index; i++) {
+            if (keyArray[i] != null && keyArray[i].equals(key) || keyArray[i] == key) {
+                valueArray[i] = value;
+            }
+        }
+        if (keyArray[index] == null && valueArray[index] == null) {
+            keyArray[index] = key;
+            valueArray[index++] = value;
+        }
```
KseniiaMakarova, line 27: do we need this check here or can we just set values?

let's exit the method on line 22 if we have replaced the value

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < index; i++) {
+            if (keyArray[i] != null && keyArray[i].equals(key) || keyArray[i] == key) {
+                valueArray[i] = value;
+                break;
+            }
+        }
+        keyArray[index] = key;
+        valueArray[index++] = value;
```
KseniiaMakarova, line 26: these lines are still executed even after you break the loop

let's use another keyword instead of `break` that will exit from the method

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
good job!

PR #204, opened 2020-12-08
implement class StorageImpl with 2 methods(put() and get());

mzrkvolodymyr:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,30 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currentLength; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                values[i] = value;
+            }
+        }
+        keys[currentLength] = key;
+        values[currentLength] = value;
+        currentLength++;
```
dmytro-hryhoruk, line 20: you adding a new element anyway even it's already present in the storage

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,30 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAXIMUM_NUMBERS_OF_ELEMENTS; i++) {
```
dmytro-hryhoruk, line 25: why constant but not current length?

dmytro-hryhoruk (changes requested):

dmytro-hryhoruk (approved):

PR #208, opened 2020-12-08
added key-value storage type implementation

oroboro-cpu:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENT_NUMBER = 10;
+    private K key;
+    private V value;
```
KseniiaMakarova, line 8: let's remove these fields - we don't need them

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
...
+            if ((arrayOfKeys[i] == null && key == null)
+                    || (arrayOfKeys[i] != null && arrayOfKeys[i].equals(key))) {
+                arrayOfValues[i] = value;
+                break;
+            }
+            if (arrayOfValues[i] == null && arrayOfKeys[i] == null) {
+                arrayOfKeys[i] = key;
+                arrayOfValues[i] = value;
+                break;
+            }
```
KseniiaMakarova, line 37: can we unite these two checks?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
...
+    }
+
+    public V getValue() {
+        return value;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if ((arrayOfKeys[i] == null && key == null)
```
KseniiaMakarova, line 28:
```suggestion
            if (arrayOfKeys[i] == key
```
it can be simplified like this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,49 @@
...
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if (key != null && key.equals(arrayOfKeys[i])
+                    || (key == null && arrayOfKeys[i] == null)) {
```
KseniiaMakarova, line 45: same
```suggestion
                    || key == arrayOfKeys[i]) {
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if ((arrayOfKeys[i] == key) || (arrayOfKeys[i] != null && arrayOfKeys[i].equals(key))) {
+                arrayOfValues[i] = value;
+                break;
+            } else if (arrayOfValues[i] == null && arrayOfKeys[i] == null) {
+                arrayOfKeys[i] = key;
+                arrayOfValues[i] = value;
+                break;
+            }
```
KseniiaMakarova, line 25: I mean the following:

**start of method**

for loop {
replace key and value;
}
}

**end of method**

what do you think?

KseniiaMakarova (approved):

PR #209, opened 2020-12-09
add correct realised methods put and get for storageImpl

ShemsheyMaxim:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,42 @@
...
+    }
 
     @Override
     public void put(K key, V value) {
+        if (length == 0) {
+            keys[length] = key;
+            values[length] = value;
+            length++;
+            return;
+        }
```
dmytro-hryhoruk, line 23: that's redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,42 @@
...
+        for (int i = 0; i < length; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
dmytro-hryhoruk, line 25: we may have one value by key = null

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ShemsheyMaxim: done

src/main/java/core/basesyntax/impl/StorageImpl.java
ShemsheyMaxim: done

dmytro-hryhoruk (approved):

PR #210, opened 2020-12-09
added solution storage

IgorBochkor:
added solution storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,13 +3,44 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] boxes;
+    private int cellNumber;
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        boxes = new Object[MAX_ITEMS_NUMBER];
+        cellNumber = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < cellNumber; i++) {
+            if (key != null && key.equals(keys[i])) {
+                keys[i] = key;
+                boxes[i] = value;
+            }
+        }
+        keys[cellNumber] = key;
+        boxes[cellNumber] = value;
+        cellNumber++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V box = null;
+        for (int i = 0; i < cellNumber; i++) {
+            if (key == null && keys[i] == null) {
+                box = (V) boxes[i];
+            }
+            if (keys[i] == null) {
+                continue;
+            }
+            if (keys[i].equals(key)) {
+                box = (V) boxes[i];
+            }
+        }
+        return box;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+        keys = new Object[MAX_ITEMS_NUMBER];
+        boxes = new Object[MAX_ITEMS_NUMBER];
+        cellNumber = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < cellNumber; i++) {
+            if (key != null && key.equals(keys[i])) {
+                keys[i] = key;
```
KseniiaMakarova, line 21: what if both `key` and `keys[i]` are null?
also, should we set the new key?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+                boxes[i] = value;
+            }
+        }
+        keys[cellNumber] = key;
+        boxes[cellNumber] = value;
+        cellNumber++;
```
KseniiaMakarova, line 27: these  lines are currently executed in any case, no matter if you replaced the value earlier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,44 @@
...
+        for (int i = 0; i < cellNumber; i++) {
+            if (key == null && keys[i] == null) {
+                box = (V) boxes[i];
+            }
+            if (keys[i] == null) {
+                continue;
+            }
+            if (keys[i].equals(key)) {
+                box = (V) boxes[i];
+            }
```
KseniiaMakarova, line 42: can we write all these checks in a single `if`?
remember how we compared two objects in previous tasks (we compared fields there)

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):
the table was not updated for the first time

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: done

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        boxes = new Object[MAX_ITEMS_NUMBER];
+        cellNumber = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
```
KseniiaMakarova, line 19: let's iterate here over the **filled** part of the array only

and check for keys equality
after this loop, if no key matches are found, we can set key and value on the index that is stored in our `cellNumber` variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+            if (boxes[i] == null && keys[i] == null) {
```
KseniiaMakarova, line 20: then we will not need this check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
+                keys[i] = key;
+                boxes[i] = value;
+                break;
+            }
+            if (key != null && keys[i] != null && key.equals(keys[i])) {
```
KseniiaMakarova, line 25: what if both `key` and `keys[i]` are null? please look at your previous solution (pair) and recall how we compared two objects (there it was object fields)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
...
     @Override
     public V get(K key) {
-        return null;
+        V box = null;
+        for (int i = 0; i < cellNumber; i++) {
+            if (key == null && keys[i] == null) {
+                box = (V) boxes[i];
+            } else if (key != null && key.equals(keys[i])) {
+                box = (V) boxes[i];
+            }
```
KseniiaMakarova, line 41: same here - this call can be written in one `if`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: did in one if

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] boxes;
```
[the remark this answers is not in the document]
IgorBochkor: +

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
...
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+        cellNumber = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < cellNumber; i++) {
+            if (key != null && keys[i] != null && key.equals(keys[i])) {
```
KseniiaMakarova, line 20: if both `keys[i]` and `key` are null - we should replace the value as well (the condition in `if` should return true)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
IgorBochkor: if ((key != null && keys[i] != null && key.equals(keys[i]))
                    || (key == null && keys[i] == null)) {
                values[i] = value;

is this correct?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,39 @@
...
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+        cellNumber = 0;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < cellNumber; i++) {
+            if ((key != null && keys[i] != null && key.equals(keys[i]))
+                    || (key == null && keys[i] == null)) {
```
KseniiaMakarova, line 21:
```suggestion
            if (key != null && key.equals(keys[i]) || key == keys[i]) {
```
great, it may be simplified a little

KseniiaMakarova (approved):
good work!

PR #211, opened 2020-12-09
add all to StorageImpl

vlykhvar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,40 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[STORAGE_CAPACITY];
+        this.values = (V[]) new Object[STORAGE_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < STORAGE_CAPACITY; i++) {
+            if (keys[i] == key) {
```
KseniiaMakarova, line 19: is it the right way to compare two objects?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,40 @@
...
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
```
KseniiaMakarova, line 26: we can unite these two conditions, right?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,40 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < STORAGE_CAPACITY; i++) {
+            if (keys[i] != null && key == null) {
+                continue;
+            } else if (keys[i] == null && key == null) {
+                return values[i];
+            } else if (key.equals(keys[i])) {
+                return values[i];
+            }
```
KseniiaMakarova, line 39: can we write all these checks in a single `if`?
remember how we compared two objects in previous tasks (we compared fields there)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
+                values[i] = value;
+                break;
+            } else if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
```
KseniiaMakarova, line 29: I mean the following:

*start of method*

for loop {
        replace key and value;
    }
}

*end of method*

what do you think?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < STORAGE_CAPACITY; i++) {
+            if (key != null && key.equals(keys[i])) {
+                return values[i];
+            } else if (key == null && keys[i] == null) {
+                return values[i];
+            }
```
KseniiaMakarova, line 40: let's see your previous solution (pair) and write everything in one `if` here

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
great job!

vlykhvar:
> great job!

thks, i almost think to push pc down from balcony

PR #216, opened 2020-12-09
Storage

Baldinini:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[LENGTH_ARRAY];
+        values = (V[]) new Object[LENGTH_ARRAY];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int index = 0; index < LENGTH_ARRAY; index++) {
+            if ((keys[index] != null && keys[index].equals(key)) && values[index] != null) {
```
KseniiaMakarova, line 18: what if `keys[index]` and `key` are both null?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,36 @@
...
+                values[index] = value;
+                break;
+            } else if (keys[index] == null && values[index] == null) {
+                keys[index] = key;
+                values[index] = value;
+                break;
+            }
```
KseniiaMakarova, line 25: we can also unite these two checks like this:

**start of method**
for loop {
replace key and value and break;
}
}
**end of method**
what do you think?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
nice job!

PR #217, opened 2020-12-09
added solution

ivlytvynets:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
...
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < STORAGE_CAPACITY; i++) {
+            if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
+            if (keys[i] == key || key != null && keys[i] == key) {
```
KseniiaMakarova, line 23: review this line please

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,38 @@
...
+                values[i] = value;
+                break;
+            }
```
KseniiaMakarova, line 26: also, let's maybe try to unite the two checks?

**start of method**
for loop {
replace key and value and break;
}
}
**end of method**
what do you think?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):
very nice!

boroda4436 (approved):

YuriiHoliuk (changes requested):
ффф

YuriiHoliuk (approved):

PR #218, opened 2020-12-09
refactored class StorageImpl

Alekearl:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,33 @@
...
+
+    public StorageImpl() {
+        values = (V[]) new Object[INITIAL_ARRAY_LENGTH];
+        keys = (K[]) new Object[INITIAL_ARRAY_LENGTH];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null || keys[i] == key) {
```
KseniiaMakarova, line 18: is `keys[i] == key` the right way to compare objects?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Alekearl: No, equals() does, but it was surprise for me too, that with this condition it shows all tests green, so I left it here.
Changed to equals().

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,13 +3,34 @@
...
+    public StorageImpl() {
+        values = (V[]) new Object[INITIAL_ARRAY_LENGTH];
+        keys = (K[]) new Object[INITIAL_ARRAY_LENGTH];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && values[i] == null
+                    || key != null && key.equals(keys[i])) {
```
KseniiaMakarova, line 19: what if both `key` and `keys[i]` are null?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Alekearl: returns true. ^__^

KseniiaMakarova (approved):
good job!)

PR #220, opened 2020-12-09
1) Commit

KostyantynProskura:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+            }
+        }
+        keys[indexCurrent] = key;
+        values[indexCurrent] = value;
+        indexCurrent++;
```
dmytro-hryhoruk, line 27: these lines execute in any case — either key was found and value replaced or not

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
+                return values[i];
+            } else if (key == null && keys[i] == null) {
+                return values[i];
+            }
+
+        }
         return null;
     }
 }
+
```
dmytro-hryhoruk, line 44:
```suggestion
        for (int i = 0; i < ARRAY_LENGTH; i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                return values[i];
            } else if (key == null && keys[i] == null) {
                return values[i];
            }
        }
        return null;
    }
}
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,41 @@
...
+        indexCurrent++;
     }
 
     @Override
     public V get(K key) {
+
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                return values[i];
+            } else if (key == null && keys[i] == null) {
```
dmytro-hryhoruk, line 36: this check is redundant keys[i] == key will cover this case

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,12 +4,39 @@
...
     public V get(K key) {
+        for (int i = 0; i < indexCurrent; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                return values[i];
+            }
+        }
         return null;
     }
 }
+
```
dmytro-hryhoruk, line 42:
```suggestion
```

dmytro-hryhoruk (approved):

PR #225, opened 2021-03-24
SV solution

SakVladyslav:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,40 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final Object[] keyStorage = new Object[MAX_STORAGE_SIZE];
+    private final Object[] valueStorage = new Object[MAX_STORAGE_SIZE];
+    private int count = 0;
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < count; i++) {
+            if (key == null && keyStorage[i] == null || key != null && key.equals(keyStorage[i])) {
```
RomanDubovskyi, line 14:
```suggestion
            if (key == keyStorage[i] || key != null && key.equals(keyStorage[i])) {
```
you can shorten this condition, because if they are both null '==' comparison is valid. 
BTW:
the best way to write this condition is to use Objects.equals(key, keyStorage[i]). We just forbid it in checklist so you can learn how to write this check manually properly. But if you understand already how to write it with null check you use Obejcts.equals

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,40 @@
...
+        count++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V result = null;
+
+        for (int i = 0; i < count; i++) {
+            if (key == null && keyStorage[i] == null || key != null && key.equals(keyStorage[i])) {
```
RomanDubovskyi, line 30: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,40 @@
...
+
+        keyStorage[count] = key;
+        valueStorage[count] = value;
+        count++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V result = null;
```
RomanDubovskyi, line 27: you don't need this variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,40 @@
...
+
+        for (int i = 0; i < count; i++) {
+            if (key == null && keyStorage[i] == null || key != null && key.equals(keyStorage[i])) {
+                result = (V) valueStorage[i];
```
```suggestion
                return (V) valueStorage[i];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,40 @@
...
+            }
+        }
+
+        return result;
```
RomanDubovskyi, line 35:
```suggestion
        }
        return null;
```

RomanDubovskyi (changes requested):

KseniiaMakarova (approved):

PR #231, opened 2021-04-04
added solution

safordog:


RomanDubovskyi (changes requested):

RomanDubovskyi (changes requested):
test to update doc status

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,56 @@
...
     public int size() {
-        return -1;
+        return indexOfEmptyElement;
+    }
+
+    private boolean equalsKeys(Object newKey, K keyInStorage) {
+        if (newKey == keyInStorage || newKey != null && newKey.equals(keyInStorage)) {
+            return true;
+        }
+        return false;
```
RomanDubovskyi, line 56:
```suggestion
        return newKey == keyInStorage || newKey != null && newKey.equals(keyInStorage));
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private static final int FIRST_INDEX_OF_ARRAY = 0;
+    private int indexOfEmptyElement;
```
RomanDubovskyi, line 8:
```suggestion
    private static final int MAX_ITEMS_NUMBER = 10;
    private int indexOfEmptyElement;
```
you don't need to set it, it's int so it will be set to 0 by default

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #240, opened 2021-04-06
Implemented the storage

cyouffy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,44 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keyArray;
+    private final Object[] valueArray;
+    private int counter;
+
+    public StorageImpl() {
+        keyArray = new Object[MAX_ITEMS_NUMBER];
+        valueArray = new Object[MAX_ITEMS_NUMBER];
+        counter = 0;
```
Max-Pochepets, line 14: no need for this assignment, primitives have default values
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keyArray;
+    private final Object[] valueArray;
```
```suggestion
    private final Object[] keys;
    private final Object[] values;
```

Max-Pochepets (changes requested):
good job

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,36 +4,35 @@
...
         counter++;
     }
 
     @Override
     public V get(K key) {
         for (int i = 0; i < counter; i++) {
-            if (key == null && keyArray[i] == null
-                    || key != null && key.equals(keyArray[i])) {
-                return (V) valueArray[i];
+            if (key == null && keys[i] == null
```
Max-Pochepets, line 33:
```suggestion
            if (key == keys[i]
```
in this case keys will be either identical, either both null

Max-Pochepets (approved):

PR #245, opened 2021-04-06
Frightening solution

LayRipper:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,68 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int KEY_IN_ARRAY = 0;
+    private static final int VALUE_IN_ARRAY = 1;
+    private static final int KEY_VALUE_PAIR_LENGTH = 2;
+
```
Max-Pochepets, line 9: redundant empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,68 @@
...
+    private int arrayLengthCounter = 0;
```
Max-Pochepets, line 10: no need to set primitives with their default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,68 @@
...
+    private Object[][] newArray = new Object[0][0];
```
Max-Pochepets, line 11: it's better to set fields in class constructor. also, let's make two arrays, one for keys and one for values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,68 @@
...
+                }
+            }
+        } else {
+            reassignNewArray();
+            addKeyValuePair(key, value);
+        }
+    }
+
+    private void reassignNewArray() {
+        newArray = new Object[arrayLengthCounter][KEY_VALUE_PAIR_LENGTH];
```
Max-Pochepets, line 43: you should each pair of key-value in your storage. for now you are loosing data by reassigning the array

Max-Pochepets (changes requested):
fix the current suggestions. you'll probably need to change the logic afterwards, but loosing data is not ok

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,68 +3,61 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
-    private static final int KEY_IN_ARRAY = 0;
-    private static final int VALUE_IN_ARRAY = 1;
-    private static final int KEY_VALUE_PAIR_LENGTH = 2;
-
-    private int arrayLengthCounter = 0;
-    private Object[][] newArray = new Object[0][0];
+    private int arrayLengthCounter;
+    private Object[] keyArray = new Object[0];
+    private Object[] valueArray = new Object[0];
```
Max-Pochepets, line 8: it's better to set non-constant fields in the constructor
```suggestion
    private Object[] keys;
    private Object[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,68 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    private static final int KEY_IN_ARRAY = 0;
-    private static final int VALUE_IN_ARRAY = 1;
-    private static final int KEY_VALUE_PAIR_LENGTH = 2;
-
-    private int arrayLengthCounter = 0;
-    private Object[][] newArray = new Object[0][0];
+    private int arrayLengthCounter;
```
Max-Pochepets, line 6:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,68 +3,61 @@
...
+    private Object[] keyArray = new Object[0];
+    private Object[] valueArray = new Object[0];
 
     @Override
     public void put(K key, V value) {
```
Max-Pochepets, line 11: all this can be simplified, without `temp` array to reduce memory cost. think about iterating through `keys` array looking for input `key`. if we found it - rewrite the corresponding value. if not - put the input key/value to the next array cell

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,68 +3,61 @@
...
             }
-        } else {
-            reassignNewArray();
-            addKeyValuePair(key, value);
         }
     }
 
-    private void reassignNewArray() {
-        newArray = new Object[arrayLengthCounter][KEY_VALUE_PAIR_LENGTH];
+    private void createNewArrays() {
```
Max-Pochepets, line 35: then we won't need this and next method

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #258, opened 2021-04-07
Storage first version

Reastador:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,74 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
```
RomanDubovskyi, line 5: This logic may be simplified greatly: 
 - create int size variable and you won't need to iterate storage each time 
 - in put method check if you have already given key (for until i < size). if you don't have it use arrK[size] and arrV[size] size++
 Smt like this

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,60 @@
...
+    private static final int ARRAY_SIZE = 10;
```
RomanDubovskyi, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```
it's not to have 2 variables with the same names

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,60 @@
...
+                }
+            }
+        }
+        return value;
     }
 
     @Override
     public int size() {
-        return -1;
+        int storageSize = 0;
```
RomanDubovskyi, line 52:
```suggestion
       return arraySize;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,60 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        V value = null;
+        boolean check;
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (key == null && valuesArray[i] != null) {
+                check = true;
+            } else {
```
```suggestion
        for (int i = 0; i < arraySize; i++) {
            if (key == keys[i] || key != null && key.equals(valuesArray[i]) {
                return your value 
            } 
```
this check is enough)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,60 @@
...
+                } else if (keyArray[i] == key && valuesArray[i] != null
+                        || keyArray[i].equals(key) && valuesArray[i] != null) {
+                    value = (V) valuesArray[i];
+                    break;
+                } else if (keyArray[i].equals(key) && valuesArray[i] == null) {
+                    return null;
+                }
+            }
+        }
+        return value;
```
RomanDubovskyi, line 47:
```suggestion
        return null;
```

RomanDubovskyi (changes requested):
put is good, rest may be simplified even more

RomanDubovskyi (approved):

PR #261, opened 2021-04-07
C001 - All Done! For Review.

Koziarevskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,64 @@
...
+    private K[] keysStorage;
+    private V[] valuesStorage;
+    private int size;
+
+    public StorageImpl() {
+        this.keysStorage = (K[]) new Object[ARRAY_LENGTH_MAX];
+        this.valuesStorage = (V[]) new Object[ARRAY_LENGTH_MAX];
+    }
+
+    public K[] getKeysStorage() {
```
RomanDubovskyi, line 16: let's remove this method, used don't need to have access directly to keys arr

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,64 @@
...
+        return keysStorage;
+    }
+
+    public void setKeysStorage(K[] keysStorage) {
```
RomanDubovskyi, line 20: same for other getters

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,64 @@
...
+        return valuesStorage;
+    }
+
+    public void setValuesStorage(V[] valuesStorage) {
+        this.valuesStorage = valuesStorage;
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < ARRAY_LENGTH_MAX; i++) {
```
RomanDubovskyi, line 34:
```suggestion
        for (int i = 0; i < size; i++) {
```
we don't need to iterate whole arr

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,64 @@
...
+            if (key == keysStorage[i] && valuesStorage[i] != null
+                    || key != null
+                    && key.equals(keysStorage[i])) {
```
RomanDubovskyi, line 37:
```suggestion
            if (key == keysStorage[i]  || key != null && key.equals(keysStorage[i])) {
```
if we iterate arr from 0 to < size, we will be able to simplify this check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,64 @@
...
+                valuesStorage[i] = value;
+                return;
+            }
+        }
+
+        if (size < ARRAY_LENGTH_MAX) {
```
RomanDubovskyi, line 43: let's remove this if condition. In case arr is full your method will work ok but value won't be saved. For use it's better to receive out of bound exception so he'll know that something went wrong (you don't need throw anything here just remove if condition)

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #262, opened 2021-04-07
First commit task3

latkova:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_STORAGE_NUMBER = 10;
```
RomanDubovskyi, line 6: let's make the variable private, we don't need to access it from outside of the class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,41 @@
...
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_STORAGE_NUMBER];
+        this.values = (V[]) new Object[MAX_STORAGE_NUMBER];
+    }
 
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_STORAGE_NUMBER; i++) {
```
RomanDubovskyi, line 18:
```suggestion
        for (int i = 0; i < size; i++) {
```
we can reduce number of iterations performed

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,41 @@
...
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_STORAGE_NUMBER; i++) {
```
```suggestion
        for (int i = 0; i < size; i++) {
```
no need to go through all values

PR #277, opened 2021-04-08
Add solution

AventusDev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,17 +4,50 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        int indexOfKey = getIndex(key);
+        if (indexOfKey >= 0) {
+            return valuesStorage[indexOfKey];
+        } else {
+            return null;
+        }
```
RomanDubovskyi, line 37:
```suggestion
        } 
       return null;
```
we don't need else after return in if

RomanDubovskyi (approved):

PR #278, opened 2021-04-08
HW-6. Ready for review.

MakeDreamCode:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,47 @@
...
 
     @Override
     public void put(K key, V value) {
+        boolean isFound = false;
+        if (currentSize < MAXIMUM_CAPACITY) {
+            for (int i = 0; i < currentSize; i++) {
+                if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                    isFound = true;
+                }
```
RomanDubovskyi, line 25:
```suggestion
            for (int i = 0; i < currentSize; i++) {
                if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
                    values[i] = value;
                    return;
                }
```
you don't need first if condition. If storage is full for user it's better to receive exception to know that they can't add anymore values. If not they may think that value was saved, but it wasn't

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,47 @@
...
+            }
+            if (!isFound && value != null) {
```
RomanDubovskyi, line 27:
```suggestion
```

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
MakeDreamCode: value != null was removed

src/main/java/core/basesyntax/impl/StorageImpl.java
MakeDreamCode: f (currentSize < MAXIMUM_CAPACITY) replaced by ArrayIndexOutOfBoundsException

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,18 +3,49 @@
...
+                if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                    isFound = true;
+                }
+            }
+            if (!isFound) {
+                keys[currentSize] = key;
+                values[currentSize] = value;
+                currentSize++;
+            }
```
```suggestion
            for (int i = 0; i < currentSize; i++) {
                if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
                    values[i] = value;
                    return;
                }
            }
                keys[currentSize] = key;
                values[currentSize] = value;
                currentSize++;
            }
```

RomanDubovskyi (approved):

RomanDubovskyi (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
MakeDreamCode: снимаю шляпу...  через return  изящней.

PR #287, opened 2021-06-07
Implemented StorageImpl.java, fixed some Code style in test file

aaananas:
Test file caused error while 'mvn clean package'

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,50 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[CAPACITY];
+        values = new Object[CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = 0;
+        for (Object currentKey : keys) {
+            if (key == currentKey || (key != null && key.equals(currentKey))) {
+                values[index] = value;
+                if (size <= index) {
+                    size++;
+                }
+                return;
+            }
+            index++;
+        }
+        if (size != CAPACITY) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = 0;
+        for (Object currentKey : keys) {
+            if (currentKey == key || (key != null && key.equals(currentKey))) {
+                return (V) values[index];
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
```
```suggestion
    private static final int MAXIMUM_CAPACITY = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[CAPACITY];
+        values = (V[]) new Object[CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = 0;
+        for (Object currentKey : keys) {
```
KseniiaMakarova, line 19: let's better use for-i loop and iterate over the filled part only

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+                return;
+            }
+            index++;
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = 0;
+        for (Object currentKey : keys) {
```
KseniiaMakarova, line 36: same here - we can remove `int index` from both methods

KseniiaMakarova (changes requested):
good job!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        for (int i = 0; i < keys.length; ++i) {
+            if (keys[i] == key || (key != null && key.equals(keys[i]))
+                    || (key != null && i == size)) {
+                values[i] = value;
+                if (size <= i) {
+                    keys[i] = key;
+                    size++;
+                }
+                return;
+            }
```
KseniiaMakarova, line 27: this logic is a bit complicated

let's iterate over the filled part of the `keys` only

if no - we will exit the loop - we just add key and value to the end of arrays

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
KseniiaMakarova, line 33: let's iterate over the filled part of the keys only

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #292, opened 2021-06-08
StorageImpl class was implemented. Tests passed.

nikdagaev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keysArray;
+    private Object[] valuesArray;
+
     @Override
     public void put(K key, V value) {
+        if (keysArray == null) {
+            keysArray = new Object[MAX_ITEMS_NUMBER];
+            valuesArray = new Object[MAX_ITEMS_NUMBER];
+        }
```
KseniiaMakarova, line 15: this should be done in the constructor
and we won't need the null check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (keysArray == null) {
+            return null;
+        }
```
KseniiaMakarova, line 30: let's remove this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
-        return -1;
+        if (keysArray == null) {
+            return 0;
+        }
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (valuesArray[i] == null) {
+                return i;
+            }
+        }
+        return MAX_ITEMS_NUMBER;
```
KseniiaMakarova, line 49: let's create a field in this class named `size` that will stand for the filled part of the arrays

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (keysArray == null) {
+            return null;
+        }
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
```
KseniiaMakarova, line 31:
```suggestion
        for (int i = 0; i < size; i++) {
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+                keysArray[i] = key;
+                valuesArray[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size(); i++) {
```
KseniiaMakarova, line 29:
```suggestion
        for (int i = 0; i < size; i++) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
-        return -1;
+        if (keysArray == null) {
+            return 0;
+        }
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (valuesArray[i] == null) {
+                return i;
+            }
+        }
+        return MAX_ITEMS_NUMBER;
```
KseniiaMakarova, line 47: let's simplify this method to 1 line

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+                return (V) valuesArray[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return keysArray != null ? size : 0;
```
KseniiaMakarova, line 41: can it be null?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (valuesArray[i] == null || keysArray[i] == key
+                    || (keysArray[i] != null && keysArray[i].equals(key))) {
+                keysArray[i] = key;
+                valuesArray[i] = value;
+                size = i + 1;
```
KseniiaMakarova, line 23: you should only increment the size if you add new key-value pair, and not just replace the value

in this method let's iterate over the filled part of the `keysArray` only

if no - we will exit the loop - we just add key and value to the end of arrays, and increment size

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keysArray;
+    private Object[] valuesArray;
+    private int size = 0;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        keysArray = new Object[MAX_ITEMS_NUMBER];
+        valuesArray = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keysArray[i] == key
+                    || (keysArray[i] != null && keysArray[i].equals(key))) {
+                keysArray[i] = key;
```
KseniiaMakarova, line 21:
```suggestion
```
there is no need to replace the key

KseniiaMakarova (approved):

PR #293, opened 2021-06-08
hw-7 added solution

OlesiaSmitiukh:
implemented method  put which  put new key-value pair, method get for getting data and mathod size which returns storage size


KseniiaMakarova (approved):
great)

PR #295, opened 2021-06-08
added logic at get(), put(), size()

Tamara-Paskalova:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (V element : valueStorage) {
+            if (element != null) {
+                count++;
+            }
+        }
+        return count;
```
KseniiaMakarova, line 49: let's create a field called `size` in this class
we will increment it if we add a new key-value pair

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+                if (keyStorage[i] == null && valueStorage[i] != null) {
+                    return valueStorage[i];
+                }
+            }
+        }
+        for (int i = 0; i < keyStorage.length; i++) {
+            if (keyStorage[i] != null && keyStorage[i].equals(key)) {
+                index = i;
+            }
+        }
```
KseniiaMakarova, line 37: let's use only 1 loop here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private static final int ARRAY_LENGTH = 10;
+    private final K[] keyStorage = (K[]) new Object[ARRAY_LENGTH];
+    private final V[] valueStorage = (V[]) new Object[ARRAY_LENGTH];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyStorage.length; i++) {
+            if (keyStorage[i] == null && valueStorage[i] == null
+                    || key == null && keyStorage[i] == null && !valueStorage[i].equals(value)
+                    || keyStorage[i] != null && keyStorage[i].equals(key)) {
```
KseniiaMakarova, line 15:
```suggestion
            if (keyStorage[i] == null && valueStorage[i] == null
                    || (key == keyStorage[i] || keyStorage[i] != null && keyStorage[i].equals(key))) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 10;
+    private final K[] keyStorage = (K[]) new Object[ARRAY_LENGTH];
+    private final V[] valueStorage = (V[]) new Object[ARRAY_LENGTH];
```
KseniiaMakarova, line 8: let's initialise the arrays in constructor

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
```
KseniiaMakarova, line 4: please see checklist - we should not use `Objects`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 10;
```
```suggestion
    private static final int MAX_ARRAY_LENGTH = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+    private final K[] keyStorage;
+    private final V[] valueStorage;
+    private int count = 0;
```
```suggestion
    private int count;
```
the default value of int is 0, so we can omit setting it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+                keyStorage[i] = key;
+                valueStorage[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyStorage.length; i++) {
```
KseniiaMakarova, line 35: should we iterate over all 10 elements or only the filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+                keyStorage[i] = key;
+                valueStorage[i] = value;
+                count++;
+                return;
+            } else if (Objects.equals(keyStorage[i], key)) {
+                keyStorage[i] = key;
+                valueStorage[i] = value;
+                return;
+            }
+        }
```
KseniiaMakarova, line 30: this part already looks ok, but we can make it a little better

let's iterate over the filled part of the `keyStorage` only

if no - we will exit the loop - we just add key and value to the end of arrays
(`count` stands for the first unfilled index)

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #296, opened 2021-06-08
add_implementation

Nikitozin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+            }
+        }
+        keys[currentCapacity] = key;
+        values[currentCapacity] = value;
+        currentCapacity++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_CAPACITY; i++) {
```
KseniiaMakarova, line 31: should we iterate over all 10 elements or only the filled part?

KseniiaMakarova (changes requested):
great job, added 1 small comment

KseniiaMakarova (approved):

PR #297, opened 2021-06-08
Create three methods, tests work good

testYurka:
Create three methods, tests work good

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,42 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private int size = 0;
+    private Object [] keys = new Object [CAPACITY];
+    private Object [] values = new Object [CAPACITY];
+
     @Override
     public void put(K key, V value) {
+        int counterSameKeys = 0;
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                counterSameKeys++;
+                values[i] = value;
+            }
+        }
+        if (counterSameKeys == 0) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null) && (key == null)) {
+                return (V) values[i];
+            }
+            if ((keys[i] != null) && (keys[i].equals(key))) {
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private int size = 0;
+    private Object [] keys = new Object [CAPACITY];
+    private Object [] values = new Object [CAPACITY];
```
KseniiaMakarova, line 9: please see checklist - we should initialise non-constant fields in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private int size = 0;
+    private Object [] keys = new Object [CAPACITY];
```
KseniiaMakarova, line 8: don't add whitespaces here
```suggestion
    private Object[] keys = new Object[CAPACITY];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
+    private int size = 0;
```
KseniiaMakarova, line 7: we may not initialise it explicitly - the default value of int is 0
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+    private Object [] keys = new Object [CAPACITY];
+    private Object [] values = new Object [CAPACITY];
+
     @Override
     public void put(K key, V value) {
+        int counterSameKeys = 0;
```
KseniiaMakarova, line 13: can we remove this variable and exit from method after we found the matching key in the loop?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null) && (key == null)) {
+                return (V) values[i];
+            }
+            if ((keys[i] != null) && (keys[i].equals(key))) {
+                return (V) values[i];
+            }
```
KseniiaMakarova, line 35: can we use 1 `if`?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
```
```suggestion
    private static final int MAXIMUM_CAPACITY = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null) && (key == null) || (keys[i] != null) && (keys[i].equals(key))) {
```
KseniiaMakarova, line 32:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

KseniiaMakarova (approved):

PR #298, opened 2021-06-08
Added Storage implementation in StorageImpl, that provides usefull me…

Optipist:
…thods to work with data; Also added StorageIsFullException

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -32,7 +32,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
     @Override
     public V get(K key) {
         index = indexOfKey(key);
-        return (index == -1)? null : values[index];
+        return (index == -1) ? null : values[index];
     }
 
     @Override
@@ -42,7 +42,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     private int indexOfKey(K key) {
         for (int t = 0; t < size; t++) {
-            if ((key == null && keys[t] == null && values[t] !=  null)
+            if ((key == null && keys[t] == null && values[t] != null)
                     || (keys[t] != null && keys[t].equals(key))) {
                 return t;
             }
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -27,6 +27,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
         }
         storage.add(value);
         keys.add(key);
+        size ++;
     }
 
     @Override
@@ -44,6 +45,6 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     @Override
     public int size() {
-        return storage.size();
+        return size;
     }
 }
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -27,7 +27,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
         }
         storage.add(value);
         keys.add(key);
-        size ++;
+        size++;
     }
 
     @Override
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -20,7 +20,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
             throw new StorageIsFullException("Storage is full!");
         }
         index = indexOfKey(key);
-        if (size != 0 && index >= 0) {
+        if (index >= 0) {
             values[index] = value;
             return;
         }
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -5,8 +5,15 @@ import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     private static final int LIST_LIMIT = 10;
-    private ArrayList<V> storage = new ArrayList<>(LIST_LIMIT);
-    private ArrayList<K> keys = new ArrayList<>(LIST_LIMIT);
+    private ArrayList<V> storage;
+    private ArrayList<K> keys;
+    private int size = 0;
+
+    public StorageImpl() {
+        storage = new ArrayList<>(LIST_LIMIT);
+        keys = new ArrayList<>(LIST_LIMIT);
+        size = 0;
+    }
 
     @Override
     public void put(K key, V value) {
@@ -15,8 +22,8 @@ public class StorageImpl<K, V> implements Storage<K, V> {
         }
 
         if (keys.indexOf(key) >= 0) {
-            storage.remove(keys.indexOf(key));
-            keys.remove(keys.indexOf(key));
+            storage.set(keys.indexOf(key), value);
+            return;
         }
         storage.add(value);
         keys.add(key);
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,42 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LIST_LIMIT = 10;
+    private ArrayList<V> storage = new ArrayList<>(LIST_LIMIT);
+    private ArrayList<K> keys = new ArrayList<>(LIST_LIMIT);
```
KseniiaMakarova, line 9: let's initialise these variables in constructor

also, it is great that you are familiar with lists, but it will be better if we use arrays at this point (under the hood arrayList uses an array)

let's create 1 more field named `size` to track the actual size of the filled part of the arrays

src/main/java/core/basesyntax/impl/StorageIsFullException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.impl;
+
+public class StorageIsFullException extends RuntimeException {
+    public StorageIsFullException(String message) {
+        super(message);
+    }
+}
```
KseniiaMakarova, line 7: good job)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,42 @@
...
     @Override
     public void put(K key, V value) {
+        if (storage.size() == LIST_LIMIT) {
+            throw new StorageIsFullException("Storage is full!");
+        }
+
+        if (keys.indexOf(key) >= 0) {
+            storage.remove(keys.indexOf(key));
+            keys.remove(keys.indexOf(key));
+        }
```
KseniiaMakarova, line 20: if we find the matching key, we should replace the value, and not remove it and add to the end

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):
not all comments are fixed - check them once more please

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LIST_LIMIT = 10;
+    private V[] storage;
+    private K[] keys;
+    private int size;
+
+    public StorageImpl() {
+        storage = (V[]) new Object[LIST_LIMIT];
+        keys = (K[]) new Object[LIST_LIMIT];
+        size = 0;
```
KseniiaMakarova, line 14:
```suggestion
```
int is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public int indexOf(Object object, Object[] objects) {
```
KseniiaMakarova, line 49: let's better make this method private

also,
 - do we need to pass `objects`?
 - can we use K instead of `Object` type?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LIST_LIMIT = 10;
+    private V[] storage;
```
```suggestion
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+    public int indexOf(Object object, Object[] objects) {
+        for (int t = 0; t < objects.length; t++) {
+            if (object == null && objects[t] == null) {
+                if (storage[t] != null) {
+                    return t;
+                }
+            }
+            if ((objects[t] != null && objects[t].equals(object))) {
+                return t;
+            }
```
KseniiaMakarova, line 58: let's use only 1 `if`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public int indexOf(Object object, Object[] objects) {
+        for (int t = 0; t < objects.length; t++) {
```
KseniiaMakarova, line 50:
```suggestion
        for (int i = 0; i < size; i++) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
     @Override
     public V get(K key) {
+        for (int t = 0; t < size; t++) {
+            if (key == null) {
+                return storage[indexOf(null, keys)];
+            }
+            if (key.equals(keys[t])) {
+                return storage[t];
+            }
+        }
```
KseniiaMakarova, line 40: there is no need to write loop here
you can call your `indexOf()` method, and return null if index = -1
or value under this index

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == LIST_LIMIT) {
+            throw new StorageIsFullException("Storage is full!");
+        }
+        if (size != 0 && indexOf(key, keys) >= 0) {
+            storage[indexOf(key, keys)] = value;
```
KseniiaMakarova, line 23: save the index into a variable and reuse it

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LIST_LIMIT = 10;
+    private V[] values;
+    private K[] keys;
+    private int size;
+    private int index;
```
KseniiaMakarova, line 10: let's make this variable local in both methods - it doesn't belong to the class conceptually (in does not describe the property of Storage)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        keys = (K[]) new Object[LIST_LIMIT];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == LIST_LIMIT) {
+            throw new StorageIsFullException("Storage is full!");
+        }
+        index = indexOfKey(key);
+        if (size != 0 && index >= 0) {
```
KseniiaMakarova, line 23:
```suggestion
        if (index >= 0) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOfKey(K key) {
+        for (int t = 0; t < size; t++) {
+            if ((key == null && keys[t] == null && values[t] != null)
+                    || (keys[t] != null && keys[t].equals(key))) {
```
KseniiaMakarova, line 46:
```suggestion
            if (key == keys[t] || keys[t] != null && keys[t].equals(key)) {
```

KseniiaMakarova (changes requested):

PR #299, opened 2021-06-08
Implemented get(), put(), getSize() in StorageImpl class

EugeneSinica:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
KseniiaMakarova, line 6: redundant empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+    private static final int SIZE = 10;
```
```suggestion
    private static final int MAXIMUM_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+    private final Object[] keys = new Object[SIZE];
+    private final Object[] values = new Object[SIZE];
```
KseniiaMakarova, line 9: please see checklist - we should initialise arrays in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+    private int countOfPairs = 0;
```
```suggestion
    private int countOfPairs;
```
0 is the default value, so we don't need to set it explicitly

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+        }
+        keys[countOfPairs] = key;
+        values[countOfPairs] = value;
+        countOfPairs++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        for (int i = 0; i < SIZE; i++) {
```
KseniiaMakarova, line 27: should we iterate over all 10 elements or the filled part only?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                return (V) values[i];
+            }
+
+        }
+        return (V) values[countOfPairs];
```
KseniiaMakarova, line 33:
```suggestion
        return null;
```

KseniiaMakarova (changes requested):
overall nice job!

KseniiaMakarova (approved):

PR #304, opened 2021-06-09
HW-7 add class StorageImpl

docer95:
and Override method put/get/size


src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,42 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH_ARRAY = 10;
+    private K[] keysWithStorage;
+    private V[] valueWithStorage;
+    private int lengthStorage = 0;
```
KseniiaMakarova, line 7:
```suggestion
    private K[] keys;
    private V[] value;
    private int currentLength;
```
the naming can be a little bit more consise
and we don't need to set 0, this is a default value

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public V get(K key) {
+        for (int i = 0; i < lengthStorage; i++) {
+            if (key != null && key.equals(keysWithStorage[i]) || keysWithStorage[i] == key) {
+                return valueWithStorage[i];
+            }
+        }
+        System.out.println("Sorry, this key absent in Storage list");
```
KseniiaMakarova, line 34: let's better remove this sout

KseniiaMakarova (changes requested):
good job! added minor comments

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -2,41 +2,40 @@
...
     }
 
     @Override
     public void put(K key, V value) {
-        for (int i = 0; i < lengthStorage; i++) {
-            if (key != null && key.equals(keysWithStorage[i]) || keysWithStorage[i] == key) {
-                valueWithStorage[i] = value;
+        for (int i = 0; i < currentLength; i++) {
+            if (key != null && key.equals(keys[i]) || keys[i] == key) {
+                this.value[i] = value;
```
KseniiaMakarova, line 18: why this?

docer95: I used Shift + F6 - rename all variables
Apparently IntelliJ set the this by default, I did not notice

PR #330, opened 2021-08-22
parametrized key-value storage implemented

lavrivskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUES_QUANTITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int putIndex;
```
boroda4436, line 9:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+        }
+        keys[putIndex] = key;
+        values[putIndex] = value;
+        putIndex++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_VALUES_QUANTITY; i++) {
```
boroda4436, line 36: Maybe it's enough to iterate only over the known count of elements, not over all array capacity?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (V value : values) {
+            size += value != null ? 1 : 0;
+        }
+        return size;
```
boroda4436, line 50: please simplify this method. It's enough to return size
```suggestion
        return size;
```

boroda4436 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        int valueIndex = 0;
+        for (K k : keys) {
```
boroda4436, line 37: you can use `for i` loop and in this case you are not required to have `valueIndex` variable

boroda4436 (changes requested):

Iryna-U (approved):

PR #331, opened 2021-08-22
added solution

konstde00:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+        int index = getIndex(key);
+        return !checkIndex(index) ? null : (V) values[index];
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public boolean checkIndex(int index) {
```
boroda4436, line 43:
```suggestion
    private boolean isValidIndex(int index) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+        if (index == -1) {
```
boroda4436, line 44: What if index = -2?

boroda4436: what if index = 100 and keys.length = 10?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    private void resize() {
+        Object[] newKeys = new Object[size + MAX_LENGTH];
```
boroda4436, line 51: If you have `MAX_LENGTH` how it is possible to have a size of the array greater than MAX_LENGTH?
Better rename your constant if you want have array resizing.

src/main/java/core/basesyntax/impl/StorageImpl.java
boroda4436: https://stackoverflow.com/a/3874378

boroda4436: Also, should this method be public?

boroda4436 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+            newKeys[i] = keys[i];
+            newValues[i] = values[i];
+        }
+        keys = newKeys;
+        values = newValues;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key != null && key.equals(keys[i]) || (key == keys[i])) {
```
boroda4436, line 64:
```suggestion
            if ((key == keys[i]) || (key != null && key.equals(keys[i]))) {
```
== if faster than `equals` so it could be used first

boroda4436 (changes requested):

boroda4436 (approved):
Great job!

PR #339, opened 2021-08-24
All necessary methods such as put() and get() been added

denyssmei:


Iryna-U (changes requested):
In this solution we should use only arrays, please rework your code

MaliukDaria (changes requested):
Travis failed... let's fix this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    private final V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[ARRAY_MAX_SIZE];
+        values = (V[]) new Object[ARRAY_MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
```
MaliukDaria, line 18: can we have key-value pair with null value?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+    private static final int ZERO_LENGTH = 0;
```
MaliukDaria, line 7: do we need this constant?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    private final K[] keys;
+    private final V[] values;
+
```
MaliukDaria, line 10: lets add `size` variable to StorageImpl class. Than we can just increase it when we add element to the storage

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+                break;
+            }
+        }
+        return lengthIterator;
+    }
+
+    private int indexOfKey(K key) {
+        int storageSize = size();
+        for (int i = 0; i < storageSize; i++) {
+            if (((keys[i] == null) && (key == null))
```
MaliukDaria, line 51: do we need this check `key == null` ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+                    || ((keys[i] != null) && keys[i].equals(key))) {
+                return i;
+            }
+        }
         return -1;
```
MaliukDaria, line 56: lets move it to constant

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
denyssmei: Yes we need it if we want to use null as key

src/main/java/core/basesyntax/impl/StorageImpl.java
denyssmei: In such way I`m trying to figure out if suck key exist in our Storage or we shoud add it. But why we can`t have null as an element?

src/main/java/core/basesyntax/impl/StorageImpl.java
denyssmei: There is the tast for null as key

MaliukDaria (changes requested):
Travis failed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int INDEX_FOR_NOT_FOUND_ELEMENT = -1;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_CAPACITY];
+        values = (V[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
MaliukDaria, line 16: we don't need to initialize `size` variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexOfKey = indexOfKey(key);
+        if (indexOfKey == INDEX_FOR_NOT_FOUND_ELEMENT) {
+            int storageSize = size();
```
MaliukDaria, line 23: do we need `storageSize` variable?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
+
+    private int indexOfKey(K key) {
+        int storageSize = size();
```
MaliukDaria, line 47: again, do we need `storageSize` variable?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+        for (int i = 0; i < storageSize; i++) {
+            if ((Objects.equals(keys[i],key))
```
MaliukDaria, line 49: checklist

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #340, opened 2021-08-24
Generics

pustoj0:
added field size and constant MAX_ITEMS_NUMBER;
added two arrays keys(<K>) and values(<V>),
that were initialized in constructor;
implemented methods put, get, and size;
for convenience created method checkEquals that checks keys
equality.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,47 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        //  Returning null from a method is a bad practice говороли вони ;D
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean checkEquals(K key1, K key2) {
```
Iryna-U, line 46: Also, should this method be public?
https://stackoverflow.com/a/3874378
```suggestion
     private boolean checkEquals(K key1, K key2) {
```

Iryna-U (changes requested):
Good job!

PR #355, opened 2021-08-27
jv-storage-solution

MiaHrynda:
…ch casted to K[] and V[].Implemented isKeyTheSame(), put(), get(), size() methods. In isKeyTheSame() implemented the logic of key comparison. In put() implemented the logic of putting values by keys. In get() implemented the logic of getting values by keys.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -42,6 +42,6 @@ public class StorageImpl<K, V> implements Storage<K, V> {
     }
 
     private boolean isKeyTheSame(K arrayKey, K key) {
-        return (arrayKey == key || arrayKey != null && arrayKey.equals(key)) ? true : false;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean isKeyTheSame(K arrayKey, K key) {
+        return (arrayKey == key || arrayKey != null && arrayKey.equals(key)) ? true : false;
```
MaliukDaria, line 45:
```suggestion
        return (arrayKey == key || (arrayKey != null && arrayKey.equals(key)));
```

MaliukDaria (changes requested):

MaliukDaria (approved):
Good Job!

PR #356, opened 2021-08-27
Storage implemented

Nazar-Hetun:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private int sizeExceptNull = 1; //two values that need to count size
+    private int isNullKey = - 1;
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            valueStorage[NULL_KEY] = value;
+            isNullKey = 0;
+            return;
+        }
```
MaliukDaria, line 18: Why do you add null keys at first position in the Storage? Let's behave with them in the same way as with non-zero key values. THen we don't need `NULL_KEY`, `sizeExceptNull` and `isNullKey` variables, and we will have just `size` variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int NULL_KEY = 0; // place 0 in valueStorage array use if key == null.
+    private V[] valueStorage = (V[]) new Object[10];
+    private K[] keyStorage = (K[]) new Object[10];
```
MaliukDaria, line 8: checklist : "All non-constant fields should be initialized in the constructor" https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_OBJECTS = 10;
```
MaliukDaria, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    private V[] valueStorage;
+    private K[] keyStorage;
+    private int size = 0;
```
MaliukDaria, line 9: we dont need to initialize `size` variable explicitly

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+
+    public StorageImpl() {
+        this.valueStorage = (V[]) new Object[MAX_OBJECTS];
+        this.keyStorage = (K[]) new Object[MAX_OBJECTS];
```
MaliukDaria, line 13: we don't need `this` keyword here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+            if (key == null) {
+                if (keyStorage[i] == null) {
+                    valueStorage[i] = value;
+                    return;
+                }
+            } else if (key.equals(keyStorage[i])) {
+                keyStorage[i] = key;
+                valueStorage[i] = value;
+                return;
+            }
```
MaliukDaria, line 28:
```suggestion
            if (keyStorage[i] == key || (keyStorage[i] != null && keyStorage[i].equals(key))) {
                valueStorage[i] = value;
                return; 
            }
            
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (keyStorage[i] == null) {
+                    return valueStorage[i];
+                }
+            } else if (key.equals(keyStorage[i])) {
+                return valueStorage[i];
+            }
```
MaliukDaria, line 44: You have two if conditions that return `valueStorage[i]` in both cases, can we combine them?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Nazar-Hetun: Ого, я і не знав що так можливо було скоротити))

MaliukDaria (approved):

PR #358, opened 2021-08-28
task complete: program work normally.

osa240:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,76 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEM_ELEMENTS = 10;
+    private K key;
+    private V value;
+    private int size;
+    private StorageImpl<K,V>[] data;
```
Max-Pochepets, line 10: this will go into infinity. let's just have two arrays containing keys / values

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -6,17 +6,21 @@
     private static final int MAX_ITEM_ELEMENTS = 10;
     private K key;
     private V value;
```
Max-Pochepets, line 8: we don't need these anymore

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -6,17 +6,21 @@
...
     private int size;
-    private StorageImpl<K,V>[] data;
 
     public StorageImpl() {
-        data = new StorageImpl[MAX_ITEM_ELEMENTS];
+        keyStorage = (K[]) new Object[MAX_ITEM_ELEMENTS];
+        valueStorage = (V[]) new Object[MAX_ITEM_ELEMENTS];
+        key = null;
+        value = null;
         size = 0;
```
Max-Pochepets, line 18: no need to set primitives. they have default values

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
osa240: fix. Like this one?

Max-Pochepets (approved):
👍

PR #359, opened 2021-08-28
realized all methods

dogovor:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    public static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+
```
MaliukDaria, line 15: we don't need to initialize `size` variable explicitly

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size < MAX_SIZE) {
+            for (int i = 0; i < size; i++) {
+                if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                    values[i] = value;
+                    size--;
```
MaliukDaria, line 24: why `size--` ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size < MAX_SIZE) {
```
MaliukDaria, line 20: What if size > MAX_SIZE or size == MAX_SIZE? Then you just not adding pair to your storage, and you do not notify the user about it in any way... If you want do add this check, let's throw some Runtime exception with some informative message

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (values[i] == null) {
+                return i;
+            }
+        }
+        return 0;
```
MaliukDaria, line 50:
```suggestion
        return size;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+
```
MaliukDaria, line 14: redundant empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
MaliukDaria: You have already checked storage size, we don't need this check `if (size < MAX_SIZE)`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_SIZE; i++) {
```
MaliukDaria, line 37: Maybe it's enough to iterate only over the known count of elements, not over all array capacity?

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #360, opened 2021-08-29
Hw-storage-rostyslav

RostyslavF5R:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ARRAY_SIZE];
+        values = (V[]) new Object[MAX_ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Max-Pochepets, line 18: what if size is already 10?

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #363, opened 2021-08-29
storage final realisation

VitaliiMYKH:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+    private K[] key;
+    private V[] value;
+    private int size;
+
+    public StorageImpl() {
+        value = (V[]) new Object[MAX_ITEMS_NUMBER];
+        key = (K[]) new Object[MAX_ITEMS_NUMBER];
+        size = 0;
+
+    }
```
Max-Pochepets, line 16:
```suggestion
    }
```
no need for this, primitives have default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private K[] key;
+    private V[] value;
```
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+    private int size;
+
+    public StorageImpl() {
+        value = (V[]) new Object[MAX_ITEMS_NUMBER];
+        key = (K[]) new Object[MAX_ITEMS_NUMBER];
+        size = 0;
+
+    }
+
+    public K[] getKey() {
```
Max-Pochepets, line 18: we don't need getters here, because we're loosing the point of `get()` method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+        return value;
+    }
+
+    public void setValue(V[] value) {
+        this.value = value;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Max-Pochepets, line 36: what if we'll have 10 items in storage and gonna put one more? everything will be okay?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+            if (this.key[i] == key || this. key[i] != null && this.key[i].equals(key)) {
```
Max-Pochepets, line 37:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -33,21 +23,24 @@ public void setValue(V[] value) {
 
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_ITEMS_NUMBER) {
+            return;
```
Max-Pochepets, line 27: let's throw an exception. because this looks like data leak

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #367, opened 2021-08-29
jv-storage task completed according requirements

Gleb-Kreshchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS = 10;
+    private int index = 0;
+    private K[] keys = (K[]) new Object[MAX_ITEMS];
+    private V[] values = (V[]) new Object[MAX_ITEMS];
```
MaliukDaria, line 9: Lets initialize this variables in constructor (checklist: https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html )

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 
     @Override
     public V get(K key) {
+        int localIndex = 0;
+        for (K localKey : keys) {
+            if (checkIfEqual(key, localKey)) {
+                return values[localIndex];
+            }
+            localIndex++;
+        }
```
MaliukDaria, line 32: maybe it's better to use `for(int i = 0;..` instead of `foreach`, then we won't need `localIndex` variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return index;
+    }
+
+    private boolean checkIfEqual(K newKey, K localKey) {
+        return (localKey == newKey || localKey != null && localKey.equals(newKey));
```
MaliukDaria, line 42:
```suggestion
        return (localKey == newKey || (localKey != null && localKey.equals(newKey)));
```

It seems to me in this case, it is better to explicitly place the parentheses

MaliukDaria (changes requested):

Gleb-Kreshchuk:
Done as requested

MaliukDaria (approved):
🔥

PR #370, opened 2021-08-30
The StorageImpl class was implemented with all the conditions from the "README.md" file

dandalb:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
```
MaliukDaria, line 18:
```suggestion
        for (int i = 0; i < size; i++) {
```

Let's go over only the elements existing in the storage, than we don't need to check values[i] for null value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_SIZE; i++) {
```
MaliukDaria, line 35: same here let's go over only the elements existing in the storage

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #378, opened 2021-09-21
Wrote an assignment jv-storage.

ASelivonchyk:
Implement Storage of key-value by creating class KeyValuePair.
Test end with Error: "cannot invoke ... because of null". I can't solve the problem.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,71 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private KayValePair<K, V>[] storage = new KayValePair[MAX_STORAGE_SIZE];
+    private int currentSize = 0;
+
     @Override
     public void put(K key, V value) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,71 @@
...
+
+    private boolean isEqual(Object firstObject, Object secondObject) {
+        if (firstObject == secondObject
+                || firstObject != null && firstObject.equals(secondObject)) {
+            return true;
+        }
+        return false;
+    }
+
+    private static class KayValePair<K, V> {
```
RomanDubovskyi, line 48:
```suggestion
    private static class KeyValePair<K, V> {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,71 @@
...
+                return;
+            }
+        }
+        storage[currentSize] = new KayValePair<>(key, value);
+        currentSize++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < storage.length; i++) {
```
RomanDubovskyi, line 27: we can iterate our array until `< currentSize` and we won't need  storage[i] != null in if

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #380, opened 2021-10-01
created class KeyValue for storing data in array, implemented methods…

mykolaskrypka:
… get, put, size for working with storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,52 @@
 package core.basesyntax.impl;
 
+import core.basesyntax.KeyValue;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final Integer storageSize = 10;
+    private final KeyValue<K,V>[] storage = new KeyValue[storageSize];
+
     @Override
     public void put(K key, V value) {
+        for (int ind = 0; ind < storageSize; ind++) {
+            if (storage[ind] == null) {
+                KeyValue<K, V> newKeyValue = new KeyValue<>(key, value);
+                storage[ind] = newKeyValue;
+                break;
+            } else if (storage[ind].getKey() == key ||
+                    (storage[ind].getKey() != null && storage[ind].getKey().equals(key))) {
+                storage[ind].setValue(value);
+                break;
+            }
+        }
+
     }
 
     @Override
     public V get(K key) {
+        for (KeyValue<K, V> kv : storage) {
+            if (kv == null) {
+                continue;
+            }
+            if (kv.getKey() == key) {
+                return kv.getValue();
+            }
+            if (kv.getKey() != null && kv.getKey().equals(key)) {
+                return kv.getValue();
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (KeyValue<K, V> kv : storage) {
+            if (kv != null) {
+                size++;
+            }
+        }
+        return size;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
 package core.basesyntax.impl;
 
+import core.basesyntax.KeyValue;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final Integer storageSize = 10;
```
RomanDubovskyi, line 7:
```suggestion
    private static final Integer MAX_SIZE = 10;
```
let's make this field constant it shouldn't change

RomanDubovskyi, line 7:
```suggestion
    private final int storageSize = 10;
```
we can use primitive here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+            }
+            if (kv.getKey() != null && kv.getKey().equals(key)) {
+                return kv.getValue();
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
```
RomanDubovskyi, line 43: not to iterate arr each time let's create size field that we will increment after putting value

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+                break;
+            }
+        }
+
     }
 
     @Override
     public V get(K key) {
+        for (KeyValue<K, V> kv : storage) {
+            if (kv == null) {
```
RomanDubovskyi, line 31: we can remove this if construtction

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+                continue;
+            }
+            if (kv.getKey() == key) {
```
RomanDubovskyi, line 34: let's combine this and following if statement into single one

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size = 0;
+    private final KeyValue<K, V>[] storage = new KeyValue[MAX_SIZE];
+
     @Override
     public void put(K key, V value) {
+        for (int ind = 0; ind < MAX_SIZE; ind++) {
```
RomanDubovskyi, line 13: you can iterate this list till size instead off MAZ_SIZE, and we would be able to remove this if storage[ind] == null

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+                storage[ind].setValue(value);
+                keyIsInArray = true;
+                break;
+            }
+        }
+        if (!keyIsInArray) {
+            KeyValue<K, V> newKeyValue = new KeyValue<>(key, value);
+            storage[size] = newKeyValue;
+            size++;
+        }
```
RomanDubovskyi, line 26:
```suggestion
        for (int ind = 0; ind < size; ind++) {
            if (storage[ind].getKey() == key
                    || (storage[ind].getKey() != null && storage[ind].getKey().equals(key))) {
                storage[ind].setValue(value);
                return;
            }
        }
            KeyValue<K, V> newKeyValue = new KeyValue<>(key, value);
            storage[size] = newKeyValue;
            size++;
```

RomanDubovskyi (changes requested):
left last comment

RomanDubovskyi (approved):

PR #410, opened 2021-10-29
implemented jv-storage

Nata-Vidzivashets:


src/main/java/core/basesyntax/impl/Element.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax.impl;
+
+class Element<K, V> {
```
Nick97-git, line 3:
```suggestion
public class Element<K, V> {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private Element<K, V>[] elements;
+    private int size;
+
+    public StorageImpl() {
+        elements = new Element[MAX_ELEMENTS_NUMBER];
+        size = 0;
```
Nick97-git, line 12: we don't need to initialize this variable explicitly because it already has this default value without it, so let's remove this line)

Nick97-git (changes requested):

Nick97-git:
Good job) But there are minor fixes that you need to do

Nick97-git (approved):

PR #412, opened 2021-10-29
added solution

marchenko-vitalii:


src/main/java/core/basesyntax/Box.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+public class Box {
```
Sofasmile, line 3: do we need this class?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size = 0;
```
Sofasmile, line 9: do we need to set int to 0?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == MAX_ITEMS_NUMBER) {
```
Sofasmile, line 18: we can delete this check because we won't have more than 10 elements.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+        }
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
+        if (key != null) {
```
Sofasmile, line 30: let's combine 2 ifs

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        this.size++;
```
Sofasmile, line 40: when should we use `this` keyword?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            return getWhenKeyIsNull();
+        } else {
+            return getWhenKeyIsNotNull(key);
```
Sofasmile, line 48: these methods are the same, try to combine them

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return this.size;
```
Sofasmile, line 72: same about `this`

Sofasmile (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
```
Sofasmile, line 32: we do the same in `if` and `else` blocks, let's combine them and create one if block instead of two

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
+                    continue;
+                }
+                if (keys[i].equals(key)) {
+                    return (V) values[i];
+                }
+            }
+        }
```
Sofasmile, line 55: same here) let's create on if block and remove else))

Sofasmile (changes requested):

Sofasmile (changes requested):

PR #414, opened 2021-10-29
homework solution

totallyDev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
     @Override
     public int size() {
-        return -1;
+        int countSize = 0;
+        for (Object value : values) {
+            if (value != null) {
+                countSize += 1;
+            }
+        }
+        return countSize;
```
Nick97-git, line 63: you don't need this code, let's create a global variable `size` and return it here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
     public V get(K key) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] == null && keys[i] == key && values[i] != null) {
+                return (V) values[i];
+            }
+            if (keys[i] != null && values[i] != null) {
+                if (keys[i].equals(key)) {
+                    return (V) values[i];
+                }
+            }
```
Nick97-git, line 50: let's use one if instead of two, for this purpose remember the answer for this question from your last lecture `How should we compare 2 objects a and b, given that 2 nulls should be considered equal?`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+                    values[i] = value;
+                    return;
+                }
+            }
+            if (keys[i] == null && values[i] == null) {
+                values[i] = value;
+                keys[i] = key;
+                return;
+            }
+        }
```
Nick97-git, line 37: you can simplify all this code, using only one for loop, and if in it, again answer the question that I mentioned in the next comment

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,54 @@
 
 import core.basesyntax.Storage;
 
+import java.util.Objects;
+
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private int size = 0;
```
Nick97-git, line 9:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,54 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (key == keys[i] && values[i] != key) {
+                return (V) values[i];
+            }
+            if (keys[i] != values[i] && Objects.equals(key, keys[i])) {
+                return (V) values[i];
+            }
```
Nick97-git, line 46: simplify this code, we can use only one if condition here instead of two

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,54 @@
...
+                return;
+            }
+            if (key == keys[i] && values[i] != key) {
+                values[i] = value;
+                return;
+            }
+            if (Objects.equals(key, keys[i])) {
+                values[i] = value;
+                return;
+            }
```
Nick97-git, line 34: simplify this code, we can use only one if condition here instead of three + you can't use `Objects.equals()` read checklist again

Nick97-git (changes requested):

Nick97-git:
Also, Travis's build must be successful

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] == key && values[i] != key || key != null && key.equals(keys[i])) {
```
Nick97-git, line 19:
```suggestion
            if (keys[i] == key || key != null && key.equals(keys[i])) {
```

Nick97-git (changes requested):

Nick97-git (approved):

PR #415, opened 2021-10-30
StorageImpl class was implemented

MykolaVoronov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_STORAGE_SIZE = 10;
+    private Object[] key;
+    private Object[] value;
```
Nick97-git, line 8:
```suggestion
    private Object[] keys;
    private Object[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+    public StorageImpl() {
+        this.key = new Object[MAX_STORAGE_SIZE];
+        this.value = new Object[MAX_STORAGE_SIZE];
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (this.key[i] == key || this.key[i] != null && this.key[i].equals(key)) {
```
Nick97-git, line 20: you don't need to use `this` keyword in this class anywhere, especially after renaming arrays

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_STORAGE_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_STORAGE_SIZE];
+        values = new Object[MAX_STORAGE_SIZE];
+        size = 0;
```
Nick97-git, line 14: we don't need to initialize this variable explicitly because it already has this default value without it, so let's remove this line)

Nick97-git (changes requested):

Nick97-git (approved):

PR #417, opened 2021-10-31
Put and get

Anatolii-Huryk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keyArray;
+    private Object[] valueArray;
+    private int size;
+
+    {
+        keyArray = new Object[MAX_SIZE];
+        valueArray = new Object[MAX_SIZE];
+        size = 0;
```
Nick97-git, line 14: we don't need to initialize this variable explicitly because it already has this default value without it, so let's remove this line)

Nick97-git (changes requested):

Nick97-git (approved):

PR #455, opened 2022-01-01
Hw storage refact

maxkmaxk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY = 10;
```
MaliukDaria, line 6:
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+    private int size = 0;
```
MaliukDaria, line 7:
```suggestion
    private int size;
```

It's initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+        return null;
+    }
+
+    private Element getElement(K key) {
+        for (int i = 0; i < size; i++) {
+            Object bufObj = array[i];
+            if (bufObj == null) {
+                continue;
+            }
+            Element element = (Element) bufObj;
```
MaliukDaria, line 62:
```suggestion
        for (int i = 0; i < size; i++) {
            Object bufObj = array[i];
            Element element = (Element) bufObj;
```

I don't think we need this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+    private boolean keyEquals(Element element, K key) {
+        K thisKey = element.getKey();
+        return thisKey == null ? key == null : thisKey.equals(key);
+    }
+
     @Override
     public int size() {
-        return -1;
+        return size;
     }
```
MaliukDaria, line 78: We should place public methods above private

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+        if (element == null) {
+            if (size >= CAPACITY) {
+                throw new RuntimeException("Couldn't add new elements to storage");
+            }
+            element = new Element(key, value);
+            array[size] = element;
+            size++;
+            return;
+        }
+        element.setValue(value);
```
MaliukDaria, line 44:
```suggestion
            if (size >= CAPACITY) {
                throw new RuntimeException("Couldn't add new elements to storage, storage is full");
            }
        Element element = getElement(key);
        if (element == null) {
            element = new Element(key, value);
            array[size] = element;
            size++;
            return;
        }
        element.setValue(value);
```

lets first check storage size

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
maxkmaxk: But, in your code exception will be thrown, even no need to add new element (just rewrite existing one).

MaliukDaria: agree, my bad

MaliukDaria (approved):

PR #462, opened 2022-01-21
HW develop merge

LinkClink:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int currentCount = 0;
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
```
Max-Pochepets, line 10: do we need to initialize primitives? does `int` have default value?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int currentCount = 0;
```
Max-Pochepets, line 7: we don't need this variable. we can operate `size`, it's enough

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #465, opened 2022-01-25
Implemented Storage key-value

pikador10:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE_OF_ARRAY = 10;
+    private final K[] keys = (K[]) new Object[SIZE_OF_ARRAY];
+    private final V[] values = (V[]) new Object[SIZE_OF_ARRAY];
```
Max-Pochepets, line 8: [Initialization in constructor](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
+    private int elementsAmount;
+    private int valuesCount;
```
Max-Pochepets, line 10:
```suggestion
    private int size;
```
no need for `valuesCount`

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
pikador10: added init in counstructor

src/main/java/core/basesyntax/impl/StorageImpl.java
pikador10: removed valuesCount

Max-Pochepets (approved):

PR #468, opened 2022-01-26
storage of key-value was implemented

sviat-h:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_CAPACITY];
+        values = (V[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
Max-Pochepets, line 14: do we need to set primitives with their default values?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null && values[i] != null) {
+                return values[i];
+            } else if (key != null && key.equals(keys[i])) {
+                return values[i];
+            }
```
Max-Pochepets, line 36: can we combine these two ifs?

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #469, opened 2022-01-27
Implemented the storage of key-value type

dima-semeniuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K[] key;
+    private V[] value;
```
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+        key = (K[]) new Object[STORAGE_SIZE];
+        value = (V[]) new Object[STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            this.key[size] = key;
+            this.value[size] = value;
+            size++;
```
Max-Pochepets, line 21:
```suggestion
            key[size] = key;
            value[size] = value;
            size++;
```
according to last comment, you don't need `this` keyword at all

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #470, opened 2022-01-27
hw7 was solved

AprioriKS:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     @Override
-    public int size() {
-        return -1;
+    public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
```
Max-Pochepets, line 38: we should substitute value for specific key, and it will not always be on the last position

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -31,11 +31,11 @@ public V get(K key) {
...
         if (get(key) == null) {
-            keys[size] = key;
-            values[size] = value;
-            size++;
+            keys[amount] = key;
+            values[amount] = value;
+            amount++;
         } else {
-            values[size - 1] = value;
+            values[amount - 1] = value;
```
Max-Pochepets, line 38: changing variable name is not what I meant. you need a method to get needed index to know exactly what to change, not simply put it at the end

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+            keys[amount] = key;
+            values[amount] = value;
+            amount++;
+        } else {
+            values[index] = value;
+        }
     }
+
+    private int getIndex(K key) {
+        int index = -1;
```
Max-Pochepets, line 44:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+        for (int i = 0; i < amount; i++) {
+            if ((key == keys[i]) || (keys[i] != null && keys[i].equals(key))) {
+                index = i;
+                break;
```
Max-Pochepets, line 48:
```suggestion
                return i;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+            }
+        }
+        return index;
```
Max-Pochepets, line 51:
```suggestion
        return -1;
```

Max-Pochepets (approved):
🔥

PR #472, opened 2022-01-27
1st implementation of jv-storage task

OleksiyTitov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keysArray;
+    private final V[] valuesArray;
+    private int size = 0;
```
```suggestion
    private int size;
```
no need to initialize primitives with their default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
+            }
+            return -1;
+        } else {
+            for (int i = 0; i < MAX_SIZE; i++) {
+                if (keysArray[i] == null) {
+                    return i;
+                }
+            }
+            return -1;
+        }
```
Max-Pochepets, line 31: you can combine all this inro one loop with one condition

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexOfKey = getKeyIndex(key);
+        if (key != null) {
```
Max-Pochepets, line 37: same here, one `if` will be enough

Max-Pochepets (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -6,58 +6,40 @@
...
+            valuesArray[indexOfKey] = value;
         }
     }
 
-    private boolean storageWithNullKeyExist(int indexOfKey) {
-        return valuesArray[indexOfKey] != null ? true : false;
+    private boolean storageWithKeyExist(int indexOfKey) {
+        if (indexOfKey >= 0) {
+            return valuesArray[indexOfKey] != null ? true : false;
+        }
```
AlexRozv, line 41: Can we simplify it and just return valuesArray[indexOfKey] != null; it will give us same result

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #473, opened 2022-01-27
Task Storage from Generics. All tests passed.

Drapikovsky:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_OF_ELEMENT = 10;
+    private K[] keyArray;
+    private V[] valueArray;
+    private int numberOfElement = 0;
```
```suggestion
    private int numberOfElement;
```
we don't need to initialize primitives with their default values

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #474, opened 2022-01-27
Storage was implemented

nastyaprokhorenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,34 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_OF_ELEMENTS = 10;
+    private int size = 0;
```
Max-Pochepets, line 7:
```suggestion
    private int size;
```
no need to initialize primitives with their default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,34 @@
...
+    private K[] keys = (K[]) new Object[MAX_NUMBER_OF_ELEMENTS];
+    private V[] values = (V[]) new Object[MAX_NUMBER_OF_ELEMENTS];
```
Max-Pochepets, line 9: [Constructor initialization](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,34 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
```
Max-Pochepets, line 18: if the key is present, we should substitute value for this key. and the key will not always be on the last position

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #476, opened 2022-01-28
Methods' solutions were added to StorageImpl class.

Liubov-Chernohor:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
```
```suggestion
    private int size;
```
no need to assign to primitives their default values

Max-Pochepets (approved):

PR #480, opened 2022-01-28
homevork Storage

black-bit0:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            //if (key == keys[i] || keys[i] != null && key.equals(keys[i])) {
+            if (Objects.equals(keys[i], key)) {
```
AlexRozv, line 34: let's remove Objects.equals() and use commented code here

Screenshot 2022-01-31 at 09 31 32

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
+        values = (V[]) new Object[MAX_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (get(key) != null) {
+                values[i] = value;
+                return;
+            }
```
AlexRozv, line 23: let's corret your if statement a little, otherwise it makes no sense
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                values[i] = value;
                return;
            }
```

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #483, opened 2022-01-29
First commit StorageTask

o240789:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
+    private int getIndex(K key) {
+        int index = -1;
+        for (int i = 0; i < size; i++) {
+            if (keyStorage[i] == key || keyStorage[i] != null && keyStorage[i].equals(key)) {
+                index = i;
+                break;
+            }
+        }
+        return index;
```
AlexRozv, line 47: let's simplify this part
```suggestion
    private int getIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (keyStorage[i] == key || keyStorage[i] != null && keyStorage[i].equals(key)) {
                return i;
            }
        }
        return -1;
```

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #484, opened 2022-01-29
Implemented methods /put/, /get/, /size/

NazariiYatsiuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 10;
+    private final K[] keyBox = (K[]) new Object[ARRAY_LENGTH];
+    private final V[] valueBox = (V[]) new Object[ARRAY_LENGTH];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
+            if (keyBox[i] == null && key == null
+                    || keyBox[i] != null && keyBox[i].equals(key)) {
```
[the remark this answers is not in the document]
karma-o: This is the only thing that I have found that needs to be done.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                valueBox[i] = value;
+                break;
+            }
+            if (keyBox[i] == null && key != null && valueBox[i] == null) {
```
karma-o, line 18: This condition is written somewhat strange but I guess you can leave it as it is.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                keyBox[i] = key;
+                valueBox[i] = value;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < ARRAY_LENGTH; i++) {
```
karma-o, line 28: Every time you are checking each element of you storage, you also check `null` values for every `null` key of it by using `ARRAY_LENGTH` as a bound for a loop, when we could check only assigned values of the array, we can do this by something like this:
```
private dynamicStorageSize = 0;    //size which will be increased dynamically

    for (int i = 0; i <= dynamicStorageSize; i++) {    //this will prevent change of the value for every null key
        //some logic...
        dynamicStorageSize++;    //storage size increases when we add a new element to storage
    }
}
```
Also this will simplify our `size()` method to just this:
```
public int size() {
    return dynamicStorageSize;
}
```
Anyway, all of the above is not necessary to do, and is just a suggestion from fellow mate ; )

karma-o: Oops, I have seen that you have a `break` in your for loop in `put()` too late, so most of my comment doesnt make any sense, but I guess you can still use it as a simplifier of the `size()` method

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: the general algorithm for the `put()` method is:
1. check for occurrences of the key in our array
(we dont have to check if both of them are null, this condition compares null values correct:
keyBox[i] == key || keyBox[i] != null && keyBox[i].equals(key))
2. check if value of that key is assigned
3. if yes -> rewrite the value
4. if no -> create new pair of key & value

karma-o (approved):

AlexRozv (approved):
well done 👍

PR #485, opened 2022-01-29
Implemented parameterized class StorageImpl.

karma-o:
* Passed all the tests;
* Objects methods were not used;
* Only 1 array was used;

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,49 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int KEY_INDEX = 0;
+    public static final int VALUE_INDEX = 1;
+    public static final int MAX_ELEMENTS = 10;
+    public static final int PAIR_LIMIT = 2;
+    private int dynamicStorageSize = 0;
+    private final Object[][] arrayOfPairs;
+
+    public StorageImpl() {
+        this.arrayOfPairs = new Object[MAX_ELEMENTS][PAIR_LIMIT];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean exists = false;
+        for (int i = 0; i <= dynamicStorageSize; i++) {
+            if ((key == arrayOfPairs[i][KEY_INDEX]
+                    || key != null && key.equals(arrayOfPairs[i][KEY_INDEX]))) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int KEY_INDEX = 0;
+    public static final int VALUE_INDEX = 1;
+    public static final int MAX_ELEMENTS = 10;
+    public static final int PAIR_LIMIT = 2;
+    private int dynamicStorageSize;
+    private final Object[][] arrayOfPairs;
+
+    public StorageImpl() {
+        this.arrayOfPairs = new Object[MAX_ELEMENTS][PAIR_LIMIT];
```
Viktoriia-Zinchenko, line 14: "this" can be removed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        dynamicStorageSize = 0;
```
Viktoriia-Zinchenko, line 15: you can delete this line of code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
```
Viktoriia-Zinchenko, line 19: can be simplified:
        if (get(key) == null) {
            arrayOfPairs[dynamicStorageSize][KEY_INDEX] = key;
            arrayOfPairs[dynamicStorageSize][VALUE_INDEX] = value;
            dynamicStorageSize++;
        } else {
            arrayOfPairs[dynamicStorageSize - 1][VALUE_INDEX] = value;
        }
}

Viktoriia-Zinchenko (commented):
you can simplify the code a bit

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: I left a comment on your solution that my variant of this method could be better, but I am not sure about this, which is a bit funny, so I guess we will have to ask mentor which one is better 😃

AlexRozv (approved):
great job 🔥

karma-o:
> great job 🔥

Can you tell us please what implementation of the `put()` method would be better, or any of them would be ok?

PR #487, opened 2022-01-30
implemented methods and passed tests

antonignatenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_SIZE = 10;
+    private K[] arrayOfKeys;
+    private V[] arrayOfValues;
+    private int pointer = 0;
+
```
AlexRozv, line 10: there is no need to initialize primitive with it's default value
```suggestion
    private int pointer;

```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+        this.arrayOfKeys = (K[]) new Object[ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int pos;
+        for (int i = 0; i < arrayOfKeys.length; i++) {
+            if (pointer != 0 && (pos = getKeyPos(key)) >= 0) {
+                arrayOfKeys[pos] = key;
+                arrayOfValues[pos] = value;
```
AlexRozv, line 22: we can use your putPairOnPosition() here as well

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+            }
+        }
+        return -1;
+    }
+
+    private boolean isEquals(K key, K key1) {
+        if (key == null) {
+            return key1 == null;
+        }
+        return key.equals(key1);
```
AlexRozv, line 48: let's simplify this method
```suggestion
        return key == key1 || (key != null && key.equals(key1));
```

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #489, opened 2022-01-30
Implemented the Storage

Viktoriia-Zinchenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+    private final V[] values;
+    private int amount;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+    }
+
     @Override
+
```
karma-o, line 17: This line is redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
     public void put(K key, V value) {
+        if (get(key) == null) {
```
karma-o, line 19: You are checking only if it is `null`, but how do we check if element with such a `key` already exists,
for example if `key == null` and `value != null`, is it just overwrites the value if it sees `null`?
I think a condition like this:
`key == keys[i] || key != null && key.equals(keys[i]`
would be better, and more readable, but I am not sure, we better ask a mentor about this.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+            keys[amount] = key;
+            values[amount] = value;
+            amount++;
+        } else {
+            values[amount - 1] = value;
+        }
```
karma-o, line 25: I think `else` with its braces is redundant and we can remove it.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < amount; i++) {
```
karma-o, line 30: Smart move with looping only by dynamically changing `amount` of storage, and not by all `STORAGE_SIZE` 🔥

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+  private static final int STORAGE_SIZE = 10;
+    private final K[] keys;
```
AlexRozv, line 7:
```suggestion
    private static final int STORAGE_SIZE = 10;
    private final K[] keys;
```

AlexRozv: let's fix identation

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #490, opened 2022-01-30
Task done. Tests passed. Codestyle checked.

lugburz-hr:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_OF_ELEMENTS = 10;
+    private static final int FIRST_INDEX = 0;
+    private int numberOfElements;
+    private final K[] arrayOfKeys;
+    private final V[] arrayOfValues;
+
+    public StorageImpl() {
+        numberOfElements = 0;
```
sviat-h, line 13: we don't need to set primitives with their default values
```suggestion
```

sviat-h (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 
     @Override
     public V get(K key) {
+        int counter = FIRST_INDEX;
+        for (Object k : arrayOfKeys) {
+            if (k != null && k.equals(key)) {
+                return arrayOfValues[counter];
+            } else if (k == null && key == null) {
+                return arrayOfValues[counter];
+            }
```
AlexRozv, line 46: we can simplify this if statement, let's do the same in put method
```suggestion
            if (key == k || k != null && k.equals(key)) {
                return arrayOfValues[counter];
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     public void put(K key, V value) {
+        int counter = FIRST_INDEX;
+        for (K keyToCheck: arrayOfKeys) {
+            if (keyToCheck != null && keyToCheck.equals(key)) {
+                arrayOfValues[counter] = value;
+                return;
+            } else if (keyToCheck == null && key == null) {
+                if (arrayOfValues[counter] == null) {
+                    numberOfElements++;
+                }
```
AlexRozv, line 27: even if we add null as a value we should increase size number, so let's remove this if check

AlexRozv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
lugburz-hr: This code is to distinguish two cases when key in storage is null. 
1) If key in array and its value == null ==> this is not an element, but an empty cell, so we must write new value there and increase number of elements. For this case is condition in line 25.
2) If keys are equal and null, but value in array is not null ==> we should re-write value and not to increase number of elements.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -14,6 +14,15 @@ public StorageImpl() {
...
 
+    /* This code is to distinguish two cases when key in storage is null.
+
+        1. If key in array and its value == null ==> this is not an element, but an empty cell,
+          so we must write new value there and increase number of elements.
+          For this case is condition in line 34.
+        2. If keys are equal and null, but value in array is not null ==>
+          we should re-write value and not to increase number of elements.
+     */
+
```
AlexRozv, line 25: let's remove commented code

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #491, opened 2022-01-31
added solution

Navijava:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < this.value; i++) {
+            if (get(key) != null) {
+                values[i] = value;
+                return;
+            }
+        }
+        this.key[this.value] = key;
+        values[this.value] = value;
+        this.value++;
```
liliia-ponomarenko, line 26: you can simplify the method using if-else construction :
`if (get(key) == null) {
        this.key[this.value] = key;
        values[this.value] = value;
        this.value++;
} else {
values[this.value - 1] = value;
}`

liliia-ponomarenko (approved):

AlexRozv (approved):
good job 👍

PR #506, opened 2022-03-05
Storage was implemented as per requirments

B-i-g-D:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    public static final int MAX_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_SIZE];
+    private int count;
```
MaliukDaria, line 10:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
+
     @Override
-    public void put(K key, V value) {
+    public void put(Object key, Object value) {
+        for (int i = 0; i < count; i++) {
+            if (keyArray[i] == key || key != null && key.equals(keyArray[i])) {
+                valueArray[i] = (V) value;
+                return;
+            }
+        }
```
MaliukDaria, line 19: давай вынесем общую логику по поиску элемента по ключу в отдельный метод

MaliukDaria (changes requested):

pom.xml
```diff
@@ -16,6 +16,7 @@
         <maven.checkstyle.plugin.configLocation>
             https://raw.githubusercontent.com/mate-academy/style-guides/master/java/checkstyle.xml
         </maven.checkstyle.plugin.configLocation>
+        <kotlin.version>1.6.20-RC</kotlin.version>
```
MaliukDaria, line 19: Why do we need kotlin here? 🙂

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    public static final int MAX_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_SIZE];
+    private int size;
+
     @Override
-    public void put(K key, V value) {
+    public void put(Object key, Object value) {
```
MaliukDaria, line 13:
```suggestion
    public void put(K key, V value) {
```

why Object?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    }
+
+    private boolean checkIfAlreadyExists(Object key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keyArray[i] == key || key != null && key.equals(keyArray[i])) {
+                valueArray[i] = value;
+                return true;
+            }
+        }
+        return false;
```
```suggestion
    private int getIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (keyArray[i] == key || key != null && key.equals(keyArray[i])) {
                return i;
            }
            return -1;
        }
```

and we can use this method in put method and get method

MaliukDaria (changes requested):

pom.xml
MaliukDaria: I still don't understand why do we need kotlin here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keyArray[i] == key || key != null && key.equals(keyArray[i])) {
+                return i;
+            }
+        }
+        return -1;
     }
```
MaliukDaria, line 32: private methods should follow after public

MaliukDaria (approved):

PR #508, opened 2022-03-14
HomeWork jv-storage

JaysonJuza:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        }
+        keyStorage[size] = key;
+        valueStorage[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyStorage.length; i++) {
+            for (int j = 0; j < valueStorage.length; j++) {
```
KseniiaMakarova, line 32: why do we need 2 nested loops here? look at your `put()` method

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #509, opened 2022-03-16
Implementation Storage 

irynahrynda:
-  added KeyValue class
- implemented methods: put, get and size in class StorageImpl

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,51 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size = 0;
+    private KeyValue<K, V>[] storage;
+
+    public StorageImpl() {
+        storage = new KeyValue[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size > 0) {
+            for (int i = 0; i < size; i++) {
+                if ((storage[i].getKey() == null && key == null)
+                        || (storage[i].getKey() != null && storage[i].getKey().equals(key))) {
+                    storage[i].setValue(value);
+                    return;
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```
0 is the default value of int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    public StorageImpl() {
+        storage = new KeyValue[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size > 0) {
+            for (int i = 0; i < size; i++) {
+                if ((storage[i].getKey() == null && key == null)
+                        || (storage[i].getKey() != null && storage[i].getKey().equals(key))) {
```
KseniiaMakarova, line 19: && is evaluated before || so you can reduce the number of brackets
also, we can compare 2 nulls with ==

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        if (size < MAX_SIZE) {
+            storage[size++] = keyValue;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (size == 0) {
+            return null;
+        }
```
KseniiaMakarova, line 37: do we need this check?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+
+        for (int i = 0; i < size; i++) {
+            if ((storage[i].getKey() == null && key == null)
+                    || (storage[i].getKey() != null && storage[i].getKey().equals(key))) {
```
KseniiaMakarova, line 41: same comment as in `put()`

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #510, opened 2022-03-16
added solution for jv-storage task

dmnichz:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_CAPACITY = 20;
```
KseniiaMakarova, line 6: max capacity should be 10

also, access modifier is missing

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    private Object[] array;
+
+    public StorageImpl() {
+        this.array = new Object[0];
```
KseniiaMakarova, line 10: why 0?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_CAPACITY = 20;
+    private Object[] array;
```
KseniiaMakarova, line 7: let's better have 2 arrays of size 10, one for keys, and one for values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        boolean isPresent = false;
+        if (array.length == 0) {
+            array = new Object[array.length + 2];
+            array[0] = key;
+            array[1] = value;
+        }
+        for (int i = 0; i < array.length; i = i + 2) {
+            if (array[i] == key || array[i] != null && array[i].equals(key)) {
+                array[i + 1] = value;
+                isPresent = true;
```
KseniiaMakarova, line 24: you can return from the method inside the loop using `return` keyword

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
dmnichz: what number can be here instead of 0?

src/main/java/core/basesyntax/impl/StorageImpl.java
dmnichz: I do not understand how to implement this with 2 arrays...

src/main/java/core/basesyntax/impl/StorageImpl.java
dmnichz: ok

src/main/java/core/basesyntax/impl/StorageImpl.java
dmnichz: in task instructions benn said you can use 1 or 2 arrays

dmnichz: in theory videos was nothing about how to solve such a task

KseniiaMakarova (approved):
great job!

PR #517, opened 2022-03-29
Implemented put and get method in StorageImpl class.

BigVolodumur:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,24 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final Map<K, V> map = new HashMap<>();
```
KseniiaMakarova, line 8: let's not use any collections here
try doing this with 2 arrays

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private K[] arrayKey = (K[]) new Object[MAX_ITEMS_NUMBER];
+    private V[] arrayValue = (V[]) new Object[MAX_ITEMS_NUMBER];
```
KseniiaMakarova, line 8: let's initialize non-static fields in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private int count = 0;
```
```suggestion
    private int count;
```
0 is the default value of int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (getIndex(key) >= 0) {
+            arrayValue[getIndex(key)] = value;
```
KseniiaMakarova, line 14: you are calling getIndex() 2 times - let's save its result into local variable and reuse it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        }
+        arrayKey[count] = key;
+        arrayValue[count] = value;
+        count++;
     }
 
     @Override
     public V get(K key) {
+        if (getIndex(key) >= 0) {
+            return arrayValue[getIndex(key)];
```
KseniiaMakarova, line 25: same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
     }
 
     @Override
     public int size() {
+        return count;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < count; i++) {
+            if (key == null && arrayKey[i] == null || key != null && key.equals(arrayKey[i])) {
```
KseniiaMakarova, line 37:
```suggestion
            if (key == arrayKey[i] || key != null && key.equals(arrayKey[i])) {
```
we can compare 2 nulls with ==

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #518, opened 2022-03-29
Generics-homework

Daniil2k-3:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+
+    private K[] keyData = (K[]) new Object[STORAGE_CAPACITY];
+    private V[] valueData = (V[]) new Object[STORAGE_CAPACITY];
```
KseniiaMakarova, line 9: let's initialize non-static fields in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```
0 is the default value of int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+
     @Override
     public void put(K key, V value) {
+        keyData[size] = key;
+        valueData[size] = value;
```
KseniiaMakarova, line 15: let's do this after the loop when it is clear that there is no such key

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key != null && key.equals(keyData[i]) || key == keyData[i]) {
```
KseniiaMakarova, line 17: you can return from the whole method using `return` inside a loop if you found a key

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #522, opened 2022-04-02
Storage was implemented

kseniabykh:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_STORAGE = 10;
+    private K[] keys = (K[]) new Object[MAX_NUMBER_STORAGE];
+    private V[] values = (V[]) new Object[MAX_NUMBER_STORAGE];
+    private int size = 0;
```
Sofasmile, line 9:
```suggestion
    private int size;
```
the default value for int is zero

Sofasmile (changes requested):
awesome))

Sofasmile (changes requested):
fix one previous comment
image

Sofasmile (approved):

PR #524, opened 2022-04-02
implemented StorageImpl with array

DLyudmila:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    return values[i];
+                }
+                continue;
+            }
+            if (key.equals(keys[i])) {
+                return values[i];
+            }
```
MaliukDaria, line 50: same here, lets combine this ifs

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        values = (V[])new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] != null && key != null && keys[i].equals(key))
+                    || (key == null && keys[i] == null)) {
+                values[i] = value;
+                return;
```
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                values[i] = value;
                return;
            }
```

we could simplify this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null)
+                    || (key != null && keys[i] != null && key.equals(keys[i]))) {
```
MaliukDaria, line 34: same here

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
DLyudmila: yes, you are right

MaliukDaria (approved):
GJ🔥

PR #529, opened 2022-04-04
Added homework solution

ZeroNeiron:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
```
VladyslavLn, line 9:
```suggestion
    private int size;
```

Default int values is 0

VladyslavLn (changes requested):

VladyslavLn (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE];
+        values = (V[]) new Object[MAX_STORAGE];
+    }
+
+    public int getKeyPosition(K key) {
```
Sofasmile, line 16: make this method private and move it methods below their usages in public or just to the end of file

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            }
+        }
+        return -1;
+    }
+
     @Override
     public void put(K key, V value) {
+        int keyPosition = getKeyPosition(key);
+        if (keyPosition != -1) {
+            keys[keyPosition] = key;
```
Sofasmile, line 29: the key is already exist in the array, so we can remove this line
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return (getKeyPosition(key) == -1) ? null : values[getKeyPosition(key)];
```
Sofasmile, line 40: better save the result of `getKeyPosition(key)` to the separate variable
```suggestion
        int index = getKeyPosition(key);
        return (index == -1) ? null : values[index];
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #533, opened 2022-04-04
implemented task

Anastasia-702:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
```
```suggestion
    private int size;
```

It will be initialized by 0 by default

MaliukDaria (approved):

PR #542, opened 2022-04-05
Implemented StorageImpl class

Syon2154:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,34 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[][] data = new Object[10][2];
```
tanya-logatskaya, line 6: You need two arrays here. One with keys and the other with values
Maybe this realization is ok, I'm not sure
So I'll leave a comment but approve

tanya-logatskaya (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
Sofasmile, line 6: better create constants for 10 and 2 numbers
and we should initialize this array in the constructor

Sofasmile (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,15 +1,18 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    private Object[][] data = new Object[10][2];
+    private static final int MAX_CAPACITY = 10;
+    private static final int KEYS_AND_VALUES_SPACE = 2;
+    private Object[][] data = new Object[MAX_CAPACITY][KEYS_AND_VALUES_SPACE];
```
Sofasmile, line 9: we should initialize this array in the constructor

Sofasmile (changes requested):

Sofasmile (approved):

PR #544, opened 2022-04-08
- was implemented the Storage

MarkShapka:
- implemented and overrided three methods

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,46 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUE_SIZE = 10;
+    private Object[] keys = new Object[MAX_VALUE_SIZE];
+    private Object[] values = new Object[MAX_VALUE_SIZE];
+    private int count = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < count; i++) {
+            if ((key == null && key == keys[i])) {
+                values[i] = value;
+                return;
+            } else if (key == null) {
+                break;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUE_SIZE = 10;
+    private Object[] keys = new Object[MAX_VALUE_SIZE];
+    private Object[] values = new Object[MAX_VALUE_SIZE];
+    private int count = 0;
```
```suggestion
    private int count;
```

we dont need to initialize int values bu 0 as it will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        for (int i = 0; i < count; i++) {
+            if ((key == null && key == keys[i])) {
+                values[i] = value;
+                return;
+            } else if (key == null) {
+                break;
+            } else if (key == keys[i] || key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
```
MaliukDaria, line 22: could we combine this if's?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     public V get(K key) {
+        for (int i = 0; i < count; i++) {
+            if (key == null) {
+                if (key == keys[i]) {
+                    return (V) values[i];
+                }
+                continue;
+            } else if (key == keys[i] || key.equals(keys[i])) {
+                return (V) values[i];
+            }
```
MaliukDaria, line 39: same here

MaliukDaria (changes requested):

Nick97-git (approved):

PR #545, opened 2022-04-09
add solution for jv-storage

SerhiiPankov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (int i = 0; i < h; i++) {
+            if (storage[0][i] != null || storage[1][i] != null) {
+                size++;
+            }
+        }
+        return size;
```
Nick97-git, line 59: you should create field size and increment it after the element was added, so you can return size itself in this method without any additional logic

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i <= size(); i++) {
+            if (storage[0][i] != null && storage[0][i].equals(key)) {
+                return (V) storage[1][i];
+            }
+            if (key == null && storage[0][i] == null && storage[1][i] != null) {
+                return (V) storage[1][i];
+            }
```
Nick97-git, line 46: you can use one if here instead of two

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int h = 5;
+    private K key;
+    private V value;
+    private final Object[][] storage = new Object[2][10];
```
Nick97-git, line 9: let's  rewrite your logic, instead of creating 2-dimensional array, let's create two arrays one for storing keys and the second one for values

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int h = 10;
+    private int size;
+    private final K[] storageK = (K[]) new Object[h];
+    private final V[] storageV = (V[]) new Object[h];
```
Nick97-git, line 9:
```suggestion
    private final K[] keys = (K[]) new Object[h];
    private final V[] values = (V[]) new Object[h];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int h = 10;
```
Nick97-git, line 6:
```suggestion
    private static final int STORAGE_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            }
+
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i <= size(); i++) {
+            if ((storageK[i] != null && storageK[i].equals(key))
+                    || (key == null && storageK[i] == null && storageV[i] != null)) {
```
Nick97-git, line 44:
```suggestion
      if (keys[i] == key || (key != null && key.equals(keys[i]))) {
```

Nick97-git (changes requested):

Nick97-git (approved):

PR #549, opened 2022-04-14
the solution was implemented

Dima-Davydenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
Nick97-git, line 6: remove empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
-        return -1;
+        int i = 0;
+        while (i < MAX_ITEMS_NUMBER) {
+            if (keys[i] != null || values[i] != null) {
+                i++;
+            } else {
+                break;
+            }
+        }
+        return i;
```
Nick97-git, line 57: create global field size and increment it when new element is added, here you could return size;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        while (j < MAX_ITEMS_NUMBER) {
+            if (keys[j] != null && keys[j].equals(key)) {
+                break;
+            } else if (keys[j] == null && key == null) {
+                break;
+            } else {
+                j++;
+            }
+        }
+        return j == MAX_ITEMS_NUMBER ? null : (V) values[j];
```
Nick97-git, line 44: you can simplify this block of code, use for loop, inside of it it's enough to have one if condition, in case it's true return found element, in case storage doesn't contain such key after for loop return -1;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+            if (keys[i] == null && values[i] == null
+                    || keys[i] == null && key == null
+                    || keys[i] != null && keys[i].equals(key)) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            } else {
+                i++;
+            }
+        }
```
Nick97-git, line 29: let's simplify this block of code using for loop, inside of it you should check if such key already exists and if it's true change its value, and exit from the method using return;
after for loop you just add a new element in case if in for loop you didn't find such key

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+    }
+
+    private boolean isKeyExist(int i, K key) {
```
MaliukDaria, line 16:
```suggestion
    private boolean isKeyEqual(int index, K key) {
```

i - not informative variable name. Private methods should follow after public as helper ones

MaliukDaria (changes requested):

MaliukDaria (changes requested):
test

MaliukDaria (approved):

PR #552, opened 2022-04-15
Task complete

AlexOriobel:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,24 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Map<K, V> data = new HashMap<>();
```
Nick97-git, line 8: https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#:~:text=Don%E2%80%99t%20use%20any%20kind%20of%20List%2C%20Set%20or%20Map

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            }
+        }
+        if (!temp) {
+            keys[count] = key;
+            values[count] = value;
+            count++;
+        } else {
+            keys[index] = key;
+            values[index] = value;
+        }
```
MaliukDaria, line 35: Could we just set a new value in a for loop and use `return`, not to use index ant temp variables?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     public V get(K key) {
-        return null;
+        int index = 0;
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null ? keys[i] == key : keys[i].equals(key))
+                    && values[i] != null) {
+                index = i;
+            }
+        }
+        return (V) values[index];
```
MaliukDaria, line 47:
```suggestion
        for (int i = 0; i < keys.length; i++) {
            if ((keys[i] == null ? keys[i] == key : keys[i].equals(key))
                    && values[i] != null) {
                return values[i];
            }
        }
        return null;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        values = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null ? keys[i] == key : keys[i].equals(key))
+                    && values[i] != null) {
+                keys[i] = key;
+                values[i] = value;
```
```suggestion
                    && values[i] != null) {
                values[i] = value;
```

I dont think we should set key here

MaliukDaria (approved):

PR #553, opened 2022-04-16
Solution try two arrays

alexgash8:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+            }
+            if (keys[i] == null & key == null) {
+                values[i] = value;
+                break;
+            }
+            if (keys[i] != null) {
+                if (keys[i].equals(key)) {
+                    values[i] = value;
+                    break;
+                }
```
MaliukDaria, line 31: can we combine some of this if`s?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    private Object[] values;
+
+    public StorageImpl() {
+        keys = (K[])new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < 10; i++) {
```
MaliukDaria, line 17:
```suggestion
        for (int i = 0; i < 10; i++) {
```

why to 10?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+            if (keys[i] == null) {
+                if (key == null) {
+                    return (V) values[i];
+                } else {
+                    continue;
+                }
+            }
+            if (keys[i].equals(key)) {
+                return (V) values[i];
+            }
```
MaliukDaria, line 48: same, can we combine this if's

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
     @Override
     public int size() {
-        return -1;
+        int counter = 0;
+        for (int i = 0; i < MAX_CAPACITY; i++) {
+            if (keys[i] != null || values[i] != null) {
+                counter++;
+            }
+        }
+        return counter;
```
MaliukDaria, line 61: lets create some class level size variable, where we will store information about storage size

MaliukDaria (changes requested):

MaliukDaria (changes requested):
test

src/main/java/core/basesyntax/impl/StorageImpl.java
alexgash8: spent 1 hour and didn't find better solution...

alexgash8: found one optimisation

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private Object[] keys;
+    private Object[] values;
```
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+                    break;
+                }
+            }
+            if (keys[i] != null) {
+                if (keys[i].equals(key)) {
+                    values[i] = value;
+                    break;
+                }
+            }
+        }
```
MaliukDaria, line 37:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])){
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+            if ((keys[i] == null) && (key == null)) {
+                return (V) values[i];
+            }
+            if (keys[i] != null) {
+                if (keys[i].equals(key)) {
+                    return (V) values[i];
+                }
+            }
+        }
         return null;
```
MaliukDaria, line 52:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])){
                return values[i];
            }
        }
        return null;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
alexgash8: got it, more beautiful

alexgash8: but i cant get how it works, 
image
&& precedence says that key.equals[keys[i]] would be evaluated anyway (as a second step), so if it passed there a null value - it should throw NullPointerExpression, but it doesn't, strange...

alexgash8: ahhhh, got it)) key != null!!!! and than to step #2

alexgash8:
i've got a little problem with instantiation generic array, could'n write another way, so used the proposal of IDEA to suppress warnings, i'm not sure done correctly...
image

MaliukDaria:
> i've got a little problem with instantiation generic array, could'n write another way, so used the proposal of IDEA to suppress warnings, i'm not sure done correctly... image

you can use SupressWarnings here

MaliukDaria (approved):

PR #555, opened 2022-04-18
1th try: implemented methods put, get, size and findKey

ihor-dmytrush:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,51 @@
...
+        int existingKeyIndex = findKey(key);
+        int size = size();
+        if (existingKeyIndex != -1) {
+            this.values[existingKeyIndex] = value;
+        } else if (size < ARRAY_CAPACITY) {
+            this.keys[size] = key;
+            this.values[size] = value;
+        } else {
+            throw new RuntimeException("Storage is full!");
+        }
```
```suggestion
    if (size == ARRAY_CAPACITY) {
      throw new RuntimeException("Storage is full!");
    }
    int existingKeyIndex = findKey(key);
    if (existingKeyIndex != -1) {
      this.values[existingKeyIndex] = value;
      return;
    }
    this.keys[size] = key;
    this.values[size] = value;
    size++;
```

it's hart to read if-else constructions, better to avoid them where we can

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,48 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_CAPACITY = 10;
+    private final K[] keys = (K[]) new Object[ARRAY_CAPACITY];
+    private final V[] values = (V[]) new Object[ARRAY_CAPACITY];
+    private int size = 0;
```
MaliukDaria, line 10:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,48 @@
...
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < ARRAY_CAPACITY; i++) {
+            if (key == null && keys[i] == null && values[i] != null) {
+                return i;
+            }
+            if (key != null && key.equals(keys[i])) {
+                return i;
+            }
```
MaliukDaria, line 45:
```suggestion
            if (key == keys[i]  && key != null && key.equals(keys[i])) {
                return i;
            }
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ihor-dmytrush: When i did these changes, mvn test is not passing
image

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
MaliukDaria: I specified the wrong operator we need || not &&

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,46 @@
...
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < ARRAY_CAPACITY; i++) {
+            if ((key == null && keys[i] == null && values[i] != null)
+                || (key != null && key.equals(keys[i]))) {
+                return i;
+            }
+        }
         return -1;
```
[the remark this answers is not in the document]
ihor-dmytrush: image
I understand ;-(... it's not working

MaliukDaria (approved):

PR #557, opened 2022-04-19
Hw storage chernenko

DmytroChernenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private final Pair<K, V>[] storage = new Pair[10];
```
MaliukDaria, line 7: lets move 10 to constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
...
+    private int count = 0;
```
```suggestion
    private int size;
```

it will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (Pair<K, V> pair : storage) {
```
MaliukDaria, line 12: I think it's better to use fori loop and limit by size variable (so not to check empty null elements)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
...
+            }
+            if (key == null) {
+                if (pair.getKey() == key) {
+                    pair.setValue(value);
+                    return;
+                }
+            } else if (key.equals(pair.key)) {
+                pair.setValue(value);
+                return;
+            }
```
MaliukDaria, line 24: could we combine this if's?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
...
+        Pair<K, V> pair = new Pair<>(key, value);
+        storage[count++] = pair;
     }
 
     @Override
     public V get(K key) {
+        for (Pair<K, V> pair : storage) {
+            if (pair == null) {
+                continue;
+            }
```
MaliukDaria, line 36: same here, use fori loop

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,100 @@
...
+            if (pair.getKey() == null && key == null) {
+                return pair.getValue();
+            } else if (pair.getKey() == null) {
+                continue;
```
MaliukDaria, line 40: why do we need to check key for null?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
DmytroChernenko: we can't use Objects class with Objects.equals()  and key can be null. so if key is null we should compare using == operator, but if key not null we should use equals. But if we can use Objects.equals it would 1 condition like `if Objects.equals(key, pair.getKey())`)

src/main/java/core/basesyntax/impl/StorageImpl.java
DmytroChernenko: the same about Objects.equals and NPE.

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,97 @@
...
+
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (storage[i].getKey() == key) {
+                    storage[i].setValue(value);
+                    return;
+                }
+            } else if (key.equals(storage[i].key)) {
+                storage[i].setValue(value);
+                return;
```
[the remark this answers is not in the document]
MaliukDaria: What about this comment? Lets combine this if's as you did in `get(K key)` method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,91 @@
...
+            }
+            return value != null ? value.equals(pair.value) : pair.value == null;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = key != null ? key.hashCode() : 0;
+            result = 31 * result + (value != null ? value.hashCode() : 0);
+            return result;
+        }
```
MaliukDaria, line 91: do you use this methods?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,91 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private final Pair<K, V>[] storage = new Pair[DEFAULT_CAPACITY];
+    private int size;
+
     @Override
     public void put(K key, V value) {
+
```
MaliukDaria, line 13: redundant empty line, lets remove it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,91 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
MaliukDaria, line 6: redundant empty line, lets remove it

MaliukDaria (changes requested):

Nick97-git (approved):

PR #565, opened 2022-05-01
the jv-storage task is completed and ready for check

oleksandr-petrunovskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < valueArray.length; i++) {
+            if (valueArray[i] == null) {
+                return i;
+            }
+        }
+        return valueArray.length;
```
MaliukDaria, line 39: lets create size variable and increment it every time we add new element

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_VOLUME = 10;
+    private final Object[] keyArray = new Object[MAX_STORAGE_VOLUME];
+    private final Object[] valueArray = new Object[MAX_STORAGE_VOLUME];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if ((keyArray[i] == null && valueArray[i] == null)
+                    || (keyArray[i] == key || (keyArray[i] != null && keyArray[i].equals(key)))) {
```
MaliukDaria, line 14: I looks like we have similar logic in `put` and `get` methods, could we move it in the separate private method?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-petrunovskyi: Yes, we can, but please tell me why? I understand that it consumes resources, but the "size" method will not work without the "put" method.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    public StorageImpl() {
+        keyArray = (K[]) new Object[MAX_STORAGE_VOLUME];
+        valueArray = (V[]) new Object[MAX_STORAGE_VOLUME];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (checkElementStorageIsFree(i)) {
```
MaliukDaria, line 19: why do we need this check? Lets first get through all elements in keyArray (from 0 to size) and check if we already had such key in storage, if yes - set new value, if not - add new element in the end of array:
```
keys[size] = key;
values[size] = value;
size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_VOLUME = 10;
+    private final K[] keyArray;
+    private final V[] valueArray;
```
```suggestion
    private final K[] keys;
    private final V[] values;
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #567, opened 2022-05-05
tests and stylechecks passed

drapail:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxArraySize = 10;
```
MaliukDaria, line 6: What about constant naming? https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+            if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            } else if (keys[i] == key || (key != null && keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                break;
+            } else if (i == maxArraySize) {
+                System.out.println("You can't place a new key, dictionary is full");
+            }
```
MaliukDaria, line 22: lets add this check at the beginning of method and throw some exception in case the size == maxArraySize

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null) {
+                return i;
+            }
+        }
+        return maxArraySize;
```
MaliukDaria, line 43: It's better to create size class level variable and increment it every time we add new key-value pair

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            } else if (keys[i] == key || (key != null && keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                break;
```
MaliukDaria, line 19: We could have key-value pair with key == null and value == null, then your logic won't work, lets go through all elements in storage (from 0 to size) and check if we already have such key, if not , add new element in the end of storage
keys[size] = key;
values[size] = value;
And think if we have some duplicated logic in `put(K key, V value)` method and `get(K key)` method witch we can move to separate method

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
drapail: If I name it like MAX_ARRAY_SIZE Maven throws style error saying I shouldn't have more then 2 consequent capital letters in a row

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxArraySize = 10;
+    private int size = 0;
```
Nick97-git, line 7:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxArraySize = 10;
```
Nick97-git, line 6:
```suggestion
    private static final int STORAGE_SIZE  = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            size++;
+            return;
+        }
+
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i]) {
+                keys[i] = key;
+                values[i] = value;
+                return;
+            } else if ((key != null && keys[i] != null) && keys[i].equals(key)) {
```
Nick97-git, line 25: unite this two ifs in one

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+                keys[i] = key;
+                values[i] = value;
+                return;
+            } else {
+                keys[size] = key;
+                values[size] = value;
+                size++;
```
Nick97-git, line 32: remove this else and move this block of code after for

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private V[] values = (V[]) new Object[maxArraySize];
+
     @Override
     public void put(K key, V value) {
+        if (size == 0) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+            return;
+        }
```
Nick97-git, line 18: you can remove this block of code

Nick97-git (changes requested):

Nick97-git (approved):

PR #572, opened 2022-05-19
implement task

gdpbalt:
- modify class StorageImpl:
-- init 2 arrays in constructor;
-- add private method findByKey() to find an element by a key;
-- add public method put(), get() and size() for working with data.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,48 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int index;
+
+    @SuppressWarnings({"unchecked"})
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_SIZE];
+        values = (V[]) new Object[MAX_STORAGE_SIZE];
+        index = -1;
+    }
+
     @Override
     public void put(K key, V value) {
+        int writeIndex = findByKey(key);
+        if (writeIndex < 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    }
+
+    private int findByKey(K key) {
+        for (int i = 0; i < size(); i++) {
+            if (key == null && keys[i] == null) {
+                return i;
+            }
+            if (key != null && key.equals(keys[i])) {
+                return i;
+            }
```
KostyaNikiforov, line 46:
```suggestion
            if (keys[i] == null || (key != null && key.equals(keys[i])) {
                return i;
            }
```
Or you can use `Objects.equals()` method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int index;
+
+    @SuppressWarnings({"unchecked"})
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_SIZE];
+        values = (V[]) new Object[MAX_STORAGE_SIZE];
```
KostyaNikiforov, line 14: Can we avoid using downcasting here?

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
gdpbalt: I don't know. Google didn't help me.
As far as I know this downcasting must be in this case.

Could you help me wi this point ?

KostyaNikiforov (approved):
LGTM:fire:

PR #576, opened 2022-05-20
implemented Storage of key-value

viktoriia-haraniuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < itemsSize; i++) {
+            if (key == null && keys[i] == null
+                    || keys[i] != null && keys[i].equals(key)) {
```
liliia-ponomarenko, line 39: you have duplicate code. Can we move this condition to a private method?

liliia-ponomarenko (changes requested):
good solution 🔥 let's improve it

liliia-ponomarenko (approved):
good job 😊

PR #578, opened 2022-05-22
Homework storage

Cosmic-M:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,95 @@
...
+        }
+        index = getIndexByKey(key);
+        if (index != INDEX_ABSENT) {
+            return values[index];
+        }
         return null;
     }
 
     @Override
     public int size() {
```
liliia-ponomarenko, line 64: you can simplify this method if you will use a counter and increase this  when putting a new item

liliia-ponomarenko (changes requested):
Let’s improve your solution ;)

Cosmic-M:
My private methods don't do the same, because they called to do absolutely different things. You must have suggested it wasn't becase they were almost in the same name. So I renamed one to avoid confuse instead of union them, because they called to do different things. But if you insist,... the matter is I  will need change logic in other methods...

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final String MESSAGE = "the storage is overloaded";
```
liliia-ponomarenko, line 7:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
...
+    private static final int STORAGE_CAPACITY = 10;
+    private static final int INDEX_ABSENT = -1;
+
```
liliia-ponomarenko, line 10:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
...
+    private int counter = 0;
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_CAPACITY];
+        values = (V[]) new Object[STORAGE_CAPACITY];
```
liliia-ponomarenko, line 17: you can avoid downcasting if use class Object in fields

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
...
+        }
+        index = getIndexByKey(key);
+        if (index != INDEX_ABSENT) {
+            keys[index] = key;
+            values[index] = value;
+            return;
+        }
+        if (size() == STORAGE_CAPACITY) {
+            throw new RuntimeException(MESSAGE);
+        }
```
liliia-ponomarenko, line 38: if the storage is overloaded what exception do you get? what will changes? you can remove this condition 
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
...
+            if (keys[i] == null && values[i] != null) {
+                return i;
+            }
+        }
+        return INDEX_ABSENT;
+    }
+
+    private int getIndexByKey(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] != null && Objects.equals(keys[i], key)) {
```
liliia-ponomarenko, line 76: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,82 @@
...
+        return (index != INDEX_ABSENT) ? values[index] : null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return counter;
+    }
+
+    private int getIndexIfKeyIsNull() {
```
liliia-ponomarenko, line 65: I meant that now you use different methods to get the index when the key is null and when != null. You can combine this into one method and simplify the logic

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < keys.length; i++) {
+                if (keys[i] != null && keys[i].equals(key)) {
+                    return i;
+                }
+            }
+        }
+        return INDEX_ABSENT;
```
liliia-ponomarenko, line 73:
```suggestion
    private int getIndexByKey(K key) {
     for (int i = 0; i < counter; i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                return i;
            }
        }
        return INDEX_ABSENT;
```
looks better? Let’s  think about how you can use it in methods put() and get()

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):
good job 👍

PR #579, opened 2022-05-22
 jv-storage-task-done

Pavlo99-gif:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE_VALUE = 6;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE_VALUE];
+        values = (V[]) new Object[MAX_SIZE_VALUE];
+    }
```
liliia-ponomarenko, line 14: Can we avoid using downcasting here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE_VALUE = 6;
```
```suggestion
    private static final int MAX_SIZE_VALUE = 6;
```
constants must have a private access modifier

liliia-ponomarenko (approved):
Great job! You can improve your solution if you want 😉

PR #580, opened 2022-05-22
methods get(), put() and size() was implemented.

Ivan-Yuriv:


liliia-ponomarenko (changes requested):
good solution 👍

liliia-ponomarenko (approved):
LGTM! 🔥

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):

PR #584, opened 2022-05-24
Implemented Storage service

ChumakAndrey:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            keysArray[currentPositionInArray] = key;
+            valuesArray[currentPositionInArray] = value;
+            currentPositionInArray++;
+        } else {
+            for (int i = 0; i < currentPositionInArray; i++) {
+                if (key != null && key.equals(keysArray[i])
+                        || key == keysArray[i]) {
+                    valuesArray[i] = value;
+                }
+            }
```
liliia-ponomarenko, line 31: this code-block duplicates method get(). Can we avoid it?

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):
LGTM! 🔥

PR #586, opened 2022-05-26
hw storage solution

TymurTsvietkov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final byte MAX_ELEMENTS= 10;
+    private final K[] keyValues;
+    private final V[] storedInfo;
+    private int elements = 0;
+
+    public StorageImpl() {
+        keyValues = (K[]) new Object[MAX_ELEMENTS];
+        storedInfo = (V[]) new Object[MAX_ELEMENTS];
```
liliia-ponomarenko, line 13: Can we avoid using downcasting here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    }
+
+    private boolean obEqual(Object o1, Object o2) {
+        return o1 == o2 || o1 != null && o1.equals(o2);
+    }
```
liliia-ponomarenko, line 18: private methods are added after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        elements++;
+        boolean newObject = true;
+        for (int i = 0; i < elements - 1; i++) {
+            if (obEqual(key, keyValues[i])) {
+                storedInfo[i] = value;
+                newObject = false;
+                elements--;
+                break;
+            }
+        }
```
liliia-ponomarenko, line 31: can we use the get() method instead of this loop?

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):
good job 👍

PR #589, opened 2022-05-27
Storage homework

IvanKunitskiy:
Was implemented two methods. put() - which accepts two values(key and value), get() - which returns specific value by key

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,42 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,42 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_VALUE = 10;
+    private int size = 0;
+    private final Object[] keyArray;
+    private final Object[] valueArray;
```
KostyaNikiforov, line 10:
```suggestion
    private final Object[] keys;
    private final Object[] values;
```
We'll already know it's an array if the name is just plural)

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #591, opened 2022-05-28
storage homework

NikitaSemisoshenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys = new Object[10];
+    private Object[] values = new Object[10];
```
KostyaNikiforov, line 7: Let's make magic numbers constants

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
     @Override
     public int size() {
-        return -1;
+        int index = 0;
+        for (Object obj : values) {
+            if (obj != null) {
+                index++;
+            }
+        }
+        return index;
```
KostyaNikiforov, line 40: Can we avoid using loop in the implementation of this method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((key == null && keys[i] == null) || ((key != null) && (key.equals(keys[i])))) {
+                return (V) values[i];
+            }
+        }
         return null;
```
KostyaNikiforov, line 29: Let's move the general logic into a separate private method 
We can write loop only once ;)

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):
Return the previous version)
This solution should be one class. And you don't have to create new objects each time you call one of those methods.
Screenshot 2022-05-30 at 14 11 09

KostyaNikiforov (approved):

PR #592, opened 2022-05-28
The class StorageImpl was implemented.

olyaklipa:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key == null ? keys[i] == null : key.equals((K) keys[i])) {
+                return (V) values[i];
+            }
+        }
```
KostyaNikiforov, line 41: Let's move the common logic into a separate private method 
We can write a loop only once ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        Object[] newKeys = new Object[keys.length + ROOM];
+        Object[] newValues = new Object[values.length + ROOM];
+        for (int i = 0; i < keys.length; i++) {
+            newKeys[i] = keys[i];
+            newValues[i] = values[i];
+        }
+        newKeys[keys.length] = key;
+        newValues[values.length] = value;
+        keys = newKeys;
+        values = newValues;
```
KostyaNikiforov, line 60: The task requires creating a vault with a maximum size of 10 cells.
It's cool that your solution allows us to put in an almost unlimited number of values, but it's actually not a very cheap operation, creating a new array of objects every time a new element is added. So let's keep it simple so that you don't have to grow the existing array each time.

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
olyaklipa: In fact, the task does not require an array to have the capacity of 10. It just says "...we assume that the maximum number of elements is 10..." But anyway, it was a useful task to do both options. 
Also, I am not sure if making sizeOfArray as a field of the class and increasing it in the add() method is a good approach. Initially, I made sizeOfArray variable local to the size() method and iterated over whole array (checking all 10 positions) looking for the null value either in the keys or in the values (it worked successfully). But then I realized that the approach where I change  sizeOfArray in add method works as well and looks much simpler. Looking forward to your comments on this matter.

KostyaNikiforov (approved):
LGTM :fire:

PR #594, opened 2022-05-29
storage implemented

oleksii-kordysh:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if (getKey(i) == null && getValue(i) == null) {
+                return i;
+            }
+        }
+        return MAX_ELEMENT_NUMBER;
```
KostyaNikiforov, line 47: Can we avoid using loop in the implementation of this method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if (keysEquals(getKey(i), key)) {
+                return getValue(i);
+            }
+        }
         return null;
     }
```
KostyaNikiforov, line 38: Let's move the common logic into a separate private method 
We can write loop only once ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ELEMENT_NUMBER; i++) {
+            if (keysEquals(getKey(i), key)) {
+                setValue(i, value);
+                break;
+            }
+            if (getKey(i) == null && getValue(i) == null) {
```
KostyaNikiforov, line 22: Do we really need to check it?

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksii-kordysh: deleted

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksii-kordysh: done

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksii-kordysh: done

KostyaNikiforov (approved):
🔥

PR #598, opened 2022-05-30
Storage implementation 

elenapokaliuk:
added implementation for method get to check the storage already contains given key
added implementation for method put to put new value and key to storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,44 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAXIMUM_STORAGE_ELEMENTS = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+    private int getKeyIndex;
+
+    public StorageImpl() {
+        keys = new Object[MAXIMUM_STORAGE_ELEMENTS];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAXIMUM_STORAGE_ELEMENTS = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+    private int getKeyIndex;
```
MaliukDaria, line 11: I dont think we need this variable. We can write private method, that gets array index by key or return -1, and use it in `put(K key, V value)` and `get(K key)` methods

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+        } else {
+            values[getKeyIndex(key)] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (getKeyIndex(key) != -1) {
+            return (V) values[getKeyIndex(key)];
+        }
```
MaliukDaria, line 41: can we use getKeyIndex(key) only ones here?(as this is quite heavy method and unnecessary reuse will negatively affect performance)

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #599, opened 2022-05-30
hw-9

zimmboco:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (key != null && key.equals(keys[i]))) {
+                values[i] = value;
+                return;
+            }
+        }
```
MaliukDaria, line 23: It seems that we have duplicated logic in `put` and `get` methods, could we move it to a separate private method?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
```
MaliukDaria, line 23: why `values[size-1]` ?
we can create method and use it in `put(K key, V value)` and `get(K key)` methods
```
    private int getIndex(K key) {
                return i;
            }
        }
        return -1;
    }
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[ELEMENTS_MAX_AMOUNT];
+        values = (V[]) new Object[ELEMENTS_MAX_AMOUNT];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (getIndex(key) != -1) {
+            values[getIndex(key)] = value;
```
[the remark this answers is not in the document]
MaliukDaria: What about this comment? It is quite important one

MaliukDaria (approved):
GJ👍🏻

PR #601, opened 2022-05-31
Created storage

MarkGodnyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
+    private int storageSize = 0;
+    private StorageImpl<K, V>[] storageBefore = new StorageImpl[1];
```
MaliukDaria, line 9: we can use two arrays to store keys and values, or one array that store same Pair<K,V> object. Lets use one of these two approaches

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private static int pairCount;
```
MaliukDaria, line 7: can we create a solution without this variable?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private int size;
+    @SuppressWarnings("unchecked")
+    private final Pair<K, V>[] pairs = new Pair[MAX_STORAGE_SIZE];
+
+    private static class Pair<K, V> {
```
Nick97-git, line 11: private classes, methods should follow after public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
     public int size() {
+        return size;
+    }
+
+    private boolean compareKeys(K key, K otherKey) {
+        return (key == null && otherKey == null) || (key != null && key.equals(otherKey));
+    }
+
+    private int getIndex(K key) {
+        StorageImpl<K, V> storage = new StorageImpl<>();
```
Nick97-git, line 49: do you need to create a new instance of this class? Can you rewrite your logic, so you won't need to do this?

Nick97-git (changes requested):

Nick97-git (approved):

PR #602, opened 2022-06-02
Homework task implemented

Illinois357:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+        keys = new Object[MAX_CAPACITY];
+        values = new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getIndex(key);
+        if (index == -1) {
+            throw new RuntimeException("The storage is already full");
+        }
```
DimaPodakin, line 22: I think we can move this to a separate method

DimaPodakin (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Illinois357: In my opinion changing the exact exception throw to the method reference, that will do the same wouldn't simplify the code

DimaPodakin (approved):

boroda4436 (approved):

PR #609, opened 2022-06-15
complete task

zipic:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int numbersSize;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[numbersSize] = key;
+            values[numbersSize] = value;
+            numbersSize++;
+        } else {
+            values[numbersSize - 1] = value;
```
MaliukDaria, line 23: why `values[numbersSize - 1] = value;` ? We need to set value to index with same key, `not numbersSize - 1`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_ITEMS_NUMBER];
+        values = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getIndex(key);
+        if (get(key) == null) {
```
MaliukDaria, line 19: `getIndex` and `get()` methods are really heavy, could we use such heavy method only ones here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                return i;
+            }
+            if (keys[i] == null && values[i] == null) {
+                return i;
+            }
+        }
         return -1;
     }
```
MaliukDaria, line 53:
```suggestion
    private int getIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || (key != null && key.equals(keys[i]))) {
                return i;
            }
        }
        return -1;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            values[index] = value;
+            size++;
+        } else {
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
MaliukDaria, line 30: could we use here our getIndex(K key) method?

MaliukDaria (changes requested):

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                return (V) values[i];
+            }
+        }
         return null;
```
MaliukDaria, line 35:
```suggestion
        int index = getIndex(key);
        return index != -1 ? (V) values[index] : null;
```

MaliukDaria (approved):
not all comments were fixed, but I can't write about it indefinitely

PR #614, opened 2022-06-30
Task Storage for checking.

Mykola-Kutsyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
```
MaliukDaria, line 23: why `size - 1` we could have the element with same key at every position

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Mykola-Kutsyi: When I make cod without "-1" test wrong. 
BUILD FAILURE

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
MaliukDaria, line 41: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html): Don’t use class Objects

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Mykola-Kutsyi: зображення
зображення
зображення

When I directly compare the keys - NullPointerException pops up, but I can't fix it.
I tried to rewrite Objects.equals - it does not compile.
I am asking for help.

Nick97-git, line 41:
```suggestion
            if (keys[index] == key && keys[index] == null
                    || (keys[index] != null && keys[index].equals(key))) {
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Mykola-Kutsyi: That's what I tried to do, but I got confused. 
Thank you.

Nick97-git (approved):

PR #615, opened 2022-07-08
added Box class, implemented key-value method with key validationÑÑ

sergeiburya:


Nick97-git (approved):
next time, it's better not to push Main and custom class for tests

sergeiburya:
Ok, thanks.

PR #617, opened 2022-07-13
implementation of the task condition

maksym-bakunov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
     public int size() {
-        return -1;
+        int size = 0;
+        for (int j = 0; j < MAX_ITEMS_NUMBER; j++) {
+            if (keys[j][0] == null & keys[j][1] == null) {
+                break;
+            }
+            size++;
+        }
+        return size;
```
MaliukDaria, line 56: lets create size parameter of StorageImpl class, that will contain info about number of elements, containing in storage, we could increment it every time we add new element(key-value pair)

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private static final int PAIR = 2;
+    private final Object [][] keys = new Object[MAX_ITEMS_NUMBER][PAIR];
+    private int size;
+
+    public StorageImpl() {
+        size = 0;
+    }
```
MaliukDaria, line 13: I don't think we need this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (size < MAX_ITEMS_NUMBER) {
```
MaliukDaria, line 17: we dont need this check `if (size < MAX_ITEMS_NUMBER) {`

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
MaliukDaria: what about this comment? I think we don't need this constructor 
```
    public StorageImpl() {
        size = 0;
    }
```

as size variable will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        int index = find(key);
+        if (index != -1) {
+            return (V) keys[index][1];
+        }
         return null;
     }
 
+    public int find(K key) {
+        int index = -1;
+
```
MaliukDaria, line 37: redundant empty line

MaliukDaria (approved):

PR #619, opened 2022-07-15
Implemented storageImpl

ssplot56:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private static final int PAIR = 2;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private final Object[][] storage = new Object[MAX_ELEMENTS_NUMBER][PAIR];
+    private int storageSize = 0;
```
MaliukDaria, line 11:
```suggestion
    private int size;
```

It will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        int index = -1;
+        for (int i = 0; i < MAX_ELEMENTS_NUMBER; i++) {
+            if (key != null && key.equals(storage[i][KEY_INDEX])
+                    || (key == null && storage[i][KEY_INDEX] == null
+                    && storage[i][VALUE_INDEX] != null)) {
+                index = i;
+                break;
+            }
+        }
+        return index;
```
MaliukDaria, line 49:
```suggestion
    public int getIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (key != null && key.equals(storage[i][KEY_INDEX])
                    || (key == null && storage[i][KEY_INDEX] == null
                    && storage[i][VALUE_INDEX] != null)) {
                return i;
            }
        }
        return -1;
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #622, opened 2022-07-22
Jv-storage hw.

denis-kravziuk:
Implemented the Storage of key-value type.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,43 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_OF_ELEMENTS = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_NUMBER_OF_ELEMENTS];
+        values = (V[]) new Object[MAX_NUMBER_OF_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
+        }
     }
 
     @Override
     public V get(K key) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,43 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
```
MaliukDaria, line 25: why `values[size - 1]`?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,43 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
MaliukDaria, line 32: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html): Don’t use class Objects

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
denis-kravziuk: By mistake, in line 23 - size++; i incremented the value of the variable size,  so the reason was to eliminate that step. Now it's clear that only one block of if-else statement would properly change the variable.

src/main/java/core/basesyntax/impl/StorageImpl.java
denis-kravziuk: My bad. That was the last line in checklist, lack of attention )

src/main/java/core/basesyntax/impl/StorageImpl.java
denis-kravziuk: Well, the idea was that after one put action we have stored one pair (key & value) and if in new pair the new key equals key in storage in order to replace the value we should eliminate step in line 23(size++). 
Tests run witthout failures. After you comment i checked the tests and my code build because there were only two pairs in the test (addTwoElementsWithSameKey()).
Now it's clear that if pairs with same key won't be put in sequence my bisness logic will collapse (
So i created new private method to get proper index in array values to replace the value.

Nick97-git (approved):

PR #623, opened 2022-07-23
Solution created.

tumanUk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (key != null && key.equals(keys[i]))) {
+                values[i] = value;
+                return;
+            }
+        }
```
MaliukDaria, line 23: It looks like we have similar logic in `get(K key)` method, could we move it in separate private method?

MaliukDaria (changes requested):

MaliukDaria (approved):
Good Job🔥

PR #624, opened 2022-07-25
Task completed v.1.0

ghost1975:


ghost1975:
please check the completion of the task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private StorageImpl<K, V>[] storage = new StorageImpl[10];
+    private K key;
+    private V value;
+    private byte size = 0;
```
```suggestion
    private byte size;
```

it will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private StorageImpl<K, V>[] storage = new StorageImpl[10];
+    private K key;
+    private V value;
```
MaliukDaria, line 8: I think It's better to implement our storage using 2 arrays(one to store keys and other to store values), or 1 array with some inner Pair class

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
 
     @Override
     public V get(K key) {
+        for (StorageImpl<K, V> currentItem : storage) {
+            if (currentItem != null && key == null && currentItem.key == null) {
+                return currentItem.value;
+            }
+            if (currentItem != null && currentItem.key != null && currentItem.key.equals(key)) {
+                return currentItem.value;
+            }
```
KostyaNikiforov, line 46:
```suggestion
            if (currentItem.key == key || currentItem.key != null && currentItem.key.equals(key)) {
                return currentItem.value;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < storage.length; i++) {
+            if (storage[i] != null && key == null && storage[i].key == null) {
+                storage[i].value = value;
+                break;
+            }
+            if ((storage[i] != null && storage[i].key != null) && (storage[i].key.equals(key))) {
+                storage[i].value = value;
+                break;
+            }
```
KostyaNikiforov, line 29: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private StorageImpl<K, V>[] storage = new StorageImpl[10];
```
KostyaNikiforov, line 6: The name of array must be plural. And also it would be better to use lists of objects instead of storages, because storage contains methods that we don't need to use on each element. Our code should be simple and safety

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
+            this.value = value;
+        }
+    }
+
+    public StorageImpl() {
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < pairs.length; i++) {
```
MaliukDaria, line 36:
```suggestion
        for (int i = 0; i < size; i++) {
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ghost1975: There is one point here. It should be like this, for (int i = 0; i < size; i++) { so that all elements of the array are processed.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+                    || (pairs[i].getKey() != null && pairs[i].getKey().equals(key)))) {
+                pairs[i].setValue(value);
+                break;
+            }
+            if (pairs[i] == null) {
+                pairs[i] = new Pair<>(key, value);
+                size++;
+                break;
+            }
+        }
```
MaliukDaria, line 46:
```suggestion
        for (int i = 0; i < size; i++) {
            if ((key == pairs[i].getKey() )
                    || (pairs[i].getKey() != null && pairs[i].getKey().equals(key))) {
                pairs[i].setValue(value);
                return;
            }
        }
        pairs[size] = new Pair<>(key, value);
        size++;
```

MaliukDaria (changes requested):

MaliukDaria (commented):
And you should not implement homework in masterbranch, you should create separate branch

src/main/java/core/basesyntax/impl/StorageImpl.java
ghost1975: Please explain why we are using return in this fragment? I understand that it doesn't actually return anything, but why not use a break in that case?

MaliukDaria: | As far as I understand, this construction simply allows me to exit the method and finish executing its code.

exactly) By calling `return` in void method you can just stop method execution.

MaliukDaria (approved):

PR #625, opened 2022-07-25
StorageImpl_1ver

Maksym-Kulakov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_PAIRS_NUMBER = 10;
+    private K key;
+    private V value;
+    private StorageImpl<K, V>[] pairs;
```
KostyaNikiforov, line 9: It is better to use arrays of objects because the storage has a lot of unnecessary functions. Your code should be simple and safety

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+
+    public StorageImpl() {
+        pairs = new StorageImpl[MAX_PAIRS_NUMBER];
+    }
+
+    public StorageImpl(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
```
KostyaNikiforov, line 18: It will be really weird for the user of your storage to see this constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+    public int getIndex(K key) {
+        for (int index = 0; index < MAX_PAIRS_NUMBER; index++) {
+            if (pairs[index] != null && pairs[index].key == key) {
+                return index;
+            } else if (pairs[index] != null && pairs[index].key != null
+                    && pairs[index].key.equals(key)) {
+                return index;
+            } else if (pairs[index] == null) {
+                return index;
+            }
```
KostyaNikiforov, line 53: Try to simplify it. Check the Objects.equals method to see the example of comparing objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        }
         return -1;
     }
 }
+
+
+
```
KostyaNikiforov, line 60: Remove empty lines)

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: changed to private

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: deleted

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Arrays of objects were used

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Objects.equals was used

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_PAIRS_NUMBER = 10;
+    private K key;
+    private V value;
```
MaliukDaria, line 9: why do we need `key` and `value` variables? Could we implement a solution without them?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_PAIRS_NUMBER];
+        values = new Object[MAX_PAIRS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int index = 0; index < MAX_PAIRS_NUMBER; index++) {
+            if (Objects.equals(keys[index], key)) {
```
MaliukDaria, line 21: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html): Don’t use class Objects
Kostya proposed to look at `Objects.equals()` method internal implementation and use it, not `Objects.equals()` itself

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (int index = 0; index < MAX_PAIRS_NUMBER; index++) {
+            if (keys[index] == null && values[index] == null) {
+                count++;
+            }
+        }
+        return MAX_PAIRS_NUMBER - count;
```
MaliukDaria, line 50: I think It's better to create a separate `size` class level variable and increment it every time you add new element to storage

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+                values[index] = value;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int index = 0; index < MAX_PAIRS_NUMBER; index++) {
+            if (Objects.equals(keys[index], key)) {
```
MaliukDaria, line 35: same, do not use `Objects` class in your implementation

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+    private Object[] values;
+
+    public StorageImpl() {
+        keys = new Object[MAX_PAIRS_NUMBER];
+        values = new Object[MAX_PAIRS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int index = 0; index < MAX_PAIRS_NUMBER; index++) {
```
MaliukDaria, line 20:
```suggestion
        for (int index = 0; index <size; index++) {
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Yes) Removed as well) Before really I thought that it`s necessary to declare all fields fm class parametrasation.

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Size class with static increment val created

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Objects class using avoided

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: removed fm method _get

Maksym-Kulakov: but can not be removed from method _put 
as before adding first elements to arrays, size = 0 and for(...) will be interrupted

src/main/java/core/basesyntax/impl/Size.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.impl;
+
+public class Size {
+    protected static int sizeCounter;
+}
```
MaliukDaria, line 5: Not size class...  `separate size class level variable` just separate variable in StorageImpl class to store size value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                    || (keys[index] != null && keys[index].equals(key))) {
+                values[index] = value;
+                break;
+            } else if (keys[index] == null && values[index] == null) {
+                keys[index] = key;
+                values[index] = value;
+                Size.sizeCounter++;
+                break;
+            }
+        }
```
```suggestion
            for (int index = 0; index < size; index++) {
      if ((keys[index] == key)
          || (keys[index] != null && keys[index].equals(key))) {
        values[index] = value;
        return;
      }
    }
    keys[size] = key;
    values[size] = value;
    size++;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/Size.java
Maksym-Kulakov: ok)) class removed - variable added to Stor..class

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: Changed accordingly

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_PAIRS_NUMBER = 10;
+    private static int sizeCounter;
```
MaliukDaria, line 7:
```suggestion
    private int size;
```

why static? If i created 2 instances of StorageImpl class to store Cats and Dogs do they need to have the same size value?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private Object[] keys;
+    private Object[] values;
+
+    public StorageImpl() {
+        keys = new Object[MAX_PAIRS_NUMBER];
+        values = new Object[MAX_PAIRS_NUMBER];
+        sizeCounter = 0;
+    }
```
MaliukDaria, line 15:
```suggestion
    public StorageImpl() {
        keys = new Object[MAX_PAIRS_NUMBER];
        values = new Object[MAX_PAIRS_NUMBER];
    }
```

size will be initialized by 0 by default

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: changed to instance level

src/main/java/core/basesyntax/impl/StorageImpl.java
Maksym-Kulakov: removed fm Constructor

MaliukDaria (approved):
👍🏻

PR #632, opened 2022-07-30
class StorageImpl was implemented

mikhail-maiorov:
-method add() was implemented
-method put() was implemented
-method size() was implemented

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,56 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int keyCells = 0;
+    private static final int valueCells = 1;
+    private static final int GENERICS_TYPE_COUNT = 2;
+    private static final int MAX_ELEMENTS_COUNT = 10;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        @SuppressWarnings("unchecked") V value = (V) storage[VALUE_CELLS][index];
+        return value;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int checkArrayForExistKey(K key) {
```
KseniiaMakarova, line 44:
```suggestion
    private int findIndex(K key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = checkArrayForExistKey(key);
+        if (index == -1) {
+            return null;
+        }
+        @SuppressWarnings("unchecked") V value = (V) storage[VALUE_CELLS][index];
```
KseniiaMakarova, line 35: does it work without this annotation?)

KseniiaMakarova (changes requested):

PR #634, opened 2022-07-30
Develop

Oleksandr-Tovstyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE = 10;
+    private Storage[] storage;
```
KseniiaMakarova, line 7: this is not good that you have field of type Storage in Storage
let's create inner class StorageNode with key and value fields?

or use 2 arrays - 1 for keys and 1 for values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
     public int size() {
-        return -1;
+        int counter = 0;
+        for (Storage element : storage) {
+            if (element == null) {
+                break;
+            }
+            counter++;
+        }
+        return counter;
```
KseniiaMakarova, line 67: can we store size as field and increment it when we put a new pair?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,78 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE = 10;
+    private int numberOfElements;
+    private StorageNode[] storage;
+
+    private class StorageNode<K, V> {
+        private K key;
+        private V value;
```
dSkarzh, line 12: create getter for this field as you did for `key`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,78 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE = 10;
+    private int numberOfElements;
+    private StorageNode[] storage;
```
dSkarzh, line 8:
```suggestion
    private static final int MAX_STORAGE_SIZE = 10;
    private int size;
    private StorageNode<K, V>[] data;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,78 @@
...
+        }
+        if (!rewritten) {
+            for (int i = 0; i < SIZE; i++) {
+                if (storage[i] == null) {
+                    storage[i] = new StorageNode<>(key, value);
+                    numberOfElements++;
+                    break;
+                }
+            }
+        }
```
dSkarzh, line 50:
```suggestion
        for (int i = 0; i < size; i++) {
            if (data[i].key != null && data[i].getKey().equals(key) || data[i].key == key) {
                data[i] = new StorageNode<>(key, value);
                return;
            }
        }
        data[size] = new StorageNode<>(key, value);
        size++;
    }
```

dSkarzh: you can simplify like this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,78 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < SIZE; i++) {
```
dSkarzh, line 55: just iterate to current size, not max size, and you won't need null check

src/main/java/core/basesyntax/Storage.java
```diff
@@ -6,4 +6,8 @@
     V get(K key);
 
     int size();
+
+    K getKey();
+
+    V getValue();
```
dSkarzh, line 12: why do you need this?

dSkarzh (changes requested):

src/main/java/core/basesyntax/Storage.java
dSkarzh: can you explain why you added this?

dSkarzh (approved):

PR #635, opened 2022-07-31
created solution

semwithuncle:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
dSkarzh, line 38: do not use `Objects`
go through checklist one more time please

dSkarzh (changes requested):

dSkarzh (approved):

PR #638, opened 2022-08-02
implements methods, create fields, create one more method, complete l…

veers-droid:
…ogic

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,50 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private final Object [] keys;
+    private final Object [] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_ARRAY_SIZE];
+        values = new Object[MAX_ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index == -1) {
+            this.keys[size] = key;
+            this.values[size] = value;
+            size++;
+        } else {
+            this.keys[index] = key;
+            this.values[index] = value;
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            this.keys[size] = key;
+            this.values[size] = value;
+            size++;
+        } else {
+            this.keys[index] = key;
+            this.values[index] = value;
+        }
+    }
+
+    private int getKeyIndex(K key) {
```
KseniiaMakarova, line 29: always move private methods below public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key)
+                    || (keys[i] != null && key != null && key.equals(keys[i]))) {
```
KseniiaMakarova, line 32:
```suggestion
                    || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    public StorageImpl() {
+        keys = new Object[MAX_ARRAY_SIZE];
+        values = new Object[MAX_ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index == -1) {
+            this.keys[size] = key;
```
KseniiaMakarova, line 20: avoid using `this` where it is not needed - this makes harder reading code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            this.values[size] = value;
+            size++;
+        } else {
+            this.keys[index] = key;
```
KseniiaMakarova, line 24: why do we need to reset key here?

KseniiaMakarova (changes requested):

PR #640, opened 2022-08-02
storage solution

shevalye:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                    || (keys[i] != null && key != null && keys[i].equals(key))) {
+                values[i] = value;
+                found = true;
+            }
+        }
+        if (size < 9 && !found) {
+            keys[size] = key;
+            values[size] = value;
+            size = size + 1;
+        }
```
dSkarzh, line 30: you can greatly simplify this, if you implement proper check
like in `Object.equals()`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null
+                    || (keys[i] != null && key != null && keys[i].equals(key))) {
+                return (V) values[i];
+            }
+        }
         return null;
```
dSkarzh, line 41: same here

dSkarzh (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean found = false;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i],key)) {
```
dSkarzh, line 21: i meant that you need to do the same check that `Object.equals()` manually
not just use `Object.equals()`, it is not allowed

dSkarzh (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
```
dSkarzh, line 4: your solution didn't pass codestyle check because of this unused import, remove it and I'll approve your solution, everything else is fine

dSkarzh (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size = 0;
```
Sofasmile, line 9: the default value for `int` is 0, so remove the initialization

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        values = new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean found = false;
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
+                values[i] = value;
+                found = true;
```
Sofasmile, line 22: we can use `return` to break the method execution
if you do it, you can remove the `boolean found` variable
```suggestion
                return;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+            }
+        }
+        if (size < 9 && !found) {
```
Sofasmile, line 25: `size < 9` - it's hardcoding
let's remove `if` condition at all

Sofasmile (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
shevalye: when remove get tests fails.

Noct2000 (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean found = false;
```
Sofasmile, line 18: you can remove this variable `boolean found = false;`

PR #641, opened 2022-08-02
Added inner static class Node<K,V> inside StorageImpl<K,V>. Implement…

justpopova:
…ed put(), get() and size() methods inside StorageImpl with using Node<K,V> instances, methods and variables.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,133 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static int STORAGE_SIZE = 10;
+    private Node[] storage = new Node[STORAGE_SIZE];
+    private K key;
+    private V value;
+    private int size = 0;
+
+    public StorageImpl() {
+    }
+
+    public StorageImpl(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    static class Node<K, V> {
+        private final K key;
+        private V value;
+        private Node<K, V> next;
+
+        Node(K key, V value, Node<K, V> next) {
+            this.key = key;
+            this.value = value;
+            this.next = next;
+        }
+
+        public final K getKey() {
+            return key;
+        }
+
+        public final V getValue() {
+            return value;
+        }
+
+        public final String toString() {
+            return key + "=" + value;
+        }
+
+        public final V setValue(V newValue) {
+            V oldValue = value;
+            value = newValue;
+            return oldValue;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+
+            if (obj instanceof Node) {
+                Node entry = (Node) obj;
+
+                return key.equals(entry.getKey())
+                        && value.equals(entry.getValue());
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            int hash = 13;
+            hash = 17 * hash + ((key == null) ? 0 : key.hashCode());
+            hash = 17 * hash + ((value == null) ? 0 : value.hashCode());
+            return hash;
+        }
+    }
+
     @Override
     public void put(K key, V value) {
+        Node<K, V> entry = new Node<K, V>(key, value, null);
+        int index = index(key);
+
+        Node<K, V> existing = storage[index];
+        if (existing == null) {
+            storage[index] = entry;
+            size++;
+        } else {
+            while (existing.next != null) {
+                if (existing.key.equals(key)) {
+                    existing.value = value;
+                    return;
+                }
+                existing = existing.next;
+            }
+
+            if ((existing.key == getKey()) || existing.getKey().equals(key)) { //equals?
+                existing.value = value;
+            } else {
+                existing.next = entry;
+                size++;
+            }
+        }
+
     }
 
     @Override
     public V get(K key) {
+        Node<K, V> entry = storage[index(key) % STORAGE_SIZE];
+
+        while (entry != null) {
+            if ((entry.key == getKey()) || entry.getKey().equals(key)) {
+                return entry.value;
+            }
+            entry = entry.next;
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    @Override
+    public K getKey() {
+        return key;
+    }
+
+    @Override
+    public V getValue() {
+        return value;
+    }
+
+    private int index(K key) {
+        if (key == null) {
+            return 0;
+        }
+        return Math.abs(key.hashCode() % STORAGE_SIZE);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
...
+
+    public StorageImpl() {
+    }
+
+    public StorageImpl(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    static class Node<K, V> {
```
KseniiaMakarova, line 20: let's add access modifier and move this class to the bottom

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static int STORAGE_SIZE = 10;
+    private Node[] storage = new Node[STORAGE_SIZE];
+    private K key;
+    private V value;
```
KseniiaMakarova, line 9: I don't think that we need these fields - they do not describe the class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```
default value is 0

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
...
+    public K getKey() {
+        return key;
+    }
+
+    @Override
+    public V getValue() {
+        return value;
+    }
+
+    private int index(K key) {
```
KseniiaMakarova, line 129:
```suggestion
    private int getIndex(K key) {
```
use verbs for method names

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
...
+
+    public StorageImpl(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    static class Node<K, V> {
+        private final K key;
+        private V value;
+        private Node<K, V> next;
```
KseniiaMakarova, line 23: maybe let's not use `next` for now?
your solution looks very similar to hashMap realization, if you implemented this by yourself - it's great, and let's leave it as it is)
but in you don't yet fully understand solution - let's maybe just add new Node to the end of Node array?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,133 @@
...
+        } else {
+            while (existing.next != null) {
+                if (existing.key.equals(key)) {
+                    existing.value = value;
+                    return;
+                }
+                existing = existing.next;
+            }
+
+            if ((existing.key == getKey()) || existing.getKey().equals(key)) { //equals?
```
KseniiaMakarova, line 91: let's remove comments
and don't use getters where you have direct access to the variable

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static int STORAGE_SIZE = 10;
+    private Node[] storage = new Node[STORAGE_SIZE];
```
KseniiaMakarova, line 7: let's move initialization to constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+    private int size;
+
+    public StorageImpl() {
+    }
+
+    private static class Node<K, V> {
```
KseniiaMakarova, line 13: move this class to the bottom of the file

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+            return key;
+        }
+
+        public final V getValue() {
+            return value;
+        }
+
+        public final String toString() {
+            return key + "=" + value;
+        }
```
KseniiaMakarova, line 32: let's remove final - we won't override these methods anyway

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+
+            if (obj instanceof Node) {
+                Node entry = (Node) obj;
+
+                return key.equals(entry.getKey())
+                        && value.equals(entry.getValue());
```
KseniiaMakarova, line 44: let's properly check each field (if both are null - then they should be considered equal)
and don't allow NPE in your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            int hash = 13;
+            hash = 17 * hash + ((key == null) ? 0 : key.hashCode());
+            hash = 17 * hash + ((value == null) ? 0 : value.hashCode());
```
KseniiaMakarova, line 53:
```suggestion
            hash = 17 * hash + (key == null ? 0 : key.hashCode());
            hash = 17 * hash + (value == null ? 0 : value.hashCode());
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+            return hash;
+        }
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < STORAGE_SIZE; i++) {
```
KseniiaMakarova, line 60: should we iterate over whole array or only filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+            if (storage[i] != null) {
```
KseniiaMakarova, line 61: let's remove this check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+                storage[i] = new Node(key, value);
+                size++;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (Node node : storage) {
```
KseniiaMakarova, line 76: should we iterate over whole array or only filled part?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
...
+            if (node != null) {
```
KseniiaMakarova, line 77: let's remove this check

KseniiaMakarova (changes requested):

Sofasmile (changes requested):
be attentive to all comments, some of the previous comments were not fixed

Sofasmile (approved):

PR #644, opened 2022-08-03
Hw solution is developed

amarkelovUA:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
+
```
dSkarzh, line 53: you can do it in one loop and one if check, take a look at how `Objects.equals()` compares objects
implement same logic here

dSkarzh (changes requested):

dSkarzh (approved):
🔥

PR #648, opened 2022-08-03
Using HashMap

Litvik21:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,27 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private HashMap<K, V> storageMap = new HashMap<>();
```
dSkarzh, line 7: please see common mistakes list and check yourself

dSkarzh (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Litvik21: Ok, i have ready made solution with using two arrays

Litvik21: how can i do it? 
i dont know, i tried delete first two branches and created new branch with using two arrays and push it, but it doesnt have connection with google table.

PR #651, opened 2022-08-04
done Storage hw

Vasul-Pankiv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+        firstEmptyIndex = index + 1;
+    }
+
+    private int contains(int firstEmptyIndex, K key) {
+        for (int i = 0; i < firstEmptyIndex; i++) {
+            if (key == keyArray[i] || (key != null && key.equals(keyArray[i]))) {
+                return i;
+            }
+        }
+        return -1;
```
Noct2000, line 44: private methods must be after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+
+    public StorageImpl() {
+        firstEmptyIndex = 0;
+        keyArray = (K[]) Array.newInstance(Object.class, MAX_ARRAY_LENGTH);
+        valueArray = (V[]) Array.newInstance(Object.class, MAX_ARRAY_LENGTH);
+    }
+
     @Override
     public void put(K key, V value) {
+        int containsIndex = contains(firstEmptyIndex, key);
```
Noct2000, line 20: you can use your field directly
```suggestion
        int containsIndex = contains(key);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+        if (containsIndex != -1) {
+            replace(containsIndex, value);
+        } else if (firstEmptyIndex < MAX_ARRAY_LENGTH) {
+            replace(firstEmptyIndex, key, value);
```
Noct2000, line 24: I think replacing the value will be enough 
```suggestion
            replace(firstEmptyIndex, value);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private K[] keyArray;
+    private V[] valueArray;
+    private int firstEmptyIndex;
+
+    public StorageImpl() {
+        firstEmptyIndex = 0;
+        keyArray = (K[]) Array.newInstance(Object.class, MAX_ARRAY_LENGTH);
+        valueArray = (V[]) Array.newInstance(Object.class, MAX_ARRAY_LENGTH);
```
Noct2000, line 15: I think using `new Object[]` would be better

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+
+    private void replace(int index, K key, V value) {
+        keyArray[index] = key;
+        valueArray[index] = value;
+        firstEmptyIndex = index + 1;
+    }
+
+    private int contains(int firstEmptyIndex, K key) {
+        for (int i = 0; i < firstEmptyIndex; i++) {
+            if (key == keyArray[i] || (key != null && key.equals(keyArray[i]))) {
```
Noct2000, line 40: AND operation has higher priority than OR

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+            }
+        }
+        return -1;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return contains(firstEmptyIndex, key) == -1
+                ? null : valueArray[contains(firstEmptyIndex, key)];
```
Noct2000, line 50: it would be better to create a new index local variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private K[] keyArray;
+    private V[] valueArray;
+    private int firstEmptyIndex;
```
Noct2000, line 10: I think it would be better
```suggestion
    private int size;
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Vasul-Pankiv: I remove firstEmptyIndex from paramethers, but i need key in case when i put a new pair in storage. I have to add key to the end of keyArray

Noct2000 (approved):

PR #653, opened 2022-08-04
Add realise of storage task

Alexhqolma:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int ARRAY_SIZE = 10;
+    private K key;
+    private V value;
+    private final Object[] keys = new Object[ARRAY_SIZE];
+    private final Object[] values = new Object[ARRAY_SIZE];
+    private int exist = -1;
+    private int keysSize = -1;
```
Noct2000, line 13: I think this will be better 
```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        if (keysSize == -1) {
+            keysSize++;
+            this.keys[keysSize] = key;
+            this.values[keysSize] = value;
+        } else {
+            for (int i = 0; i < keysSize + 1; i++) {
+                if (key != null && key.equals(keys[i]) || key == keys[i]) {
+                    exist = i;
+                }
+            }
```
Noct2000, line 26: I think you can create a private method that returns the index of the element in the keys

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                this.keys[keysSize] = key;
+                this.values[keysSize] = value;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key != null && key.equals(keys[i]) || key == keys[i]) {
```
Noct2000, line 41:
```suggestion
            if (key == keys[I] || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                keysSize++;
+                this.keys[keysSize] = key;
+                this.values[keysSize] = value;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
Noct2000, line 40: you can use your size 
```suggestion
        for (int i = 0; i < size; i++) {
```

src/main/java/core/basesyntax/Cat.java
```diff
@@ -0,0 +1,55 @@
+package core.basesyntax;
+
+import java.util.Objects;
+
+public class Cat {
```
Noct2000, line 5: I think this class is redundant

Noct2000 (changes requested):

src/main/java/core/basesyntax/Cat.java
```diff
@@ -0,0 +1,55 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Cat cat = (Cat) o;
+        return Objects.equals(name, cat.name)
+                && Objects.equals(color, cat.color);
+    }
```
Noct2000, line 41: fix naming here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int ARRAY_SIZE = 10;
+    private K key;
+    private V value;
```
Noct2000, line 9: where are you using these two fields?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+        return size;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key != null && key.equals(keys[i]) || key == keys[i]) {
+                exist = i;
+            }
+        }
+        return exist;
```
Noct2000, line 59: I think exist field is redundant and you can return i directly

Noct2000 (changes requested):
if you don't use class Cat, you can remove it

Noct2000 (changes requested):

Noct2000 (approved):

PR #666, opened 2022-08-06
first solution

fliarkovskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (V value : values) {
+            if (value != null) {
+                count++;
+            }
+        }
+        return count;
```
Noct2000, line 50: here you could return this `private int size`

Noct2000 (changes requested):

Noct2000 (approved):

PR #680, opened 2022-08-07
storage-task

EagleAndrew:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,37 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_RANGE = 10;
+    private int size;
+    private final K[] keys = (K[]) new Object[MAX_RANGE];
+    private final V[] values = (V[]) new Object[MAX_RANGE];
```
Sofasmile, line 10: let's initialize these fields in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,37 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (get(key) == null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
+        }
```
Sofasmile, line 20: does this logic work correctly?
we should replace the value if the storage already contains this key
But now you are replacing the last element in the storage

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,37 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_RANGE; i++) {
+            if (Objects.equals(keys[i], key)) {
```
Sofasmile, line 26: the last point in checklist
https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#:~:text=Don%E2%80%99t%20use%20class%20Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,37 @@
...
+            values[size] = value;
+            size++;
+        } else {
+            values[size - 1] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_RANGE; i++) {
```
Sofasmile, line 25: should we iterate over whole array or only filled part?

Sofasmile (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
+                    values[i] = value;
+                }
+            }
+        }
```
Sofasmile, line 28:
```suggestion
            for (int i = 0; i < size; i++) {
                if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
                    values[i] = value;
                    return;
                }
            }
            keys[size] = key;
            values[size] = value;
            size++;
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #681, opened 2022-08-08
Was implemented Storage of key-value by 10 objects

valentyn-slutskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if ((keys[i] == null && values[i] == null)
+                    || keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
Noct2000, line 20:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public int size() {
-        return -1;
+        int counter = 0;
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (keys[i] != null || values[i] != null) {
+                counter++;
+            }
+        }
+        return counter;
```
Noct2000, line 46: I think would be better to create `private int size` field, init it in the constructor and change it in your methods

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 
     @Override
     public int size() {
-        return -1;
+        /*int counter = 0;
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (keys[i] != null || values[i] != null) {
+                counter++;
+            }
+        }*/
```
Noct2000, line 51: don't leave the old code as a comment!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
```
Noct2000, line 19:
```suggestion
        for (int i = 0; i < size; i++) {
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i <= size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                size += values[i] == null ? 1 : 0;
```
Noct2000, line 21: this is redundant 
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                keys[i] = key;
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
```
Noct2000, line 25: there too
```suggestion
            } else {
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        int index = size;
+        for (int i = 0; i <= size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                index = i;
+                break;
+            }
+        }
+        keys[index] = key;
+        values[index] = value;
+        size += index == size ? 1 : 0;
```
Noct2000, line 28: if I want to rewrite my value by some key what should I do?

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
valentyn-slutskyi: You should use this code, I think everything will be well. By default, index equals size, it means if there is no such key in the array, this key will written in the end of array (26 and 27 lines) and size will increase by 1 (28 line). But, if this key already exist in array, 22 line rewrite index, and then 26 and 27 lines rewrite key and value, also this case 28 lines will not increase size.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = size;
+        for (int i = 0; i <= size; i++) {
```
Noct2000, line 20:
```suggestion
        for (int i = 0; i < size; i++) {
```

Noct2000 (approved):

PR #685, opened 2022-08-08
first commit to hw-storage

YarlVas:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+                return result;
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size++;
```
Noct2000, line 42: what happens if I check my size multiple times?

Noct2000 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
YarlVas: I thought size() was a method that was called from the put method, adding one each time to the index of the array cell for the new values

Noct2000 (approved):

PR #693, opened 2022-08-08
create_strorage_methods

Jhuzeppe:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[LENGTH];
+        this.values = new Object[LENGTH];
```
Sofasmile, line 13: we should use `this` only if we have the same naming for field and input parameter or local variable
```suggestion
        keys = new Object[LENGTH];
        values = new Object[LENGTH];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size;i++) {
```
Sofasmile, line 18:
```suggestion
        for (int i = 0; i < size; i++) {
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #695, opened 2022-08-08
implemented methods get, put, size in class StorageImpl.java

azarovartur:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final String EMPTY_KEY = "empty";
```
MaliukDaria, line 7: why do we need this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+    private final Object[] keys = new Object[10];
+    private final Object[] values = new Object[10];
```
MaliukDaria, line 9: lets make number 10 as constant field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+
+    {
+        Arrays.fill(keys,EMPTY_KEY);
```
MaliukDaria, line 12: I really dont think we need this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] != null && keys[i].equals(EMPTY_KEY)) {
+                return i;
+            }
+        }
+        return keys.length;
     }
```
MaliukDaria, line 54: Lets create some `size` variable, where we can store info about how many elements are currently in out storage and use this variable in loops of `put` and `get` methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+                values[i] = value;
+                break;
+            }
+            if (keys[i] != null && keys[i].equals(EMPTY_KEY)) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
+        }
     }
```
MaliukDaria, line 28:
```suggestion
        for (int i = 0; i < size ; i++) {
            if ((key  == keys[i]) || (keys[i] != null && keys[i].equals(key))) {
                values[i] = value;
                return;
            }
        }
                keys[size] = key;
                values[size] = value;
                size++;
            }
    }
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: Ok, but in readme for this task, i was should realize the method size() - that returns storage size
Or i should realize method size() as getter for size variable?

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: Agreed, it was my bad idea how to implement difference between null key (when we use null as key for data in array VALUES) and empty key (without data in array VALUES)

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: done

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: I agree

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final Object[] keys = new Object[MAX_STORAGE_SIZE];
+    private final Object[] values = new Object[MAX_STORAGE_SIZE];
```
```suggestion
    private final K[] keys;
    private final V[] values;
    ....
    
    public StorageImpl() {
        keys = (K[]) new Object[MAX_STORAGE_SIZE];
        values = (V[]) new Object[MAX_STORAGE_SIZE];
    }
```
it is better to move initialization to constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (key == keys[i]) {
+                    return (V) values[i];
+                }
+                continue;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                return (V) values[i];
+            }
```
aaananas, line 35:
```suggestion
            if (key == keys[i] || (keys[i] != null && keys[i].equals(key))) {
                return values[i];
            }
```
You can easily simplify if's

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final Object[] keys = new Object[MAX_STORAGE_SIZE];
+    private final Object[] values = new Object[MAX_STORAGE_SIZE];
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key == keys[i]) || (keys[i] != null && keys[i].equals(key))) {
```
aaananas, line 14:
```suggestion
            if (key == keys[i] || (keys[i] != null && keys[i].equals(key))) {
```

aaananas (changes requested):
Great job

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: Thank you, I'll keep it on mind

src/main/java/core/basesyntax/impl/StorageImpl.java
azarovartur: Really) looks much more simple
Thanks

aaananas (approved):

PR #696, opened 2022-08-10
First solution of jv-storage homework

slvkobzar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
     @Override
     public int size() {
-        return -1;
+        int storageSize = 0;
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] != null || values[i] != null) {
+                storageSize++;
+            }
+        }
+        return storageSize;
```
MaliukDaria, line 58: lets create some `size` variable in this class witch will store number of elements in storage

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+            }
+            if (key.equals(keys[i])) {
+                values [i] = value;
+                break;
+            }
+            if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
```
MaliukDaria, line 29: This method could be much simpler with `size` variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+                if (keys[i] == null) {
+                    return (V) values[i];
+                } else {
+                    continue;
+                }
+            }
+            if (key.equals(keys[i])) {
+                return (V) values[i];
+            }
+        }
```
MaliukDaria, line 46: as well as this one

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
```
aaananas, line 47:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
                return (V) values[i];
            }
        }
```
can be easily simplified

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                        break;
+                    } else {
+                        continue;
+                    }
+                }
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                }
+            }
+        }
```
aaananas, line 31: better move setting key value if key is not present (first if) to the end, because if the key is actually present, you would iterate over the whole storage in method get(key) and next in the second if
Also think about simplifying second if))

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys = new Object[MAX_ITEMS_NUMBER];
+    private Object[] values = new Object[MAX_ITEMS_NUMBER];
+    private int size = 0;
+
```
```suggestion
    private K[] keys;
    private V[] values;
    private int size;

    public StorageImpl() {
        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
        values = (V[]) new Object[MAX_ITEMS_NUMBER];
    }
```
size is initialized with 0 by default
better initialize objects in constructors

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                values[i] = value;
+                keyIsPresent = true;
+            }
+        }
+        if (keyIsPresent == false) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
                values[i] = value;
                return;
            }
        }
            keys[size] = key;
            values[size] = value;
            size++;
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                return (V) values[i];
```
aaananas, line 33:
```suggestion
                return values[i];
```

aaananas (approved):

PR #697, opened 2022-08-11
Newbranch

SerhiiAN:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+            if (keys[i] == key && values[i] != null || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                size++;
+                break;
+            }
+        }
```
MaliukDaria, line 23:
```suggestion
        for (int i = 0; i <size(); i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size(); i++) {
+            if (this.keys[i] == key || this.keys[i] != null && this.keys[i].equals(key)) {
+                return values[i];
+            }
+        }
```
MaliukDaria, line 32: It looks like we have similar logic in `put(K key, V value)` method, could we move it to separate private method?

MaliukDaria (changes requested):
newBranch - not very informative name

MaliukDaria (approved):

PR #703, opened 2022-08-19
StorageImpl implemented with generics

SpaceUrgent:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+    private Cell[] cells = new Cell[STORAGE_CAPACITY];
```
```suggestion
    private Cell<K, V>[] cells;
```
move the initialization to constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+
```
```suggestion
     public StorageImpl() {
         cells = new Cell[STORAGE_CAPACITY];
     }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
     @Override
     public void put(K key, V value) {
+        int keyIndex;
+        if (size == 0 || (keyIndex = getIndexByKey(key)) < 0) {
```
```suggestion
        int keyIndex = getIndexByKey(key);
        if (keyIndex < 0) {
```
if size == 0 getIndexByKey() would return -1

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+            cells[keyIndex].setValue(value);
+        }
     }
 
     @Override
     public V get(K key) {
+        int keyIndex;
+        if ((keyIndex = getIndexByKey(key)) >= 0) {
+            return (V) cells[keyIndex].getValue();
+        }
```
```suggestion
        int keyIndex  = getIndexByKey(key);
        if (keyIndex >= 0) {
            return (V) cells[keyIndex].getValue();
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
 
     @Override
     public int size() {
+        return this.size;
+    }
+
+    private int getIndexByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if (cells[i] != null
+                    && (cells[i].key != null ? cells[i].key.equals(key) : cells[i].key == key)) {
```
```suggestion
            if (cells[i].key == key 
                     || cells[i].key != null  && cells[i].key.equals(key)) {
```
cell[i] won't be null in this implementation
if you put some cell in storage it will never be removed, only the value can be changed
so it easily can be simplified

aaananas (changes requested):
Great job

aaananas (approved):

PR #704, opened 2022-08-21
interface methods are implemented. Implementation of the task through…

oAlExAnDeRo:
… an array.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,11 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.1</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,55 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
+    private static final int MAX_ITEMS_OF_ELEMENTS = 10;
+    private final Pair<K, V>[] pairs;
+    private int size;
+
+    public StorageImpl() {
+        pairs = new Pair[MAX_ITEMS_OF_ELEMENTS];
     }
 
-    @Override
-    public V get(K key) {
+    public static class Pair<K, V> {
+        private final K key;
+        private V value;
+
+        public Pair(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+
+        public V getValue() {
+            return value;
+        }
+    }
+
+    private Pair<K, V> getPairByKey(K key) {
+        for (int i = 0; i < size; i++) {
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -25,6 +25,11 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.1</version>
+        </dependency>
```
MaliukDaria, line 32: why do we need this dependency?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+            this.key = key;
+            this.value = value;
+        }
+
+        public V getValue() {
+            return value;
+        }
+    }
+
+    private Pair<K, V> getPairByKey(K key) {
```
MaliukDaria, line 28: private methods should be under public as helper ones

MaliukDaria (changes requested):
Good Job! Left small comments

Rommelua:
Good job! But what if we call put() method more than 10 times? I think we should throw RuntimeException with the informative message "Storage is full" for example.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_OF_ELEMENTS = 10;
+    private final Pair<K, V>[] pairs;
+    private int size;
+
+    public StorageImpl() {
+        pairs = new Pair[MAX_ITEMS_OF_ELEMENTS];
+    }
+
+    public static class Pair<K, V> {
```
aaananas, line 14: this class should be private
also private classes, methods should be after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+        try {
+            Pair<K, V> pair = getPairByKey(key);
+            if (pair == null) {
+                pairs[size++] = new Pair<>(key, value);
+            } else {
+                pair.value = value;
+            }
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Storage is full", e);
+        }
```
aaananas, line 39:
```suggestion
        if (size == MAX_CAPACITY) {
            throw new RuntimeException("Storage is full", e);
        }
        Pair<K, V> pair = getPairByKey(key);
        if (pair == null) {
            pairs[size++] = new Pair<>(key, value);
        } else {
            pair.value = value;
        } 
```
better check size

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_OF_ELEMENTS = 10;
```
aaananas, line 6:
```suggestion
    private static final int MAX_CAPACITY = 10;
```

aaananas (changes requested):

aaananas (approved):

PR #705, opened 2022-08-22
Storage was implemented

not4und:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+            values[keyIndex] = value;
+            return;
+        }
+        if (size < 10) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Key: '" + key + "' not found and Storage is full.");
+        }
```
```suggestion
        if (keyIndex >= 0) {
            values[keyIndex] = value;
            return;
        } else {
             keys[size] = key;
            values[size] = value;
            size++;
        }
```

You do not need this check, and you are not asked to throw out exceptions in the problem statement

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int keyIndex = getKeyIndex(key);
+        return keyIndex >= 0 ? values[keyIndex] : null;
```
zachar-chorny, line 36: Also, put the keyIndex >= 0 check together with the getKeyIndex method call into a separate private method. It's better to do this because you use this code in both methods

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
not4und: How? What do you mean?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
```suggestion
    private int size;
```
size is 0 by default

aaananas (approved):
great job

PR #706, opened 2022-08-23
jv-storage

VolodymyrBarik:
7 out of 7 tests done

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,54 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final K[] keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+    private final V[] values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    private int storageSize = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key != null && keys[i] == null && values[i] != null) {
+                continue;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            }
+            if (key == null && keys[i] == null && value != null && values[i] == null) {
+                values[i] = value;
+                storageSize++;
+                break;
+            }
+            if (key == null && keys[i] == null && value != null && values[i] != null) {
+                values[i] = value;
+                break;
+            }
+            if (keys[i] == null && key != null) {
+                keys[i] = key;
+                values[i] = value;
+                storageSize++;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int index = 0; index < storageSize; index++) {
+            if (key == null && keys[index] == null) {
+                return values[index];
+            }
+            if (keys[index] != null && keys[index].equals(key)) {
+                return values[index];
+            }
+        }
         return null;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final K[] keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+    private final V[] values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    private int storageSize = 0;
+
     @Override
```
aaananas, line 11:
```suggestion
    private final K[] keys;
    private final V[] values;
    private int size;
    
    public StorageImpl() {
        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
       values = (V[]) new Object[MAX_ITEMS_NUMBER];
    }

    @Override
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                values[i] = value;
+                storageSize++;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int index = 0; index < storageSize; index++) {
```
aaananas, line 41:
```suggestion
        for (int i = 0; i < storageSize; i++) {
```
use i inside for

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+            if (key == null && keys[index] == null) {
+                return values[index];
+            }
+            if (keys[index] != null && keys[index].equals(key)) {
+                return values[index];
+            }
```
aaananas, line 47:
```suggestion
            if (key = keys[i] 
                        || keys[i] != null && keys[i].equals(key)) {
                return values[i];
            }
```
can be simplified

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                values[i] = value;
+                break;
+            }
+            if (keys[i] == null && key != null) {
+                keys[i] = key;
+                values[i] = value;
+                storageSize++;
+                break;
+            }
+        }
```
aaananas, line 36:
```suggestion
        for (int i = 0; i < keys.length; i++) {
             if (keys[i] == key 
                          || key != null && key.equals(keys[i])) {
                          values[i] = value;
                          return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```
can be simplified

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        values[storageSize] = value;
+        storageSize++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < storageSize; i++) {
+            if (key == keys[i]
+                    || keys[i] != null && keys[i].equals(key)
+                    || key == null && keys[i] == null && values[i] != null) {
```
Rommelua, line 35:
```suggestion
                    || key == null && keys[i] == null) {
```

Rommelua (changes requested):
Good job. Think about what if we call put() method more than 10 times? I think we should throw RuntimeException with the informative message "Storage is full" for example.

aaananas (changes requested):

aaananas (approved):

PR #707, opened 2022-08-23
Solution was added

Kostiantyn-Sacheiev:
1. Added constructor
2. Realized method size()
3. Created method findKeyIndex() which return index of the key on array or return - 1 if key don't find
4. Realized methods put() and get()

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -2,18 +2,47 @@ package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
 
+import java.util.Objects;
+
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+            size++;
+        } else {
+            values[findKeyIndex(key)] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return findKeyIndex(key) == -1 ? null : (V) values[findKeyIndex(key)];
```
aaananas, line 30:
```suggestion
        int index = findKeyIndex(key);
        return index == -1 ? null : (V) values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (findKeyIndex(key) == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[findKeyIndex(key)] = value;
+        }
```
aaananas, line 25:
```suggestion
        int index = findKeyIndex(key) == -1;
        if (index == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
        } else {
            values[index] = value;
        }
```

aaananas (changes requested):

aaananas (approved):

PR #709, opened 2022-08-24
Homework storage solution

dmytro-stahiv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys = (K[]) new Object[MAX_SIZE];
+    private V[] values = (V[]) new Object[MAX_SIZE];
+    private int size = 0;
```
```suggestion
    private int size;
```

`size` variable will be initialized by 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null) {
+                values[i] = value;
+                return;
+            }
+
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
```
MaliukDaria, line 22: could we combine these `if` checks?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && key == null) {
+                return values[i];
+            }
+
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                return values[i];
+            }
```
MaliukDaria, line 39: same, lets combine these `ifs`, and think if we can move common logic from get and put methods to separate private method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (int i = 0; i < values.length; i++) {
+            if (values[i] != null) {
+                count++;
+            }
+        }
+        return size;
```
MaliukDaria, line 52:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
+
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
MaliukDaria, line 32:
```suggestion
        for (int i = 0; i < size; i++) {
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
dmytro-stahiv: I create private method, but I cannot use it correctly in put and get methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys = (K[]) new Object[MAX_SIZE];
+    private V[] values = (V[]) new Object[MAX_SIZE];
```
Rommelua, line 8: See [Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html) - All non-constant fields should be initialized in the constructor

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+    private K key;
+    private V value;
```
zachar-chorny, line 11: Why do you need these fields?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+
+    private V info(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null)
+                    || ((keys[i] == key) || (keys[i] != null && keys[i].equals(key)))) {
+                return values[i];
+            }
+        }
+        return null;
     }
```
zachar-chorny, line 62:
```suggestion
    private V getValueByKey(K key) {
        for (int i = 0; i < size; i++) {
            if ((keys[i] == null && key == null)
                    || ((keys[i] == key) || (keys[i] != null && keys[i].equals(key)))) {
                return values[i];
            }
        }
        return null;
    }
```
You don't use this method anywhere think about what code in your methods you could replace with this method.

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (getValueByKey(key) == values[i]) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
```
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null key.equals(keys[i])) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                size++;
+            }
+        }
+        return size;
+    }
+
+    private V getValueByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null)
+                    || ((keys[i] == key) || (keys[i] != null && keys[i].equals(key)))) {
```
```suggestion
            if (keys[i] == key 
                    || keys[i] != null && keys[i].equals(key)) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (int i = 0; i < values.length; i++) {
+            if (values[i] != null) {
+                size++;
+            }
+        }
+        return size;
```
aaananas, line 42:
```suggestion
        return size;
```

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private V getValueByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key)
+                    || ((keys[i] == key) || (keys[i] != null && keys[i].equals(key)))) {
```
aaananas, line 42:
```suggestion
                    || (keys[i] != null && keys[i].equals(key)))) {
```

aaananas (approved):

PR #714, opened 2022-08-31
v1.0 created methods which put values to storage, get values from sto…

yadiver82:
…rage and calculate size of filled storage

### DIFF
--- a/src/main/java/core/basesyntax/Storage.java
+++ b/src/main/java/core/basesyntax/Storage.java
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
     void put(K key, V value);
 
     V get(K key);
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -2,18 +2,47 @@ package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked") // do not remove this line
+
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_ELEMENTS = 10;
+    private K key;
+    private V value;
+    private K[] keyArray = (K[]) new Object[MAX_ARRAY_ELEMENTS];
+    private V[] valueArray = (V[]) new Object[MAX_ARRAY_ELEMENTS];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if ((keyArray[i] == null && valueArray[i] == null)
+                    || (key != null && key.equals(keyArray[i]))
+                    || (key == null && keyArray[i] == null)) {
+                keyArray[i] = key;
+                valueArray[i] = value;
+                return;
+            }
+        }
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
aaananas, line 4:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
...
+        for (int i = 0; i < keyArray.length; i++) {
+            if ((keyArray[i] == null && valueArray[i] == null)
+                    || (key != null && key.equals(keyArray[i]))
+                    || (key == null && keyArray[i] == null)) {
+                keyArray[i] = key;
+                valueArray[i] = value;
+                return;
+            }
+        }
+        throw new RuntimeException("Can't put the data into a Storage - Storage is full ");
```
aaananas, line 25:
```suggestion
        keyArra[size] = key;
        valueArray[size] = value;
        size++;
```
if the key is not present, you put the key and value on the last free position
if you want to throw an exception for the situation when the storage is full, you can add a check at the beginning of this method. You should check if size == MAX_ARRAY_ELEMENTS

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,55 @@
...
+                    || (key == keyArray[i])) {
+                valueArray[i] = value;
+                return;
+            }
+        }
+        if (size < MAX_ARRAY_ELEMENTS) {
+            keyArray[size] = key;
+            valueArray[size] = value;
+            size++;
+        } else {
```
aaananas, line 36:
```suggestion
        for (int i = 0; i < size; i++) {
            if ((key != null && key.equals(keyArray[i]))
                    || (key == keyArray[i])) {
                valueArray[i] = value;
                return;
            }
        }
        if (size < MAX_ARRAY_ELEMENTS) {
            keyArray[size] = key;
            valueArray[size] = value;
            size++;
        } else {
```
it will work without the first if

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,55 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
+
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_ELEMENTS = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked") // do not remove this line
+
```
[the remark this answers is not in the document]
aaananas: remove this line

aaananas (changes requested):

aaananas (approved):

PR #726, opened 2022-09-17
Storage Solution

OleksandrKolisnyk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+        keys = (K[]) new Object[MAX_SIZE_STORAGE];
+        values = (V[]) new Object[MAX_SIZE_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexForPut = sizeStorage;
+        boolean isKeyExist = false;
+        for (int i = 0; i < sizeStorage; i++) {
+            if (Objects.equals(key,(keys[i]))) {
```
kozhukhovsky, line 23: image
[Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html)

kozhukhovsky (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    public StorageImpl() {
+        sizeStorage = 0;
+        keys = (K[]) new Object[MAX_SIZE_STORAGE];
+        values = (V[]) new Object[MAX_SIZE_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexForPut = sizeStorage;
+        boolean isKeyExist = false;
```
kozhukhovsky, line 20: This is redundant logic.
We still run through the array, and if we have a key, we have an index, it is enough to put a new value at this index.
If we do not have such a value, then after that it is simply enough to put the key and value in the required arrays.

kozhukhovsky (changes requested):

kozhukhovsky (approved):
Good job!

PR #734, opened 2022-09-28
Implement StorageImpl class

shmex-j:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,64 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 5: Do you really need this annotation?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,64 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STORAGE_CAPACITY = 10;
```
liliia-ponomarenko, line 7: This constant should  be private

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,64 @@
...
+                if (keys[i] == null) {
+                    values[i] = value;
+                    return;
+                }
+                break;
+            }
+            if (key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
```
liliia-ponomarenko, line 31: Can we combine this conditional?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,64 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null) {
+                return (V) values[i];
+            } else if (key == null) {
+                continue;
+            }
+            if (key.equals(keys[i])) {
+                return (V) values[i];
+            }
+        }
```
liliia-ponomarenko, line 57: the same here 
and I guess you can use method ```get()``` in method  ```put()```😉

liliia-ponomarenko (changes requested):
good solution! Let’s improve it😉

liliia-ponomarenko (approved):
good job! 🔥

PR #736, opened 2022-09-29
implemented Storage interface in the StorageImpl class according to t…

igorgalenko:
…he assignment

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,44 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int ARRAY_BOUND = 10;
+    private K[] storageKeys;
+    private V[] storageValues;
+    private int storageSize;
+    private int currentIndex;
+
+    public StorageImpl() {
+        storageKeys = (K[]) new Object[ARRAY_BOUND];
+        storageValues = (V[]) new Object[ARRAY_BOUND];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            storageValues[currentIndex] = value;
+            return;
+        }
+        storageKeys[storageSize] = key;
+        storageValues[storageSize] = value;
+        storageSize++;
     }
 
     @Override
     public V get(K key) {
+        for (currentIndex = 0; currentIndex < storageKeys.length; currentIndex++) {
+            if (key == storageKeys[currentIndex]) {
+                return storageValues[currentIndex];
+            }
+            if (storageKeys[currentIndex] != null && storageKeys[currentIndex].equals(key)) {
+                return storageValues[currentIndex];
+            }
+        }
         return null;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        storageKeys = (K[]) new Object[ARRAY_BOUND];
+        storageValues = (V[]) new Object[ARRAY_BOUND];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            storageValues[currentIndex] = value;
+            return;
+        }
```
hronograf, line 23: it is interesting solution, really cool!
also, when we use `currentIndex` across different functions it can produce bugs, 
because we have some shared state for our functions
so it would be better if we had function like `private int indexOf(K key)`

but solution is really interesting 🧠

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
 
     @Override
     public V get(K key) {
+        for (currentIndex = 0; currentIndex < storageKeys.length; currentIndex++) {
+            if (key == storageKeys[currentIndex]) {
+                return storageValues[currentIndex];
+            }
+            if (storageKeys[currentIndex] != null && storageKeys[currentIndex].equals(key)) {
+                return storageValues[currentIndex];
+            }
```
hronograf, line 37: here we have the same body for our if-statements
let's merge them, so we will have one if-statement

hronograf (changes requested):
Cool!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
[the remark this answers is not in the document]
igorgalenko: I really don't understand where it appeared from) I didn't do it on purpose

src/main/java/core/basesyntax/impl/StorageImpl.java
igorgalenko: Yeah...but after refactoring my code not so elegance :(

src/main/java/core/basesyntax/impl/StorageImpl.java
igorgalenko: Yes, I saw this rule but I decided that in this way text will be more readable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     public int size() {
-        return -1;
+        int i = 0;
+        for (; i < storageValues.length; i++) {
+            if (storageValues[i] == null) {
+                break;
+            }
+        }
+        return i;
     }
```
hronograf, line 50: `storageSize` was fine class field, because it is state of the object, not function

hronograf (changes requested):
one small change)

hronograf (approved):
LGTM! ⚙️

PR #739, opened 2022-10-02
Implement the Storage of key-value

Bodddd:
I can implement solutions with using nested class?

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,86 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPABILITY = 10;
+
+    static class StorageBox<K, V> {
+        private final K key;
+        private V value;
+
+        public StorageBox(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+
+        public K getKey() {
+            return key;
+        }
+
+        public V getValue() {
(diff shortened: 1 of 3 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,86 @@
...
+        } catch (KeyContainException e) {
+            fillStorageBox(key, value);
+        }
+    }
+
+    private void fillStorageBox(K key, V value) {
+        StorageBox[] newBoxes = new StorageBox[boxes.length + 1];
+        for (int i = 0; i < boxes.length; i++) {
+            newBoxes[i] = boxes[i];
+        }
```
ChumakAn, line 75: what about using System.arraycopy?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,86 @@
...
+        } catch (KeyContainException e) {
+            System.out.println(e.getMessage());
+            return null;
+        }
     }
 
     @Override
     public int size() {
-        return -1;
+        return boxes.length;
```
ChumakAn, line 60: it's better to create the size variable and increment/decrement it after adding/removing the element, cause your 'boxes' array is static one, once you create it, it should be of size 10, but it doesn’t mean that it initially has 10 StorageBoxes inside

ChumakAn (changes requested):
Good! It's cool you've added custom exceptions, let's make the logic of the solution better

ChumakAn (approved):
LGTM

PR #740, opened 2022-10-02
implemented task

vov4ik89:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private int storageSize = 0;
+    private final Object[] keyStorage;
+    private final Object[] valueStorage;
```
KostyaNikiforov, line 9:
```suggestion
    private final Object[] keys;
    private final Object[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private int storageSize = 0;
```
KostyaNikiforov, line 7:
```suggestion
    private int size = 0;
```
We already see the class name)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    private boolean containsKey(K key, V value) {
+        for (int i = 0; i < storageSize; i++) {
+            if (equalsObjects(keyStorage[i], key)) {
+                valueStorage[i] = value;
+                return true;
+            }
+        }
+        return false;
+    }
```
KostyaNikiforov, line 49: Let's reuse common logic  here

KostyaNikiforov (changes requested):

PR #741, opened 2022-10-03
Homework

alex-anofriev:


src/main/java/core/basesyntax/impl/StorageData.java
```diff
@@ -0,0 +1,45 @@
+package core.basesyntax.impl;
+
+@SuppressWarnings("unchecked")
+public class StorageData<K, S> {
+
```
KostyaNikiforov, line 5: Empty line

src/main/java/core/basesyntax/impl/StorageData.java
```diff
@@ -0,0 +1,45 @@
...
+    private final int initialCapacity = 10;
```
KostyaNikiforov, line 6: It is better to make this field constant

src/main/java/core/basesyntax/impl/StorageData.java
```diff
@@ -0,0 +1,45 @@
+package core.basesyntax.impl;
+
+@SuppressWarnings("unchecked")
+public class StorageData<K, S> {
```
KostyaNikiforov, line 4:
```suggestion
public class StorageData<K, V> {
```
Why S?)
V - is a short for `Value`

KostyaNikiforov, line 4: Seems like this class is redundant. You could add all implementation in the StorageImpl

src/main/java/core/basesyntax/impl/StorageData.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final int initialCapacity = 10;
+    private final Object[] keys = new Object[initialCapacity];
+    private final Object[] values = new Object[initialCapacity];
+    private final int defaultIndex = 0;
```
KostyaNikiforov, line 9: It can be constant too

src/main/java/core/basesyntax/impl/StorageData.java
```diff
@@ -0,0 +1,45 @@
...
+    private int index = 1;
```
KostyaNikiforov, line 10: And the value `1` can be taken from a constant with the name that explains why did you choose exactly this value

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,49 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY = 10;
+    private static final int DEFAULT_INDEX = 0;
+    private static final int NUMBER_ONE = 1;
```
KostyaNikiforov, line 9: Constant names must share the context with the one who reads the code. Name NUMBER_ONE doesn't tell me more than `1`. 
```suggestion
    private static final int INITIAL_INDEX = 1;
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #742, opened 2022-10-04
implemented Storage interface

GrishaSeniv:


ChumakAn (changes requested):
good solution! But do we really need the countIndex variable? It's redundant, we can use size in loops and others for the reason

ChumakAn (approved):
LGTM

PR #743, opened 2022-10-04
Solve the task, it pass all tests and maven clean package

CyberDogFK:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_SIZE = 10;
+
+    private final Container<K, V>[] containers;
```
ChumakAn, line 8: are you sure we need to complicate the storage and use Container? 
I suppose we can use only K[] and V[] arrays

ChumakAn (changes requested):
Interesting solution! 
Please reconsider using the container, or maybe say the reasons we really need this 
Don't you think it complicates all the operations?

src/main/java/core/basesyntax/impl/StorageImpl.java
ChumakAn: Screenshot 2022-10-04 at 17 53 21

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+    private int counter;
+
+    public StorageImpl() {
+        containers = new Container[ARRAY_SIZE];
+        counter = 0;
+
```
ChumakAn, line 14: remove empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_SIZE = 10;
+
+    private final Container<K, V>[] containers;
+    private int counter;
+
+    public StorageImpl() {
+        containers = new Container[ARRAY_SIZE];
+        counter = 0;
```
ChumakAn, line 13: is this a variable for size? let the variables be more self-descriptive

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+
+    }
+
     @Override
     public void put(K key, V value) {
+        if (counter > ARRAY_SIZE) {
+            counter = 0;
```
ChumakAn, line 20: why are you making it null ? instead of throwing an exception on printing some message?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+        if (get(key) != null) {
+            rewriteValueByKey(key, value);
+        } else {
+            containers[counter] = new Container<>(key, value, counter);
+            counter++;
+        }
+    }
+
+    private void rewriteValueByKey(K key, V value) {
+        Container<K, V> lookingContainer = getContainerByKey(key);
```
ChumakAn, line 31: just suggestion, but 
```suggestion
        Container<K, V> existingContainer = getContainerByKey(key);
```

ChumakAn (changes requested):
Good solution, only minor comments, well done! 👏

ChumakAn (approved):
LGTM

PR #745, opened 2022-10-05
solution was added

vyshnevsky:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int NOT_FOUND = -1;
+    private int size;
+    private Pair<K, V> [] items;
+
+    public StorageImpl() {
+        items = new Pair[10];
```
ChumakAn, line 11: 10 can be a constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
     @Override
     public void put(K key, V value) {
+        int index = findIndexByKey(key);
+        if (index != NOT_FOUND) {
+            items[index].setValue(value);
+            return;
+        }
+        items[size] = new Pair<>(key, value);
+        size++;
+    }
```
ChumakAn, line 23: add the check if the storage is not full

src/main/java/core/basesyntax/impl/StorageImpl.java
ChumakAn: like the max size of storage

ChumakAn (changes requested):
Interesting solution! Let's improve that one👍

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        int index = findIndexByKey(key);
+        if (index != NOT_FOUND) {
+            items[index].setValue(value);
+            return;
+        }
+        for (int i = 0; i < items.length; i++) {
+            if (i == items.length - 1) {
+                System.out.println("Array is full");
+            }
+        }
```
ChumakAn, line 27: why do you need this? you can only compare the size to the MAX_ITEMS_NUM

ChumakAn (changes requested):
Only one small comment. Well done

ChumakAn (approved):

PR #747, opened 2022-10-06
Implement Storage solution

anastasiavasylyk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = -1;
+        for (int i = 0; i < elementQuantity; i++) {
+            if (Objects.equals(keys[i], key)) {
+                index = i;
+                values[index] = value;
+                break;
+            }
```
ChumakAn, line 25: you have very similar piece of code here with the lines 36-40, try to avoid code duplication and divide it in method to reuse

ChumakAn (changes requested):
Good! Let's make it better and fix the common mistake

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
```
ChumakAn, line 4: image

ChumakAn (changes requested):
Good! let's check common mistakes

ChumakAn (approved):
Well done!

PR #748, opened 2022-10-06
Implemented StorageImpl class

GHSpeedWagon:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            keys[index] = (K) TRIGGER_FOR_NULL;
+        } else {
+            keys[index] = key;
+        }
+        values[index] = value;
+        index++;
+        size++;
```
KostyaNikiforov, line 27: Seems like you can use only one field and remove another one

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private static final String TRIGGER_FOR_NULL = "null";
+    private final K[] keys;
+    private final V[] values;
```
KostyaNikiforov, line 9: You can make these fields of a type Object  to avoid casting

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private static final String TRIGGER_FOR_NULL = "null";
```
KostyaNikiforov, line 7: Why do you need this constant ?

KostyaNikiforov (changes requested):

GHSpeedWagon:
private static final int MAX_STORAGE_SIZE = 10; we need this for get length for our array of K and V

GHSpeedWagon:
private static final String TRIGGER_FOR_NULL = "null"; if key will be null we will add the string "null" and in future/ we will not have out of sync

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private final K[] keys = (K[]) new Object[MAX_STORAGE_SIZE];
+    private final V[] values = (V[]) new Object[MAX_STORAGE_SIZE];
+    private int index;
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            keys[index] = (K) TRIGGER_FOR_NULL;
```
KostyaNikiforov, line 16: Why can't you just save null here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            key = (K) TRIGGER_FOR_NULL;
+        }
+        for (int i = MAX_STORAGE_SIZE; i > 0; i--) {
+            if (key.equals(keys[i - 1])) {
+                size--;
```
KostyaNikiforov, line 32: You don't need to reduce size while reading the value

KostyaNikiforov: Anyway you can't reuse this place in the array

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: because null is not saving in array/ at least test is not working

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: if i will not redice my size will more then test asks
изображение_2022-10-07_143511712

GHSpeedWagon:
Why are u asking it to me? Need i change my code or know what is it work?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,75 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (canAddWithSameKey(key, value)) {
+            return;
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
```
hronograf, line 26: why do we assign values inside method `canAddWithSameKey` and inside method `put`?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,75 @@
...
+        if (size == ARRAYS_ARE_EMPTY) {
+            return null;
+        }
+        if (key != null) {
+            for (int i = size; i >= 0; i--) {
+                if (key.equals(keys[i - 1])) {
+                    return values[i - 1];
+                }
+            }
+        }
```
hronograf, line 39: maybe we can combine this logic with logic of `findValueWithNullKey` method
it seems to be similar

hronograf (changes requested):
Good!
we can define a method like `indexOf` in the `List` interface and then reuse it
it should help

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: we set values in  canAddWithSameKey method if key is  equal to previous/ and in put method we are assign the values and keys if they are not the same as previous

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,74 @@
...
+                if (key.equals(keys[i])) {
+                    index = i;
+                    break;
+                }
+            }
+            if (keys[i] == key & values[i] != null) {
+                index = i;
+                break;
+            }
+        }
```
hronograf, line 72: - should we start this cycle from `i = size` or `i = size-1`?
- if statements have similar body, can we combine them?
- maybe we don't need `index` variable and can just return the value

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: 1. I am starting looking for from end for pass addTwoElementsWithNullKey() test
 2. - completed 
 3. - completed
 4. ok, but what if we have same key earlier than the previous one? - if we call this method we are always check it
 5. also, it is... - completed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,66 @@
...
+    private V findValue(K key) {
+        int index = size;
+        for (; index >= 0; index--) {
+            if (key != null && key.equals(keys[index])
+                    || keys[index] == key & values[index] != null) {
+                return values[index];
+            }
+        }
+        return values[size];
     }
```
hronograf, line 66: it will cause `IndexOutOfBoundsException` if we have a storage with 10 elements and try to a get value by non-existing key (because size will be `10`, but max possible index will be `9`)
you can check it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,66 @@
...
+        if (size != ARRAYS_ARE_EMPTY) {
+            return key != null && key.equals(keys[size - 1])
+                    || key == null && keys[size - 1] == null;
+        }
+        return false;
+    }
+
+    private V findValue(K key) {
+        int index = size;
+        for (; index >= 0; index--) {
```
hronograf, line 59: do we need `index` variable that initialized out of cycle?
let's write `for (int index = size; ...`
and also we will have `IndexOutOfBoundsException` if we don't write `index = size - 1` here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,66 @@
...
+            }
+            if (keys[size - 1] == null) {
+                values[size] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
```
hronograf, line 34: what if we have same key earlier than the previous one?
you can create the `main` method and check it there

hronograf (changes requested):
Better

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: image
we have size 1 as we need


image
and we have size 5 as we need

did i understand u?

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: k. correcting

src/main/java/core/basesyntax/impl/StorageImpl.java
hronograf: let's do something like
```
int storageMaxSize = 10;
for (int i = 0; i < storageMaxSize; i++) {
  storage.put(i, "element #" + i);
}

System.out.println(storage.get(1));
storage.put(1, "abracadabra");
System.out.println(storage.get(1));
```
please, check if it works properly

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,56 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int CELL_FOR_NULL = 0;
+    private static final int START_INDEX = 1;
```
hronograf, line 9: why do we have separate cell for null? 
what if I have 10 elements and all keys are not null?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,56 @@
...
+        if (key == null && value != null) {
+            values[CELL_FOR_NULL] = value;
+            return;
+        }
+        for (int i = START_INDEX; i < size; i++) {
+            if (key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
hronograf, line 30: here and in the `get` method we have similar logic
let's introduce a private method like `indexOfKey` and use it

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,56 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int START_INDEX = 1;
```
hronograf, line 8: let's do such implementation, in which we will not need `START_INDEX` or similar constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,56 @@
...
+    private int indexOFkey(K key) {
+        if (key != null) {
+            for (int i = START_INDEX; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
+        return 0;
     }
```
hronograf, line 56: we should be able to save a value with `key=null`
so we should be able find it here

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,56 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return values[indexOFkey(key)];
     }
```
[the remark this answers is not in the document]
GHSpeedWagon: we are return value[0] pos

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: if we  put key with null/ we will find it on value[0]/ if we didnt put this keu/ we will return null on value[0] because on value[0] we can have or null with key null or not null with key null

src/main/java/core/basesyntax/impl/StorageImpl.java
GHSpeedWagon: for why?

hronograf: because now we have extra space in our `keys` array for null-key
so we can't add 10 elements with non-null keys, however our `MAX_SIZE` is `10`

hronograf (changes requested):

GHSpeedWagon:
so its last attempt
(i added my lasy solution at 16:59)
it will be unlucky if it if it is not correct, because i ll have first ddlf

src/main/java/core/basesyntax/impl/StorageImpl.java
hronograf: the issue here is that we have some extra space for null-key even if we will not put this null-key to our storage

hronograf (changes requested):

GHSpeedWagon:
i know/ ok

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,82 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    //private static final int START_INDEX = 1;
```
hronograf, line 8: let's remove commented code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,82 @@
...
+    }
+
+    private void addWithNullKey(K key, V value) {
+        if (key == null) {
+            if (keys[0] != null) {
+                K[] keysWithNull = (K[]) new Object[MAX_SIZE];
+                System.arraycopy(keys, 0, keysWithNull, 1, MAX_SIZE - 1);
+                V[] valuesWithNullKey = (V[]) new Object[MAX_SIZE];
+                System.arraycopy(values, 0, valuesWithNullKey, 1, MAX_SIZE - 1);
+                valuesWithNullKey[0] = value;
```
hronograf, line 72: we shouldn't use `System.arraycopy` in this task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,82 @@
...
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
+        if (key == null) {
+            return 0;
+        }
+        return UNEXPECTED_KEY;
+    }
```
hronograf, line 63:
```suggestion
    private int indexOfKey(K key) {
        for (int i = 0; i < size; i++) {
            // TODO try to figure out how we can compare keys if key=null (implementation of Objects.equals may help you)
            if (key.equals(keys[i])) { 
                return i;
            }
        }
        return UNEXPECTED_KEY;
    }
```
after we implement `indexOfKey` correctly, we can remove `addWithNullKey` method and use `indexOfKey` in the `put` method to simplify it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,82 @@
...
+        if (i != UNEXPECTED_KEY) {
+            if (key == null && value != null) {
+                values[i] = value;
+                return;
+            }
+            if (key != null && key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
hronograf, line 32: you will be able to simplify this after you change implementation of `indexOfKey`

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,50 @@
...
     @Override
     public int size() {
-        return -1;
+        return size = keys[0] == null && values[0] != null ? 1 : size;
+    }
+
+    private int indexOfKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                return i;
```
hronograf, line 46: nice!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,50 @@
...
+        if (indexOfKey(key) != UNEXPECTED_KEY) {
+            return values[indexOfKey(key)];
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size = keys[0] == null && values[0] != null ? 1 : size;
```
hronograf, line 40: we should be able to just return size here

hronograf (changes requested):

hronograf (approved):
Finally done! 💥

PR #750, opened 2022-10-06
implemented put, get and size methods

nazariipertsak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+
+    public StorageImpl() {
+        keyStorage = new Object[SIZE];
+        valueStorage = new Object[SIZE];
+        count = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (count == 10) {
```
ChumakAn, line 19: use the size constant, not 10

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+                    || (count != 0 && key == keyStorage[i]) && valueStorage[i] != null) {
+                valueStorage[i] = value;
+                break;
+            } else if (i == keyStorage.length - 1) {
+                keyStorage[count] = key;
+                valueStorage[count] = value;
+                count++;
+                break;
+            }
+        }
```
ChumakAn, line 33: what about checking in this method if any element has such key( in the loop) and if not, only add the new element (out of for-loop)
for now, you are adding the redundant condition block in for-loop

ChumakAn (changes requested):
Good, let's make it better, maybe you have some ideas how to avoid code duplication in put and get method?

nazariipertsak:
> Good, let's make it better, maybe you have some ideas how to avoid code duplication in put and get method?

But I don't understand, what code is duplicated in that methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        if (count == SIZE) {
+            throw new RuntimeException("Can't add new object to storage, because storage is full!");
+        }
+        for (int i = 0; i < keyStorage.length; i++) {
+            if (checkKey(key, i)
+                    || (count != 0 && key == keyStorage[i]) && valueStorage[i] != null) {
+                valueStorage[i] = value;
+                return;
+            }
+        }
```
ChumakAn, line 28: can be separated in the other method to make it more readable

ChumakAn (approved):
LGTM, but take a notice at comment I left

PR #755, opened 2022-10-07
homework has been done

AndriiMatveev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
+    private StorageImpl[] storages = new StorageImpl[10];
```
KostyaNikiforov, line 9: Wow, it looks like your array elements have a lot of redundant functionality

KostyaNikiforov: Let's use two arrays of K and V, or crate a new class for your records

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
...
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (StorageImpl<K, V> str : storages) {
+            if (str != null) {
+                size++;
+            }
+        }
+        return size;
```
KostyaNikiforov, line 59: You can just add a field size and increase it after new element was added

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (StorageImpl<K, V> str : storages) {
+            if (str != null && Objects.equals(str.key, key)) {
+                return str.value;
+            }
+        }
         return null;
```
KostyaNikiforov, line 48: Let's move common logic to the separate private method

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,33 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
+    private ArrayList<K> keys = new ArrayList<>();
+    private ArrayList<V> values = new ArrayList<>();
```
hronograf, line 10: common mistake
image

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,33 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
```
hronograf, line 8: do we need these fields?
looks like we don't use them anywhere

hronograf (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys = new Object[10];
+    private Object[] values = new Object[10];
```
hronograf, line 8: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+        }
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                size++;
+                return;
+            }
+        }
     }
```
hronograf, line 30: image
here and in the `get` method we have similar logic
let's introduce a private method like `indexOfKey` and use it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals((K) keys[i], key)) {
+                values[i] = value;
+                if (size == 0) {
+                    size++;
+                }
```
hronograf, line 18: do we need 
```
if (size == 0) {
    size++;
}
```
also it would be better to have `i < size;` as a condition for loop

hronograf (changes requested):

hronograf (approved):
Nice!

PR #756, opened 2022-10-07
Created StorageImpl class.

PavloPolovyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_STORAGE_SIZE = 15;
+    private static final int MINIMUM_STORAGE_SIZE = 5;
```
ChumakAn, line 7: do you really need this?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+        }
+        this.storageSize = storageSize;
+        currentSize = 0;
+        this.entries = new Object[storageSize];
+        isGrowable = false;
+    }
+
+    static class Entry<K, V> {
+        private final K key;
+        private final Integer hashCode;
```
ChumakAn, line 41: you don't really need to use the hashCode here

ChumakAn (changes requested):
Good, but let's remember that we should read properly the description of task. Minimum size is not 5. Don't you think you're ivercomplicating the logic with hash code calculations? If not, please explain the reason

src/main/java/core/basesyntax/impl/StorageImpl.java
PavloPolovyi: I decided to add minimum requirement for storage size, to minimize index collisions.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+        if (storageSize <= MINIMUM_STORAGE_SIZE) {
+            throw new IllegalArgumentException("Can't create "
+                    + "Storage with size less than or equal "
+                    + MINIMUM_STORAGE_SIZE);
+        }
+        this.storageSize = storageSize;
+        currentSize = 0;
+        this.entries = new Object[storageSize];
+        isGrowable = false;
+    }
```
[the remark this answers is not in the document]
PavloPolovyi: I decided to add minimum requirement for storage size, to minimize index collisions.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+        private final K key;
+        private final Integer hashCode;
+        private V value;
+        private Entry<K, V> next;
+
+        Entry(K key, V value, Integer hashCode) {
+            this.key = key;
+            this.value = value;
+            this.hashCode = hashCode;
+            next = null;
```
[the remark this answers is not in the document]
PavloPolovyi: I store next value for entries, for which calculated index is the same.  Like in one bucket.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+            this.next = next;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
     @Override
     public void put(K key, V value) {
+        Entry<K, V> entry;
+        checkCapacity();
+        int hashCode = getHashCode(key);
```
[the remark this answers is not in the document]
PavloPolovyi: why? i check if there is enough space, and then calculate hashCode for key. This check is fast enough, there are no loops, just simple if branching and statements. Hashcode i use to calculate index of entry array, where i put my Entry<K, V>.

src/main/java/core/basesyntax/impl/StorageImpl.java
PavloPolovyi: i use hashCode field here, because i don't want to call private method getHashCode() every time i need it.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+            entries[index] = entry;
+            currentSize++;
+        }
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public V get(K key) {
+        Entry<K, V> entry;
+        Integer hashCode = getHashCode(key);
```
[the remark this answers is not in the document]
PavloPolovyi: why? i use this hashCode to calculate position of entry in entry array : int index = Math.abs(hashCode % storageSize)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,161 @@
...
+
+    public StorageImpl() {
+        storageSize = DEFAULT_STORAGE_SIZE;
+        currentSize = 0;
+        entries = new Object[DEFAULT_STORAGE_SIZE];
+        isGrowable = false;
+    }
+
+    public StorageImpl(boolean isGrowable) {
+        this();
```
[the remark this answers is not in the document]
PavloPolovyi: why twice? if i don't place this line (this()) here, there would be an array of entries with initial size equals zero.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,160 @@
...
+    private Object[] entries;
+
+    public StorageImpl() {
+        storageSize = DEFAULT_STORAGE_SIZE;
+        currentSize = 0;
+        entries = new Object[DEFAULT_STORAGE_SIZE];
+        isGrowable = false;
+    }
+
+    public StorageImpl(boolean isGrowable) {
```
liliia-ponomarenko, line 21: do you use this constructor?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,160 @@
...
+        this();
+        this.isGrowable = isGrowable;
+    }
+
+    public StorageImpl(int storageSize) {
```
liliia-ponomarenko, line 26: the same here. Let’s  remove redundant  code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,160 @@
...
+        this.storageSize = storageSize;
+        currentSize = 0;
+        this.entries = new Object[storageSize];
+        isGrowable = false;
+    }
+
+    static class Entry<K, V> {
```
liliia-ponomarenko, line 33: This class can be private (the same for getters and setters). Also, usually nested  private classes are placed at the end of class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,160 @@
...
+        int hashCode = getHashCode(key);
+        int index = Math.abs(hashCode % storageSize);
+        if (entries[index] != null) {
+            Entry<K, V> existed = (Entry<K, V>) entries[index];
+            if (existed.getHashCode().equals(hashCode)) {
+                existed.setValue(value);
+            } else {
+                entry = new Entry<>(key, value, hashCode);
+                existed.setNext(entry);
+                currentSize++;
```
liliia-ponomarenko, line 85: You can’t have more than 2 objects in the bucket. you need to iterate over the singly linked list as you do in the get method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,160 @@
...
 
+    @SuppressWarnings("unchecked")
     @Override
     public V get(K key) {
+        Entry<K, V> entry;
+        Integer hashCode = getHashCode(key);
+        int index = Math.abs(hashCode % storageSize);
+        if (entries[index] != null) {
+            entry = (Entry<K, V>) entries[index];
+            if (entry.getHashCode().equals(hashCode)) {
```
liliia-ponomarenko, line 102: does this guarantee that the keys are the same? whether two objects can have the same hashcode

liliia-ponomarenko (changes requested):
Let's improve your solution 😉

liliia-ponomarenko (approved):
good job 🏆

PR #757, opened 2022-10-07
My storage program

anton-pushkar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        keyArray = (K[])new Object[MAX_SIZE];
+        valueArray = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int keyIndex = getKeyIndex(key);
+        if (keyIndex != WRONG_KEY) {
+            valueArray[keyIndex] = value;
+            keyArray[keyIndex] = key;
```
ChumakAn, line 22: you don't need to change the key, it should be the same

ChumakAn (changes requested):
Well done! Only one minor comment

liliia-ponomarenko (approved):
Great job 👏

PR #758, opened 2022-10-07
realize StorageImpl class, implemented put, get, size, indexOf methods

yehbreus:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     @Override
     public void put(K key, V value) {
+        if (this.size == START_SIZE || get(key) == null) {
+            this.key[size] = key;
+            this.value[size] = value;
+            size++;
+        } else {
+            this.value[this.indexOf(get(key))] = value;
+        }
     }
```
hronograf, line 27: - `if (this.size == START_SIZE || get(key) == null) {` - do we need `this.size == START_SIZE` here?
- looks like we go through `keys` inside `get` method (20 and 25 line) and after that through `values` inside `indexOf` method, 
it would be better to get index of the key from `indexOf` method and then use it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < this.size + 1; i++) {
+            if (key == null && this.key[i] == null) {
+                return this.value[i];
+            }
+            if (this.key[i] != null && this.key[i].equals(key)) {
+                return this.value[i];
+            }
+        }
```
hronograf, line 38: after we change `int indexOf(V value)` to `int indexOf(K key)` this logic will be similar to `indexOf` method, so let's reuse it here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
-        return -1;
+        return this.size;
+    }
+
+    private int indexOf(V value) {
+        int index = -1;
+        for (int i = 0; i < this.size; i++) {
+            if (this.value[i].equals(value)) {
+                index = i;
+            }
```
hronograf, line 52: we can just return `i` here, because we don't need to go through full array if we found an element at first place
also, `index` variable may be redundant

hronograf (changes requested):

hronograf (approved):
Good! 🏎️

PR #759, opened 2022-10-07
Implemented StorageImpl class

NSay0:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size;
+    private int lastElement;
```
hronograf, line 8: do we need field `lastElement`?
I think it always has value `size-1`, so we don't need duplication of information here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
...
+            }
+            return false;
+        }
+        boolean finded = false;
+        for (K existingKey : keys) {
+            if (key.equals(existingKey)) {
+                finded = true;
+                break;
+            }
+        }
```
hronograf, line 42: I think we have similar logic here
we go through array and check it's values
let's combine this logic

also we have similar logic in the `getIndex` method, so maybe we can use it here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
...
+
+    private int getIndex(K key) {
+        if (isKeyExist(key)) {
+            int index = 0;
+            for (int i = 0; i <= lastElement; i++) {
+                if (key == null && keys[i] == null) {
+                    break;
+                } else if (keys[i] != null && keys[i].equals(key)) {
+                    break;
+                }
```
hronograf, line 54: we have two if statements with the same body, let's combine them
also, consider `key == keys[i]` comparison (instead of `key == null && keys[i] == null`)
and we can just write `return i;`, so we don't need `index` variable

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/impl/StorageImpl.java
NSay0: Sure, removed in new commit.
+removed in new commit

src/main/java/core/basesyntax/impl/StorageImpl.java
NSay0: Simplified and combined with indexOf method(Reduntant isExist method deleted).
+simplified in new commit

src/main/java/core/basesyntax/impl/StorageImpl.java
NSay0: Simplified and renamed to indexOf.
+simplified in new commit

hronograf (approved):
Nice! 🔥 
also @NSay0 read about [Ordering of class contents](https://mate-academy.github.io/style-guides/java/java.html#s3.4.2-ordering-class-contents)

PR #762, opened 2022-10-07
implemented business logic to storage program

muzykaVP:


liliia-ponomarenko (changes requested):
Let’s read common mistakes again

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private int size = 0;
+    private Object[] valuesList = new Object[10];
+    private Object[] keysList = new Object[10];
```
hronograf, line 8: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keysList.length; i++) {
+            if (keysList[i] == null && valuesList[i] != null && keysList[i] == key
```
hronograf, line 13: do we need `keysList[i] == null && valuesList[i] != null` here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+                    || keysList[i] != null && keysList[i].equals(key)) {
+                valuesList[i] = value;
+                break;
+            }
+            if (keysList[i] == null && valuesList[i] == null) {
```
hronograf, line 18: can we just use `else` here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keysList.length; i++) {
+            if (keysList[i] == key && valuesList[i] != null
+                    || keysList[i] != null && keysList[i].equals(key)) {
+                return (V) valuesList[i];
+            }
+        }
```
hronograf, line 34: we have similar logic here and in the `put` method
let's introduce a method `indexOfKey` or something similar

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    }
+
+    private Integer indexOfKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keysList[i] == key || keysList[i] != null && keysList[i].equals(key)) {
+                return i;
+            }
+        }
+        return size;
+    }
```
hronograf, line 24: it would be better to return `-1` or `null` here, for example, and handle this situation

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+            valuesList[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return (V) valuesList[indexOfKey(key)];
     }
```
hronograf, line 39: what if we add 10 elements and after that we will call `get` method?
you can check it in your `main` method
it is related to the previous comment

hronograf (changes requested):

hronograf (approved):
Nice!

PR #768, opened 2022-10-08
realised key-value methods in StorageImpl class

apc10bbe:


liliia-ponomarenko (changes requested):
Read common mistakes again

liliia-ponomarenko (changes requested):
Read common mistakes again

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        }
+        return key == null && key == keys[i];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean putOrNot = false;
+        for (int i = 0; i < keys.length; i++) {
+            if (checkValue(key, i)) {
+                values[i] = value;
```
liliia-ponomarenko, line 30: you have a similar code on lines 45-47. Can you move this logic to a separate private method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+                putOrNot = true;
+            }
+        }
+        if (!putOrNot) {
+            keys[count] = key;
+            values[count] = value;
+            count++;
+        } else {
+            count = NULL_SIZE;
```
liliia-ponomarenko, line 39: It’s hardcode. It works only for our test cases if we increase the array in tests, your solution won't pass them.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+    public boolean checkValue(K key, int i) {
+        if (key != null && key.equals(keys[i])) {
+            return true;
+        }
+        return key == null && key == keys[i];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean putOrNot = false;
```
liliia-ponomarenko, line 27: you can avoid using this flag, but if you want to leave this flag give a more suitable name ([t.10](https://www.comp.nus.edu.sg/~cs2103/AY1617S1/contents/coding-standards-java.html#specific-naming-conventions))

liliia-ponomarenko (changes requested):
Let’s  fix your solution 😉

src/main/java/core/basesyntax/impl/StorageImpl.java
apc10bbe: this code is change values[i], but at 45-47 lines that code is return values[i]. Can i leave as is?

apc10bbe: ok, i move this logic to separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return count;
+    }
+
+    private boolean checkValue(K key, int i) {
+        return ((key == keys[i]) || (key != null && key.equals(keys[i])));
```
liliia-ponomarenko, line 41:
```suggestion
        return (key == keys[i] || key != null && key.equals(keys[i]));
```
Redundant  brackets

liliia-ponomarenko (approved):
Great 😊

PR #770, opened 2022-10-09
hw-storage

Iryna-a:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private static final int START_INDEX = 0;
+    private int currentIndex;
```
hronograf, line 8: maybe `size` would be a better name for this field

hronograf (approved):
On the first try! 🚀

PR #772, opened 2022-10-09
storege HW solution

B1b1z9na:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
...
+    private int currantStorageSize = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currantStorageSize; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                return;
+            }
+        }
```
hronograf, line 18: we have similar logic inside `get` method
let's introduce a method that should return an index of key and use it
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_MAX_SIZE = 10;
+    private Object[] keys = new Object[STORAGE_MAX_SIZE];
+    private Object[] values = new Object[STORAGE_MAX_SIZE];
+    private int currantStorageSize = 0;
```
hronograf, line 9: image

hronograf, line 9:
```suggestion
    private Object[] keys = new Object[STORAGE_MAX_SIZE];
    private Object[] values = new Object[STORAGE_MAX_SIZE];
    private int currentStorageSize = 0;
```

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return currentStorageSize;
+    }
+
+    public int findIndexByKey(K key) {
```
hronograf, line 43:
```suggestion
    private int findIndexByKey(K key) {
```

hronograf (approved):
Nice!

PR #776, opened 2022-10-10
implemented the Storage of key-value type

ana-rom:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        }
+        if (arrayIndex > 0 && (key == arrayK[arrayIndex - 1]
+                || key != null && key.equals(arrayK[arrayIndex - 1]))) {
+            arrayK[arrayIndex - 1] = key;
+            arrayV[arrayIndex - 1] = value;
+        } else {
+            arrayK[arrayIndex] = key;
+            arrayV[arrayIndex] = value;
+            arrayIndex++;
+        }
```
liliia-ponomarenko, line 29: this solution is valid only when storage has two or fewer objects. It works for our test cases but it is  the wrong solution

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     public void put(K key, V value) {
+        getIndex(key);
+        arrayK[index] = key;
+        arrayV[index] = value;
+        index++;
+        size++;
+
+    }
+
+    public int getIndex(K key) {
```
liliia-ponomarenko, line 27: this method can be private and move after public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        for (int i = 0; i < arrayK.length; i++) {
+            if (key == arrayK[i] || key != null && key.equals(arrayK[i])) {
+                index = i;
+                return index;
+            }
+        }
+        return 0;
```
liliia-ponomarenko, line 34: ```0``` it’s an available value for the array. Better use ```-1```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private int size;
+
+    public StorageImpl() {
+        arrayK = new Object[MAX_ELEMENTS];
+        arrayV = new Object[MAX_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        getIndex(key);
```
liliia-ponomarenko, line 19: you need to create a local variable for this index value and add a conditional. If ```index == -1``` add new item, if index was find just update value by index

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS = 10;
+    private final Object[] arrayK;
+    private final Object[] arrayV;
+    private int index;
```
liliia-ponomarenko, line 9: better make index local variable

liliia-ponomarenko (changes requested):
Let’s  fix your solution 😉

liliia-ponomarenko (approved):
great job! 🏆

PR #782, opened 2022-10-10
Storage implementation added. I have take into account the "common-mi…

luka-lozovyk:
…stakes-file". I find it more complicated to use arrays, so I would like to use lists, if it's OK.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,49 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
+import java.util.List;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
+    private List<K> keys = new ArrayList();
+    private List<V> values = new ArrayList<>();
+
     @Override
     public void put(K key, V value) {
+        if (!keys.contains(key)) {
+            this.key = key;
+            this.value = value;
+            keys.add(key);
+            values.add(value);
+        } else {
+            int indexOfValueToBeRewritten = values.indexOf(keys.indexOf(key));
+            values.add(indexOfValueToBeRewritten + 1, value);
+        }
     }
 
     @Override
-    public V get(K key) {
-        return null;
+    public String toString() {
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append(keys.get(0) + ", " + values.get(0));
+        for (int i = 1; i < keys.size(); i++) {
+            stringBuilder.append("\n").append(keys.get(i) + ", " + values.get(i));
+        }
+        return stringBuilder.toString();
+    }
+
+    @Override
+    public V get(Object key) {
+        if (!keys.contains(key)) {
+            return null;
+        } else {
+            return values.get(keys.indexOf(key));
+        }
     }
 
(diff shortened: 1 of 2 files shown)


hronograf (changes requested):
Please, read common mistakes more carefully

luka-lozovyk:
> Please, read common mistakes more carefully

I have read common mistakes line by line.  Your remark is about using of lists, or there are any other mistakes?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K key;
+    private V value;
```
hronograf, line 8: do we need this?
```
private K key;
private V value;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+    private int count = 0;
```
hronograf, line 9: let's move initialization to constructor (`count = 0;`)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+            values[count] = value;
+            count = count + 1;
+        } else {
+            values[indexOfKey(keys, key)] = value;
+        }
+    }
+
+    private int indexOfKey(K[] keys, K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == key || key != null && key.equals(keys[i])) {
```
hronograf, line 31: we can access `keys`, because it is a field of our class, so let's change method signature to 
`private int indexOfKey(K key)`
also, we can rewrite loop condition from `i < keys.length;` to `i < count;`, so we will not need to go through full array

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+    private V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[STORAGE_SIZE];
+        this.values = (V[]) new Object[STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (indexOfKey(keys, key) == -1 || key == null && get(key) != null) {
```
hronograf, line 20: after you fix previous comment, you will be able to remove this part:
```suggestion
        if (indexOfKey(keys, key) == -1) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
-        return null;
+    public V get(Object key) {
+        int indexByKey = 0;
+        for (int i = 0; i < count; i++) {
+            if (keys[i] == key || key != null && key.equals(keys[i])) {
+                indexByKey = i;
+            }
+        }
+        return values[indexByKey];
     }
```
hronograf, line 47: let's try to use `indexOfKey` method here

hronograf (changes requested):
read also about [Ordering of class contents](https://mate-academy.github.io/style-guides/java/java.html#s3.4.2-ordering-class-contents)

PR #788, opened 2022-10-11
Implement all methods

DeonOni:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+        }
+        keys[size()] = key;
+        values[size()] = value;
+        countAmountOfObjectsAdded++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i <= size(); i++) {
+            if (keys[i] == key || key != null && key.equals(keys[i])) {
```
anastasiavasylyk, line 33: repeated code (DRU - don`t repeat yourself)  line - 33, 20 . Better to do this check in a separate method.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int countAmountOfObjectsAdded;
```
anastasiavasylyk, line 9: you can improve the name of field. For example countOfAddedObjects or amountOfAddedObjects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
```
anastasiavasylyk, line 13: it`s will be better if you will use this. when you initialize your field.

src/main/java/core/basesyntax/impl/StorageImpl.java
DeonOni: This 2 if statements are not the same. The one at line 20 does return nothing, return statement there for ending methods job, but the one at line 33 return the value of given index.

anastasiavasylyk: I don`t tell you about body of this two if. you can make method private boolean compareKey(K key) for example, and this method will return you true or false as a result of your compare (keys[i] == key || key != null && key.equals(keys[i])) in your if statement

anastasiavasylyk (changes requested):

anastasiavasylyk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size(); i++) {
+            if (keyCompare(key, keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
liliia-ponomarenko, line 24: You can improve it. Create a private method that returns the index of the item in the array or -1 if the item is absent in the array. You can replace code duplicates in methods ```put()``` and ```get()```

liliia-ponomarenko (changes requested):
Good solution 😉 Let’s improve it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[STORAGE_SIZE];
+        this.values = (V[]) new Object[STORAGE_SIZE];
+        this.amountOfAddedObjects = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (keyIndex(key) > -1 && keyCompare(key, keys[keyIndex(key)])) {
```
liliia-ponomarenko, line 19:
```suggestion
        if (keyIndex(key) > -1) {
```
I guess this conditional is redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
     @Override
     public int size() {
+        return amountOfAddedObjects;
+    }
+
+    public boolean keyCompare(K key, K keyToCompare) {
+        return keyToCompare == key || key != null && key.equals(keyToCompare);
+    }
+
+    public int keyIndex(K key) {
```
liliia-ponomarenko, line 45: these methods should be private

liliia-ponomarenko (approved):
good job! 😉

PR #789, opened 2022-10-12
HW-solution

Kulibchenko:


aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private int index(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
aaananas, line 39:
```suggestion
            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
```

aaananas (approved):
remember about a new branch for your solution

PR #791, opened 2022-10-13
Jv storage hw

TymurProkhorov:


liliia-ponomarenko (changes requested):
Not modify test class. Fix problems with  JavaCI checks.

liliia-ponomarenko (approved):
LGTM! 🔥

PR #797, opened 2022-10-17
Solution was added.

KatyaHumeniuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     @Override
     public int size() {
-        return -1;
+        int storageSize = 0;
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] != null || values[i] != null) {
+                storageSize++;
+            }
+        }
+        return storageSize;
```
zachar-chorny, line 43: You must have a size class variable, which is what you will return here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
+                return values[i];
+            } else if (key == null && keys[i] == null) {
+                return values[i];
+            }
```
zachar-chorny, line 30: Why two different ifs with the same body?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    private static final int COUNT_OF_ARRAY = 10;
+    private K [] keys = (K[]) new Object[COUNT_OF_ARRAY];
+    private V [] values = (V[]) new Object[COUNT_OF_ARRAY];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null && values[i] == null)
+                    || (keys[i] == null && key == null && values[i] != null)
+                    || (keys[i] != null && keys[i].equals(key))) {
```
zachar-chorny, line 15: With the addition of the size variable, this check can be simplified

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int COUNT_OF_ARRAY = 10;
+    private K [] keys = (K[]) new Object[COUNT_OF_ARRAY];
+    private V [] values = (V[]) new Object[COUNT_OF_ARRAY];
```
aaananas, line 8: lets initialize arrays in costructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+    private int storageSize = 0;
```
```suggestion
    private int storageSize;
```
int value is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+                    || (keys[i] != null && keys[i].equals(key))) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
+            keys[storageSize] = key;
+            values[storageSize] = value;
+            storageSize++;
+            break;
+        }
```
aaananas, line 24: when you are iterating from 0 to size you know that those cells are taken
you only check if the key is already present and rewrite your value
so in for loop you can call `return;` and finish the method

storageSize - stores the size and also it is the next empty cell in your storage
after for-loop just set key and value on the storageSize index

aaananas (changes requested):

zachar-chorny (approved):

PR #803, opened 2022-10-21
Solved Storage

o-shyshkan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final byte MAX_ITEMS_NUMBER = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        size = 0;
```
```suggestion
```
int value is 0 by default

aaananas (approved):

PR #806, opened 2022-10-26
26/10 commit

Watxesnavo:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_CAPACITY = 10;
+    private static int SIZE;
```
kateryna-deshchinska, line 7: why this field static? it related to class, or to object?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[START_CAPACITY];
+        values = (V[]) new Object[START_CAPACITY];
+        SIZE = 0;
```
kateryna-deshchinska, line 14:
```suggestion
```

kateryna-deshchinska (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int startCapacity = 10;
```
Rommelua, line 6: image
This comment was about SIZE. You should be able to create many StorageImpl instances and each should have its own 'size'. That is the reason why it can not be static.
startCapacity - completely another variable. We do not change it like 'size' - it is constant. It must be static.

Rommelua (changes requested):
Good work, just rollback change with **startCapacity**

zachar-chorny (approved):
Well done

PR #809, opened 2022-10-27
Implemented the Storage of key-value type that is parameterized with two types of data

maks-untilov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int DEFAULT_CAPACITY = 10;
+    private static int size;
```
aaananas, line 9: size should not be static
because if you create a few instances of this class this field will have the same value for 2 different storages

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_CAPACITY];
+        values = (V[]) new Object[DEFAULT_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && values[i] == null || Objects.equals(keys[i], key)) {
```
aaananas, line 21: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V value = null;
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(keys[i], key)) {
```
aaananas, line 33: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
     @Override
     public int size() {
-        return -1;
+        size = 0;
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] != null) {
+                size++;
+            }
+        }
+        return size;
```
aaananas, line 49:
```suggestion
        return size;
```
lets change size if you add new values in put method

aaananas (changes requested):

aaananas (approved):

PR #812, opened 2022-10-29
solution task about generetic

VictoriaKh-T:
I think I need help with this

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,61 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_ARRAY = 10;
+    private int count = 0;
+    private K key;
+    private V value;
+    private StorageImpl[] mas = new StorageImpl[MAX_ARRAY];
+    private StorageImpl[] arrayOfElements;
+    
     @Override
     public void put(K key, V value) {
+        StorageImpl<K, V> storage = new StorageImpl<>();
+        storage.key = key;
+        storage.value = value;
+        if (!isIn(key)) {
+            mas[count] = storage;
+            count++;
+        }
     }
-
+    
     @Override
     public V get(K key) {
-        return null;
+        
+        for (int i = 0; i < mas.length; i++) {
+            if (mas[i] == null) {
+                continue;
+            }
+            if (key != null) {
+                if (key.equals(mas[i].key)) {
+                    value = (V) mas[i].value;
+                }
+            } else {
+                if (key == (mas[i].key)) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_ARRAY = 10;
```
zachar-chorny, line 6:
```suggestion
    private static final int ARRAY_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+    private int count = 0;
```
zachar-chorny, line 7:
```suggestion
    private int size;
```
int is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+    private K key;
+    private V value;
+    private StorageImpl[] mas = new StorageImpl[MAX_ARRAY];
```
zachar-chorny, line 10: If you want to do this via a class then create an internal static class with Key & Value fields instead of using StorageImpl

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+            }
+            if (key != null) {
+                if (key.equals(mas[i].key)) {
+                    mas[i].value = value;
+                }
+            } else {
+                if (key == (mas[i].key)) {
+                    mas[i].value = value;
+                }
+            }
```
zachar-chorny, line 83: You've got code repeating here, wouldn't it be better to immediately replace the match you've found?

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
VictoriaKh-T: I made all over again. After learning arrayList

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        this.keys = (K[]) new Object[ARRAY_SIZE];
+        this.values = (V[]) new Object[ARRAY_SIZE];
+    }
+    
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                size--;
```
aaananas, line 21:
```suggestion
                return;
```

aaananas (approved):

PR #813, opened 2022-10-30
implementation the storage

BusyhinD:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int CAPACITY_OF_STORAGE = 10;
+    private StorageImpl<K, V>[] storages;
+    private K key;
+    private V value;
```
zachar-chorny, line 9: If you want to do this via a class, create an intrinsic class with key & value fields

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private int countOfElements;
```
zachar-chorny, line 10:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+
+    public StorageImpl() {
+        storages = new StorageImpl[CAPACITY_OF_STORAGE];
+        countOfElements = 0;
```
zachar-chorny, line 14:
```suggestion
```
int is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean isSameKey = false;
```
zachar-chorny, line 19: Use return; instead of this variable in the loop

zachar-chorny (changes requested):

Rommelua (approved):

PR #815, opened 2022-10-31
edit StorageImpl

YuriiTrokhymenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER = 10;
+    private Object[] keys;
+    private Object[] values;
+    private K key;
+    private V value;
```
zachar-chorny, line 11: You don't need these two fields

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
+
```
zachar-chorny, line 12: add size field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
+        }
+        return result;
+    }
+
     @Override
     public void put(K key, V value) {
+        int volume = size();
+        if (volume == 0) {
+            keys = new Object[MAX_NUMBER];
+            values = new Object[MAX_NUMBER];
```
zachar-chorny, line 37: Arrays should be created in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
+        }
+        int index = getIndex(key);
+        if (index < 0) {
+            if (volume <= MAX_NUMBER) {
+                keys[volume] = key;
+                values[volume] = value;
+            }
+        } else {
+            values[index] = value;
+        }
```
zachar-chorny, line 47: You have made things very complicated, you just need to go through the keys array and check if there is already such a key, if there is, replace the value in the values array by its index, and if not, add new values to the keys and values array. Use size variable for index of added objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
+        if (values == null) {
+            return 0;
+        }
+        int result = 0;
+        for (int i = 0; i < MAX_NUMBER; i++) {
+            if (values[i] != null) {
+                result = result + 1;
+            }
+        }
+        return result;
```
zachar-chorny, line 70:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        if (index < 0) {
+            return null;
+        }
+        return (V) values[index];
```
zachar-chorny, line 56: Here iterate through the key array and return the value under the index of the matching key, if there is no such key return null

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,71 @@
...
+            }
+        } else {
+            for (int i = 0; i < volume; i++) {
+                if (key.equals(keys[i])) {
+                    result = i;
+                }
+            }
+        }
+        return result;
+    }
```
```suggestion
```

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,65 @@
...
+
+    public boolean equalsKey(K key, Object o) {
+        if (key == null) {
+            return (o == null);
+        }
+        if (o == null) {
+            return false;
+        }
+        return o.equals(key);
+    }
```
zachar-chorny, line 25:
```suggestion
    public boolean equalsKey(K firstKey, K secondKey) {
        if (firstKey == secondKey || firstKey != null && firstKey.equals(secondKey)){
                return true;
        }
        return false;
    }
```

zachar-chorny (changes requested):

PR #816, opened 2022-10-31
Jv storage solution

DenisKalyakin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final byte MAX_ITEMS_NUMBER = 10;
+    private final K[] arrayKeys;
+    private final V[] arrayValues;
+    private byte index = 0;
+    private boolean nullMarker = false;
```
zachar-chorny, line 10:
```suggestion
    private int size;
```
You don't need this variable, let me explain how to simplify your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < arrayKeys.length; i++) {
+            if (key == null && arrayKeys[i] == null) {
+                return arrayValues[i];
+            }
+            if (arrayKeys[i] != null && arrayKeys[i].equals(key)) {
+                return arrayValues[i];
+            }
```
zachar-chorny, line 52: This ifs can be combined

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final byte MAX_ITEMS_NUMBER = 10;
+    private final K[] arrayKeys;
+    private final V[] arrayValues;
```
zachar-chorny, line 8:
```suggestion
    private final K[] keys;
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (index == MAX_ITEMS_NUMBER) {
+            index = 0;
+        }
+        for (int i = 0; i < arrayKeys.length; i++) {
+            if (arrayKeys[i] != null && arrayKeys[i].equals(key)) {
+                arrayValues[i] = value;
+                index--;
```
zachar-chorny, line 25:
```suggestion
            if (arrayKeys[i] == key || arrayKeys[i] != null && arrayKeys[i].equals(key)) {
                arrayValues[i] = value;
                return;
```
If written this way, the code below can be greatly simplified to just 3 lines without any additional checks

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+    public StorageImpl() {
+        arrayValues = (V[]) new Object[MAX_ITEMS_NUMBER];
+        arrayKeys = (K[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (index == MAX_ITEMS_NUMBER) {
+            index = 0;
+        }
```
zachar-chorny, line 21:
```suggestion
```
We don't need this

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
DenisKalyakin: Thank you, it means a lot to me

DenisKalyakin:
Thanks, the only thing I didn't understand was how to do without nullMarker.
I need to somehow know when a null is already in the array keys

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+            }
+            values[size] = value;
+            nullMarker = true;
+            size++;
+        }
+        if (key != null) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
aaananas, line 42:
```suggestion
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```
it can be simplified

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
DenisKalyakin: my jaw dropped) Thank you so much!

Rommelua (approved):

PR #826, opened 2022-11-09
home work

Volodymyr-Mykychak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private K[] keys = (K[]) new Object[DEFAULT_CAPACITY];
+    private V[] values = (V[]) new Object[DEFAULT_CAPACITY];
```
aaananas, line 8: initialize in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```
int value is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
+
     @Override
-    public void put(K key, V value) {
+    public void put(K key, V value) { // key = null value = fjk
```
aaananas, line 12:
```suggestion
    public void put(K key, V value) { 
```

aaananas (changes requested):

aaananas (approved):

PR #827, opened 2022-11-10
new method (isEquals) added

Kateryna-Takhtai:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_ELEMENTS = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```
int variable is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        V value = null;
+        for (int i = 0; i < keys.length; i++) {
```
aaananas, line 33:
```suggestion
        for (int i = 0; i < size; i++) {
```

aaananas (approved):

PR #828, opened 2022-11-10
1

dimonkoval:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        boolean rewritten = false;
+        for (int i = 0; i < size; i++) {
+            if (checkOnEquals(key, i)) {
+                arrayValue[i] = value;
+                rewritten = true;
+                break;
```
kateryna-deshchinska, line 18: you can put return instead, and then you don't need rewritten value)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean checkOnEquals(K key, int i) {
```
kateryna-deshchinska, line 43:
```suggestion
    private boolean checkOnEquals(K key, int i) {
```

kateryna-deshchinska (changes requested):

kozhukhovsky (approved):

PR #836, opened 2022-11-21
first working version

Sanderfuck:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+
+    private int size;
+    private K[] keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+    private V[] values = (V[]) new Object[MAX_ARRAY_LENGTH];
```
aleksandr-bevza, line 10: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        if (checkedIndex >= 0) {
+            values[checkedIndex] = value;
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+    }
+
+    private int checkIndex(K key) {
```
aleksandr-bevza, line 24: If method's name starts with "check" - it is expected that it will returns boolean value. It's more logical to use "get" here
```suggestion
    private int getIndex(K key) {
```

aleksandr-bevza (changes requested):

aaananas (approved):

PR #837, opened 2022-11-26
StorageImpl class was implemented

katrevichdv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+        private Cell(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+
+    }
+
+    public StorageImpl() {
+        this.storage = new Cell[MAX_SIZE];
+        this.count = 0;
```
aleksandr-bevza, line 24: Class field with type "int" is 0 by default. No need to set it with 0 again
```suggestion
```

aleksandr-bevza (approved):

PR #838, opened 2022-11-26
added solution for jv-storage task

vgladkyi:


src/main/java/core/basesyntax/Storage.java
```diff
@@ -6,4 +6,5 @@
     V get(K key);
 
     int size();
+
```
aleksandr-bevza, line 9: Do we need this empty line to be added?)

aleksandr-bevza (approved):

PR #839, opened 2022-11-27
added resizable array basic functions

ItsUnbidden:


src/main/java/core/basesyntax/App.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import core.basesyntax.impl.StorageImpl;
+
+public class App {
```
Rommelua, line 5: We do not need this class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys;
+    private Object[] values;
+
+    public StorageImpl() {
+        keys = new Object[]{};
+        values = new Object[]{};
```
Rommelua, line 11: Task: If you use arrays for your solution, we assume that the maximum number of elements in our Storage is 10.
Create a 10-element array. We do not need an increment method. Keep it simple.

Rommelua (changes requested):

PR #840, opened 2022-11-27
fixed size version

ItsUnbidden:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
 
     @Override
     public int size() {
+        int arraySize = 0;
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (values[i] != null) {
+                arraySize++;
+            }
+        }
+        return arraySize;
```
Rommelua, line 43: Create a field size and just return it here. You do not need to calculate it each time.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    }
+
+    private int getKeyIndex(Object key) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if ((this.keys[i] == null && key == null) || (this.keys[i] != null
```
Rommelua, line 48:
```suggestion
            if (this.keys[i] ==  key || (this.keys[i] != null 
```

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    }
+
+    private int getKeyIndex(Object key) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (((this.keys[i] == key) || (this.keys[i] != null 
+                    && this.keys[i].equals(key))) && values[i] != null) {
+                return i;
+            }
+        }
         return -1;
```
```suggestion
    private int getKeyIndex(Object key) {
        for (int i = 0; i < size; i++) {
            if (this.keys[i] == key || this.keys[i] != null 
                    && this.keys[i].equals(key)) {
                return i;
            }
        }
        return -1;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index == -1) {    
+            size++;
+            int lastElement = size - 1;
+            keys[lastElement] = key;
+            values[lastElement] = value;
+        } else {
+            values[index] = value;
+        }
```
```suggestion
        if (index == -1) { 
            keys[size] = key;
            values[size] = value;
            size++;
        } else {
            values[index] = value;
        }   
```

aaananas (changes requested):

PR #841, opened 2022-11-28
constant was added

metelska-anastasiia:
2 arrays of parametrised datas were created
methods put(), get(), size() were overrided

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,41 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER = 10;
+    private final K[] keys;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER = 10;
```
Rommelua, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
Rommelua, line 9:
```suggestion
    private int size;
```

Rommelua (approved):
Approмув but have few comments.

PR #842, opened 2022-12-04
Created:

nkharyshyn:
- 2d array to storage all keys and values,
 - keyAlreadyExist method to check if value with this key already exist
 - variables counter (to count all values) and replacedElements (to count all replaced elements)

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -2,18 +2,49 @@ package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENT_NUMBER = 10;
+    private static final int ELEMENT_OF_PAIR_NUMBER = 2;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,49 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENT_NUMBER = 10;
+    private static final int ELEMENT_OF_PAIR_NUMBER = 2;
+    private static final int NUMBER_OF_KEY = 0;
+    private static final int NUMBER_OF_VALUE = 1;
+    private final Object[][] storages = new Object[MAX_ELEMENT_NUMBER][ELEMENT_OF_PAIR_NUMBER];
```
zachar-chorny, line 11: Two arrays, or an object with key and value fields, are more suitable for Storage. This variant is difficult to read. Also initialize arrays in the constructor

zachar-chorny (changes requested):

aleksandr-bevza (approved):

PR #844, opened 2022-12-05
Storage solution

stepanpeknyy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    private static final int MAX_NUMBER = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object [MAX_NUMBER];
+        this.values = (V[]) new Object [MAX_NUMBER];
+        size = 0;
+    }
```
aaananas, line 15:
```suggestion
    private int size;

    public StorageImpl() {
        this.keys = (K[]) new Object [MAX_NUMBER];
        this.values = (V[]) new Object [MAX_NUMBER];
    }
```
class fields always have default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = index(key);
+        keys[index] = key;
+        values[index] = value;
+        if (index == size) {
+            size++;
+        }
```
aaananas, line 24: this logic is a bit confusing
lets return -1 from index() method, in general, it is usually considered as "not found" in programming
String methods indexOf() (and similar methods) have the same approach

if -1 is returned that would mean that you should add key and value to the end
else just rewrite value by index

aaananas (changes requested):

Rommelua (approved):

PR #845, opened 2022-12-05
completed java storage task

Ka1zerov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] key;
+    private final V[] value;
+    private int size = 0;
```
aaananas, line 10: class fields always have default values
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+
+    public StorageImpl() {
+        this.key = (K[]) new Object[MAX_SIZE];
+        this.value = (V[]) new Object[MAX_SIZE];
+    }
+
+    public int getPosition(K key) {
```
aaananas, line 17:
```suggestion
    private int getPosition(K key) {
```
this should be private method, because only StorageImpl class should use it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(this.key[i], key)) {
```
aaananas, line 19: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+                this.key[size] = key;
+                this.value[size] = value;
+                this.size++;
+            } else {
+                System.out.println("Storage is full!");
+            }
+        } else {
+            this.value[getPosition(key)] = value;
+        }
     }
```
aaananas, line 39: lets call method getPosition() once and write that value to a variable
because you are calling this method 2 times, and iterate over the array twice

also don't use key word "this" when it is not necessary

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        if (getPosition(key) == -1) {
+            return null;
+        }
+        return this.value[this.getPosition(key)];
```
aaananas, line 46: the same with 2 iterations over the array

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] key;
+    private final V[] value;
```
```suggestion
    private final K[] keys;
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private int size;
+
+    public StorageImpl() {
+        this.key = (K[]) new Object[MAX_SIZE];
+        this.value = (V[]) new Object[MAX_SIZE];
+    }
+
+    private int getPosition(K key) {
```
aaananas, line 16: private methods should go after public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            if (this.size < MAX_SIZE) {
+                this.key[size] = key;
+                this.value[size] = value;
+                this.size++;
+            } else {
+                System.out.println("Storage is full!");
+            }
+        } else {
+            this.value[keyPosition] = value;
+        }
```
```suggestion
        int keyPosition = getPosition(key);
        if (keyPosition == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
        } else {
            this.value[keyPosition] = value;
        }
```

aaananas (changes requested):

Rommelua (approved):

PR #847, opened 2022-12-06
HW-STORAGE

Bogdan-Pokhilko:
In StorageImpl Class created 2 arrays, constant int for storage length and arraysize.

created the constructor for StorageImpl.
implemented put() method, where with help of for loop I've managed to check if key == null and if there is already a key value in the keyArr. implemented get() method, with the for loop where and the same logic as put method. implemented size() method, with for loop to check if elements in the array are not equal to null and return the size of the array.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,46 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_LENGTH = 10;
+    private int arrSize = 0;
+    private K[] keyArr;
+    private V[] valueArr;
+
+    public StorageImpl() {
+        this.keyArr = (K[]) new Object[MAX_STORAGE_LENGTH];
+        this.valueArr = (V[]) new Object[MAX_STORAGE_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < arrSize; i++) {
+            if (key == keyArr[i] || key != null && key.equals(keyArr[i])) {
+                valueArr[i] = value;
+                return;
+            }
+        }
+        keyArr[arrSize] = key;
+        valueArr[arrSize] = value;
+        arrSize++;
     }
 
     @Override
     public V get(K key) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE = 10;
+    private int arrSize = 0;
```
zachar-chorny, line 7:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private K[] keyArr;
+    private V[] valueArr;
```
zachar-chorny, line 9:
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (K key : keyArr) {
+            if (key != null) {
+                return arrSize;
+            }
+        }
+        return arrSize;
```
zachar-chorny, line 46:
```suggestion
        return size;
```

zachar-chorny (changes requested):
Good job

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (K key : keys) {
+            if (key != null) {
+                return size;
+            }
+        }
+        return size;
```
zachar-chorny, line 46:
```suggestion
        return size;
```
Why would you need all this if this line would return the same result?)

zachar-chorny (changes requested):
Not all comments have been fixed.

src/main/java/core/basesyntax/impl/StorageImpl.java
Bogdan-Pokhilko: My bad, I thought I fixed it, rookie mistake :)

karma-o (approved):

PR #850, opened 2022-12-07
homework solution

ivan-vovnyanko:
homework solution

### DIFF
--- a/src/main/java/core/basesyntax/impl/Element.java
+++ b/src/main/java/core/basesyntax/impl/Element.java
@@ -1,0 +1,27 @@
+package core.basesyntax.impl;
+
+public class Element<K, V> {
+    private K key;
+    private V value;
+
+    public K getKey() {
+        return key;
+    }
+
+    public void setKey(K key) {
+        this.key = key;
+    }
+
+    public V getValue() {
+        return value;
+    }
+
+    public void setValue(V value) {
+        this.value = value;
+    }
+
+    public Element(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+}
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,43 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    Element[] elements;
+    int size = 0;
+    {
+        elements = new Element[10];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size != 0) {
+            for (int i = 0; i < size; i++) {
+                if (elements[i].getKey() == key
+                        || elements[i].getKey() != null && elements[i].getKey().equals(key)) {
+                    elements[i].setValue(value);
+                    return;
+                }
+            }
+        }
+        Element element = new Element(key, value);
+        elements[size] = element;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (size != 0) {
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/impl/Element.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.impl;
+
+public class Element<K, V> {
```
aaananas, line 3:
```suggestion
public class Node<K, V> {
```
usually classes that store elements in collections are called Node
also lets make this class an inner class in StorageImpl - because is should not be used outside that class

aaananas, line 3:
```suggestion
public class Node<K, V> {
```
usually classes that store elements in collections are called Node
also lets make this class an inner class in StorageImpl - because is should not be used outside that class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private Element[] elements;
+    private int size = 0;
+
+    {
+        elements = new Element[10];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size != 0) {
```
aaananas, line 15: you dont need this if check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Element[] elements;
+    private int size = 0;
+
+    {
+        elements = new Element[10];
+    }
```
aaananas, line 11:
```suggestion
    private Element[] elements;
    private int size;

    public StorageImpl() {
        elements = new Element[10];
    }
```
class fields always have default values

you should use constructors instead of initialization blocks

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+            }
+        }
+        Element element = new Element(key, value);
+        elements[size] = element;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (size != 0) {
```
aaananas, line 31: you dont need this check here

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Node[] elements;
+    private int size;
+
+    public StorageImpl() {
+        elements = new Node[10];
```
```suggestion
        elements = new Node[MAX_SIZE];
```
avoid magic numbers

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+    }
+
+    public int getSize() {
+        return size;
+    }
+
+    public void setSize(int size) {
+        this.size = size;
+    }
+
```
aaananas, line 58:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,82 @@
...
+    class Node<K, V> {
```
aaananas, line 59:
```suggestion
    private class Node<K, V> {
```
this class should be private because only StorageImpl should be able to use it

aaananas (approved):

PR #859, opened 2022-12-14
Storage solution implemented

dreeeamcatcher:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private K[] keys;
+    private V[] values;
+    private int emptyCell = 0;
```
Rommelua, line 9:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+        return index != -1 ? values[index] : null;
     }
 
     @Override
     public int size() {
+        return emptyCell;
+    }
+
+    private int findIndex(K key) {
+        if (key == null) {
```
Rommelua, line 45: remove this if and use one loop

Rommelua (changes requested):

PR #861, opened 2022-12-15
Hw jv storage solution

Iskamele:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,59 @@
...
+@SuppressWarnings("-Xlint:unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int ENTRY_ITEMS_NUMBER = 10;
+    private final Entry<K, V>[] entries;
+    private int size;
+
+    public StorageImpl() {
+        entries = new Entry[ENTRY_ITEMS_NUMBER];
+        size = 0;
```
kozhukhovsky, line 15:
```suggestion
```

0 setting by default for int variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,59 @@
...
+                return;
+            }
+        }
+        try {
+            entries[size] = new Entry<>(key, value);
+            size++;
+        } catch (IndexOutOfBoundsException e) {
+            throw new RuntimeException("The storage can have a maximum of"
+                    + ENTRY_ITEMS_NUMBER + " entries/", e);
+        }
```
kozhukhovsky, line 32:
```suggestion
        try {
            entries[size] = new Entry<>(key, value);
            size++;
        } catch (IndexOutOfBoundsException e) {
            throw new RuntimeException("The storage can have a maximum of"
                    + ENTRY_ITEMS_NUMBER + " entries/", e);
        }
```

Not required for this assignment. But it's great that you figured it out!

kozhukhovsky (approved):

PR #862, opened 2022-12-16
jv-storage first solution

kcasPilot:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                ? values[getKeyIndex(key)]
+                : null;
     }
 
     @Override
     public int size() {
+        int count = 0;
+        for (V datum : values) {
+            if (datum != null) {
+                count++;
```
zachar-chorny, line 38: just create class level variable - size and increase it during additions

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        this.keys[lastIndex] = key;
+        this.values[lastIndex] = value;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return getKeyIndex(key) >= 0
+                ? values[getKeyIndex(key)]
+                : null;
```
```suggestion
        return getKeyIndex(key) >= 0 ? values[getKeyIndex(key)] : null;
```
If you can call another method just once it is better to do so

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                count++;
+            }
+        }
+        return count;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null && key == null)
+                    || ((keys[i] != null && key != null) && keys[i].equals(key))) {
```
zachar-chorny, line 47:
```suggestion
            if ((keys[i] ==key) || ((keys[i] != null) && keys[i].equals(key))) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            this.values[getKeyIndex(key)] = value;
+            return;
+        }
+        int lastIndex = size();
+        this.keys[lastIndex] = key;
+        this.values[lastIndex] = value;
```
zachar-chorny, line 23:
```suggestion
        int index = getKeyIndex(key);
        if (index != -1) {
            this.values[index] = value;
            return;
        }
        this.keys[size] = key;
        this.values[size] = value;
        size++;
```

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
kcasPilot: In this case your variant doesn't work correctly because in condition (line 17 my code) I'm checking the value (not index of key) even if we receive key == null.
So if key == null and value != null I understand that we have an existing pair and I have to change the value only.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     }
 
     @Override
     public int size() {
+        return dataSize;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == key) || ((keys[i] != null && key != null) && keys[i].equals(key))) {
```
zachar-chorny, line 41:
```suggestion
            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int dataSize;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_ELEMENTS_NUMBER];
+        this.values = (V[]) new Object[MAX_ELEMENTS_NUMBER];
+        this.dataSize = 0;
```
zachar-chorny, line 14:
```suggestion
```
it's 0 by deafult

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
```
zachar-chorny, line 19: if inside getKeyIndex use dataSize instead of keys.length everything works wonderfully. Do as I did last time here
Screen Shot 2022-12-18 at 1 29 52 PM

zachar-chorny (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
kcasPilot: It works ) Thanks!

zachar-chorny (approved):
Good job!

PR #868, opened 2022-12-17
Hw jv storage are completed

taras-zhyravetskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private int size;
+    private V[] valueArray = (V[]) new Object[MAX_ELEMENTS_NUMBER];
+    private K[] keyArray = (K[]) new Object[MAX_ELEMENTS_NUMBER];
```
lysytskapolina, line 9: All non-constant fields should be initialized in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        for (int i = 0; i < MAX_ELEMENTS_NUMBER; i++) {
+            if (valueArray[i] == null && keyArray[i] == null) {
+                valueArray[i] = value;
+                keyArray[i] = key;
+                size++;
+                return;
+            }
+            if (keysEquals(key, keyArray[i])) {
+                valueArray[i] = value;
+                return;
```
lysytskapolina, line 22: Try not to complicate if-else with return constructions, maybe just one if() will be enough

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    }
+
+    private boolean keysEquals(K firstObj, K secondObj) {
+        if (firstObj == secondObj) {
+            return true;
+        }
+        if (firstObj == null || secondObj == null) {
+            return false;
+        }
+        return firstObj.equals(secondObj);
```
lysytskapolina, line 49: (suggested change) Maybe it's more practical to use a ternary operator here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean keysEquals(K firstObj, K secondObj) {
```
lysytskapolina, line 42: (suggested change) Is it so necessary to create this method? Try to optimize the if() statement in your put() method.

lysytskapolina (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
taras-zhyravetskyi: Left one if construction and change max i value in a loop from MAX_ELEMENTS_NUMBER to size

src/main/java/core/basesyntax/impl/StorageImpl.java
taras-zhyravetskyi: Rewrote the method

src/main/java/core/basesyntax/impl/StorageImpl.java
taras-zhyravetskyi: I prefer to put the comparison of the input and existing keys into a separate method since this operation is repeated twice in the code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private int size;
+    private V[] valueArray;
+    private K[] keyArray;
```
lysytskapolina, line 9: (suggested only) how about names keys and values?)

lysytskapolina (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
Nick97-git, line 9: remove `Array`

Nick97-git (approved):

PR #877, opened 2022-12-19
done

Richaa2:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final StorageImpl<K,V>[] list = new StorageImpl[MAX_ITEMS_NUMBER];
```
nacenik, line 7: According to the Single-Responsibility Principle: 

> A class should have one and only one reason to change, meaning that a class should have only one job.

Better to create inner class `Node` or `Pair,` to segregate implementation logic from saving data type logic.

Also, move object creation to the no-args constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+    private K typeK;
```
nacenik, line 8: Move this field to the inner class. Fix in all places
```suggestion
    private K key;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+    public StorageImpl(K typeK, V nameV) {
+        this.typeK = typeK;
+        this.typeV = nameV;
+    }
+
+    public StorageImpl() {
+    }
+
     @Override
     public void put(K key, V value) {
```
nacenik, line 21: Please, refactor this method using only one loop.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (list[i].typeK == null) {
+                    list[i].typeV = value;
+                    size--;
```
nacenik, line 26: Your storage should contain `key == null `. Don't change the size in this case.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+            }
+        }
+        this.list[size] = new StorageImpl<>(key, value);
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && list[i].typeK == null) {
```
nacenik, line 49: Why your key-value storage could contain a null value and key, but couldn't contain a null key and not-null value?

nacenik (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Richaa2: i`m sorry, i don't understand fully,
 my storage can contain null key and not-null value because it's been in tests

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
+
+    class Pair {
+        private static final int MAX_ITEMS_NUMBER = 10;
+        private final StorageImpl.Pair[] list = new StorageImpl.Pair[MAX_ITEMS_NUMBER];
```
aleksandr-bevza, line 52: "Pair" is an element in your storage. It should have only "key" and "value" fields. "Pair" should not have array of other Pairs
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+
+        private Pair(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+
+        private Pair() {
+        }
+    }
+
```
aleksandr-bevza, line 64: empty line

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #882, opened 2022-12-19
storage can save the key

logoltip:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_COUNT = 10;
```
Dmytro-Mk, line 6:
```suggestion
    private static final int MAX_ELEMENTS_NUMBER = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_COUNT];
+        values = (V[]) new Object[MAX_STORAGE_COUNT];
```
Dmytro-Mk, line 13:
```suggestion
        keys = new Object[MAX_STORAGE_COUNT];
        values = new Object[MAX_STORAGE_COUNT];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || key != null && key.equals(keys[i])) {
+                return values[i];
+            }
+        }
```
Dmytro-Mk, line 35: let's move duplicate code to separate method: areKeysEqual(Object firstKey, Object secondKey)

Dmytro-Mk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
logoltip: Common mistakes says: Don’t use class Objects

Dmytro-Mk: Object**s** is an util class, we are not using it here

Dmytro-Mk (approved):

Nick97-git (approved):

PR #883, opened 2022-12-20
implemented class StorageImpl<K, V>

andriiyaroslav:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
```
Dastya, line 6:
```suggestion
    private static final int MAX_NUMBER_OF_ELEMENTS =  10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private int size;
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_CAPACITY];
+        this.values = (V[]) new Object[MAX_CAPACITY];
```
Dastya, line 13: you can do keys and values as Object arrays, so you can avoid unchecked cast
also you don't need to use "this." for keys and values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
```
Dastya, line 19: let's move this check to a separate bolean method

Dastya (changes requested):

PR #899, opened 2022-12-24
implements all methods

Butovych:


lysytskapolina (approved):
👍

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                return values[i];
+            }
+        }
```
KostyaNikiforov, line 35: I see common logic here. Let's move it in a private method

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Butovych: Done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keyEquals(key, i)) {
+                values[i] = value;
+                return;
+            }
+        }
```
KostyaNikiforov, line 23: You can move a whole loop which is searching for an index at the separate method

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        if (getIndex(key) == -1) {
+            return null;
+        }
+        return values[getIndex(key)];
```
KostyaNikiforov, line 32:
```suggestion
        int index = getIndex(key);
        if (index == -1) {
            return null;
        }
        return values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAXIMUM_CAPACITY];
+        values = (V[]) new Object[MAXIMUM_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (getIndex(key) != -1) {
+            values[getIndex(key)] = value;
```
KostyaNikiforov, line 19: same)

KostyaNikiforov (approved):

PR #912, opened 2023-01-08
Homework jv-storage done

Oleh8:


src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,43 @@
+package core.basesyntax.impl;
+
+public class Pair<K, V> {
```
aaananas, line 3: lets make this class an inner class in StorageImpl 
you wont need getters that way

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxSize = 10;
```
aaananas, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```
this should be a constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    private Pair[] pairArr;
```
```suggestion
    private Pair<K, V>[] pairArray;
```
use full names

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+                    || (pairArr[i].getKey() != null && pairArr[i].getKey().equals(key)))) {
+                pairArr[i] = new Pair<>(key, value);
+                return;
+            } else if (pairArr[i] == null) {
+                pairArr[i] = new Pair<>(key, value);
+                actualSize++;
+                return;
+            }
+        }
     }
```
aaananas, line 28:
```suggestion
    public void put(K key, V value) {
        for (int i = 0; i < actualSize; i++) {
            if (pairArr[i].getKey() == key)
                    || (pairArr[i].getKey() != null && pairArr[i].getKey().equals(key)))) {
                pairArr[i] = new Pair<>(key, value);
                return;
            } 
        }
        pairArr[i] = new Pair<>(key, value);
        actualSize++;
    }
```

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Oleh8: a cant pass tests if I put here capital letters

src/main/java/core/basesyntax/impl/StorageImpl.java
Oleh8: this:
 pairArr[i] = new Pair<>(key, value);
is out of for loop...  does not work that way

PR #917, opened 2023-01-11
mik-jv-storage solution sent for review ch1, 11.01.23  22-59

911Mik:


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
okuzan, line 4:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxNumber = 10;
```
okuzan, line 6: this should be class-level constant, not just final field. Consider more precise naming, like DEFAULT_CAPACITY

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        int foundKey = -1;
+        for (int i = 0; i <= realSize; i++) {
+            if (key == null && keys[i] == null) {
+                foundKey = i;
+                break;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                foundKey = i;
+                break;
+            }
```
okuzan, line 22: this can be done with one if, and you have the same logic in the get method. Move that logic to a new method and call it here and in `get`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final int maxNumber = 10;
+    private int realSize = -1;
```
okuzan, line 7: let's repurpose this to `size`  or `index` variable which will mean how many entries are there in our storage. Its having a value of -1 is a bit misleading, since size can't be negative. Instead, create a new method (more about it in another comment) that will return an index based on the passed K key, and if the value is not found - return -1.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int realSize = 0;
+    private K[] keys = (K[]) new Object[DEFAULT_CAPACITY];
+    private V[] values = (V[]) new Object[DEFAULT_CAPACITY];
```
okuzan, line 9: initialize these fields in [constructor](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int realSize = 0;
```
```suggestion
    private int size = 0;
```

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private int size = 0;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_CAPACITY];
+
```
okuzan, line 13: extra empty line

okuzan (approved):
Good job! 👍

PR #919, opened 2023-01-14
maven checked

BoShMan78:


README.md
```diff
@@ -1,18 +1,18 @@
...
-Storage<Integer, Box> storage = new StorageImpl<>(); 
+StorageImpl<Integer, Box> storage = new StorageImpl<>(); 
 Box box = new Box();
 storage.put(22, box); 
 Box value = storage.get(22); // returns the Box
 int size = storage.size(); // returns storage size
 ```
 You should solve this task using one or two arrays (this depends on your implementation). 
-If you use arrays for your solution, we assume that the maximum number of elements in our `Storage` is 10.
+If you use arrays for your solution, we assume that the maximum number of elements in our `StorageImpl` is 10.
```
karma-o, line 17: remove this changes

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_NUMBER = 10;
+    private final Object[] keyArray = new Object[MAX_NUMBER];
+    private final Object[] valueArray = new Object[MAX_NUMBER];
```
karma-o, line 6: you should initialize these fields in constructor, like this: `keyArray = (K[]) new Object[MAX_NUMBER];`
```suggestion
    private final K[] keyArray;
    private final V[] valueArray;
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void put(Object key, Object value) {
+        boolean alreadyContain = false;
```
karma-o, line 10:
```suggestion        
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public int size() {
+        int counter = 0;
+        for (int i = 0; i < 10; i++) {
+            if (null != valueArray[i]) {
+                counter++;
+            }
+        }
+        return counter;
+    }
```
karma-o, line 56: remove this, add a `size` field to your storage, and increment it each time you are adding a new value
```Suggestion
    public int size() {
        return size;
    }
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_NUMBER = 10;
+    private final Object[] keyArray = new Object[MAX_NUMBER];
+    private final Object[] valueArray = new Object[MAX_NUMBER];
+
+    @Override
+    public void put(Object key, Object value) {
+        boolean alreadyContain = false;
+        for (int i = 0; i < MAX_NUMBER; i++) {
```
karma-o, line 11: iterate only up to the `size`, this will boost your performance, check other loops like this

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
...
+                    valueArray[i] = value;
+                    alreadyContain = true;
+                    break;
+                } else {
+                    continue;
+                }
+            } else if (key.equals(keyArray[i])) {
+                valueArray[i] = value;
+                alreadyContain = true;
+            }
```
karma-o, line 23: make it a single `if` with a `keyArray[i] == key || key != null && key.equals(keyArray[i])` condition this will cover both null and an equals cases.
Same thing with the `get(key)` method

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+        if (!alreadyContain) {
+            keyArray[size()] = key;
+            valueArray[size()] = value;
+        }
+    }
+
+    @Override
+    public Object get(Object key) {
```
karma-o, line 32:
```suggestion
    public Object get(K key) {
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_NUMBER = 10;
+    private final Object[] keyArray = new Object[MAX_NUMBER];
+    private final Object[] valueArray = new Object[MAX_NUMBER];
+
+    @Override
+    public void put(Object key, Object value) {
```
karma-o, line 9: why did we need to parametrize a class if you didn't use it in any way?)
```suggestion
    public void put(K key, V value) {
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_NUMBER = 10;
```
karma-o, line 4: Let's be clear with naming and access modifiers
```suggestion
    private static final int MAX_SIZE = 10;
```

src/test/java/core/basesyntax/StorageImplImplTest.java
```diff
@@ -1,10 +1,9 @@
 package core.basesyntax;
 
-import core.basesyntax.impl.StorageImpl;
 import org.junit.Assert;
 import org.junit.Test;
 
-public class StorageImplTest {
+public class StorageImplImplTest {
```
karma-o, line 6: remove this changes

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,57 @@
+package core.basesyntax;
```
karma-o, line 1: let's not move the files to some other directories and let them stay where they were, in the future

karma-o (changes requested):

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,53 @@
...
+                    alreadyContain = true;
+                    break;
+                } else {
+                    continue;
+                }
+            } else if (key.equals(keyArray[i])) {
+                valueArray[i] = (V) value;
+                alreadyContain = true;
+            }
+        }
```
aaananas, line 25:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keyArray[i] || key != null && key.equals(keyArray[i])) {
                    valueArray[i] = (V) value;
                    return;
                }
        }
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        if (!alreadyContain) {
+            keyArray[size()] = (K) key;
+            valueArray[size()] = (V) value;
+            size++;
+        }
```
aaananas, line 30:
```suggestion
            keyArray[size] = key;
            valueArray[size] = value;
            size++;
        
```

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,53 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private final V[] valueArray = (V[]) new Object[MAX_SIZE];
```
aaananas, line 6: see common mistakes to fix it

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (key == keyArray[i]) {
+                    return (V) valueArray[i];
+                } else {
+                    continue;
+                }
+            } else if (key.equals(keyArray[i])) {
+                return (V) valueArray[i];
+            }
```
aaananas, line 44: simplify it like i did `put`

aaananas (changes requested):

README.md
karma-o: not fixed

src/main/java/core/basesyntax/StorageImpl.java
karma-o: not fixed, try to remove this variable

karma-o (changes requested):

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,42 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keyArray;
+    private final V[] valueArray;
+    
+    private int size = 0;
```
aaananas, line 8:
```suggestion
    private int size;
```

aaananas (approved):

PR #923, opened 2023-01-17
solved task

AntonKurylo:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Pair<K, V>[] pairs;
+    private int size = 0;
```
Rommelua, line 8:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+
+    public StorageImpl() {
+        this.pairs = new Pair[MAX_SIZE];
+    }
+
+    private boolean isEquals(Object firstOne, Object secondOne) {
```
Rommelua, line 14: Private ones (methods, classes) should be after public ones.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (isEquals(pairs[i].getKey(), key)) {
+                pairs[i].setValue(value);
+                return;
+            }
+        }
+
```
Rommelua, line 26: empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (isEquals(pairs[i].getKey(), key)) {
+                return pairs[i].getValue();
+            }
+        }
+
```
Rommelua, line 38: empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+        pairs[size] = new Pair<>(key, value);
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Rommelua, line 33: use for-each

Rommelua (changes requested):
good work

src/main/java/core/basesyntax/impl/StorageImpl.java
AntonKurylo: I will have null values and have to check for that if I use for-each. I don't think it's necessary to use for-each.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+    public StorageImpl() {
+        this.pairs = new Pair[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (isEquals(pairs[i].getKey(), key)) {
+                pairs[i].setValue(value);
+                return;
```
Rommelua, line 19:
```suggestion
        for (Pair pair : pairs) {
            if (isEquals(pair.getKey(), key)) {
                pair.setValue(value);
                return;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
Rommelua: For-each is easier to read. It is preferred to use **always** when you do not need **i** counter and you need just iterate through the elements. 
For-i loop for iteration we use only in some tricky issues - if you need to compare array[i] with array[i+1] for example.
Here you need just check all pairs one by one, so, you do not need **i** itself - it is an extra variable.

Rommelua (changes requested):

Rommelua (approved):

PR #924, opened 2023-01-19
realisation class StorageImpl

bigvini:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size = 0;
+    private K[] keys = (K[]) new Object[MAX_SIZE];
+    private V[] values = (V[]) new Object[MAX_SIZE];
+
+    public StorageImpl() {
+    }
```
okuzan, line 13: instead of having empty constructor, better do like described in the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+                values[size] = value;
+            }
+        }
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (Objects.equals(key, keys[i])) {
```
okuzan, line 36: [checklist issue](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-use-class-objects)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+    private V[] values = (V[]) new Object[MAX_SIZE];
+
+    public StorageImpl() {
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == 0) {
+            values[0] = value;
+            keys[0] = key;
```
okuzan, line 19: rewrite your code so that you don't need to handle some edge case specifically

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(key, keys[i]) || key == keys[i]) {
```
okuzan, line 22: firstly don't use objects, secondly, you should reuse this logic which you have in two methods, extract it to a separate method, like indexOf(), and call it here and in get()

okuzan (changes requested):

aleksandr-bevza (approved):

PR #925, opened 2023-01-20
implemented methods from interface Storage

yaroslav-ometiukh:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int now;
```
Ivan95kos, line 9: not informative naming
```suggestion
    private int size;
```

Ivan95kos (approved):

PR #930, opened 2023-01-28
commit#1

Andrei-Bondarenko-MRPL:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+            }
+        }
+        arrayKey[size] = key;
+        arrayValue[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_SIZE; i++) {
```
karma-o, line 31: Checking up to the `size` would be enough since we know for sure that all other indexes after that will be empty
```suggestion
        for (int i = 0; i < size; i++) {
```

karma-o (changes requested):
Well done! Except for this minor comment, everything looks good

karma-o (approved):
:fire:

PR #932, opened 2023-01-30
StorageImpl class logic was implemented

MaxSkory:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null) {
+                return i;
+            }
+        }
+        return ARRAY_SIZE;
```
okuzan, line 61: better create `size` variable and keep track of it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                while (keys[i] != null) {
+                    i++;
+                }
+                values[i] = value;
+                return;
+            }
+            if (key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
```
okuzan, line 34: this can be simpler
insert key and value and the index of size++;
for equality combine your checks with one-liner: 
`if (key == null && keys[i] == null || key != null && key.equals(keys[i]))`

okuzan (changes requested):

okuzan (changes requested):
Let me suggest a better approach: create a method that will search for key in keys, and returns its index if found, or -1 if not. You can then reuse this method in put and get. You will have simple if-else statement, if index is returned, set V value in values array at that index, if not - set new key-value at size++;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_DEFAULT_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size = 0;
```
```suggestion
    private int size;
```
class variables have default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
aaananas, line 42: see common mistakes to fix it

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_DEFAULT_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
andriy-shulepa, line 8: use K and V types instead of Object

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
MaxSkory: why?

andriy-shulepa: So you don't need cast elements later every time when you operate with them. LIke in your code on line 31

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        } else {
+            values[keyIndex] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int keyIndex = getKeyIndex(key);
+        return keyIndex == -1 ? null : (V) values[keyIndex];
```
andriy-shulepa, line 31:
```suggestion
        return keyIndex == -1 ? null : values[keyIndex];
```

andriy-shulepa (approved):

PR #933, opened 2023-01-30
class StorageImpl is implemented

YaroslavKostenko89:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K key;
+    private V value;
+    private int size = 0;
+    private final StorageImpl<K, V>[] storage = new StorageImpl[MAX_SIZE];
+
+    public StorageImpl() {
+    }
```
okuzan, line 13: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

okuzan: why do you need recursion? You have StorageImpl containing an array of itself. Just use an Object[] array for keys, and another one for values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+            storage[size] = new StorageImpl<>(key, value);
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == storage[i].getKey()
+                    || (storage[i].getKey() != null && storage[i].getKey().equals(key))) {
```
okuzan, line 40: extract to an external method that will return index based on key

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                storage[i].setValue(value);
+                isExist = true;
+            }
+        }
+        if (!isExist && size < MAX_SIZE) {
+            storage[size] = new StorageImpl<>(key, value);
+            size++;
+        }
```
okuzan, line 33: rewrite with a call to the external method, so that you will have if-else statement and isExist won't be needed

okuzan (approved):
Good job!

PR #936, opened 2023-02-02
hw-10-solution

serhii-shyian:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_ARRAY = 10;
+    private final K[] keys = (K[]) new Object[MAX_SIZE_ARRAY];
+    private final V[] values = (V[]) new Object[MAX_SIZE_ARRAY];
```
aaananas, line 8: see common mistakes to fix it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (getIndex(key) >= 0) {
+            return values[getIndex(key)];
+        }
         return null;
```
aaananas, line 28:
```suggestion
        int index = getIndex(key);
        if (index  != -1) {
            return values[index];
        }
        return null;
```

aaananas (changes requested):

andriy-shulepa (approved):
Good job!

PR #940, opened 2023-02-04
hw-jv-storage

ihor-sydorenko:


src/test/java/core/basesyntax/StorageImplTest.java
```diff
@@ -42,6 +42,13 @@ public void addElementsToStorage() {
     @Test
```
aaananas, line 42: dont change Test file, return the previous version

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private int arraySize;
```
Rommelua, line 7:
```suggestion
    private int size;
```

Rommelua (approved):

PR #941, opened 2023-02-05
Implemented three methods, created constant and fields, override two …

BetelgeuseUA:
…methods hashCode(), equals()

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,60 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static int INDEX = 0;
+    private K key;
+    private V value;
+    private StorageImpl[] storage;
+
+    public StorageImpl() {
+        storage = new StorageImpl[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        this.key = key;
+        this.value = value;
+        storage[INDEX] = this;
+        INDEX++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V valueStorage = null;
+        for (int i = 0; i < storage.length; i++) {
+            if (storage[i].key.equals(key)) {
+                valueStorage = (V) storage[i].value;
+            }
+        }
+        return valueStorage;
     }
 
     @Override
     public int size() {
-        return -1;
+        return INDEX + 1;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = 17;
+        if (key != null) {
+            result = 31 * result + key.hashCode();
+            result = 31 * result + value.hashCode();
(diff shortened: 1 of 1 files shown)


andriy-shulepa (changes requested):
Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the mvn clean package command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,96 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private int lastIndexOfArray = 0;
+    private int index = 0;
```
```suggestion
    private int size;
```
you can store just 1 value
```size``` is a perfect name, it stores current size, and last available index

File: src/main/java/core/basesyntax/impl/StorageImpl.java [hunk 2]
```diff
@@ -3,17 +3,96 @@
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if ((obj instanceof StorageImpl)) {
+            return false;
+        }
+        return obj.hashCode() == keys[index].hashCode();
+    }
```
aaananas, line 72:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,96 @@
...
+                return values[index];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return lastIndexOfArray;
```
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,96 @@
...
 
     @Override
     public V get(K key) {
+        index = 0;
+        for (; index < lastIndexOfArray; index++) {
+            if (isKey(key)) {
+                return values[index];
+            }
+        }
         return null;
```
aaananas, line 43:
```suggestion
        for (int i = 0; i < size; i++) {
            if (keysEquals(key, keys[i])) {
                return values[index];
            }
        }
        return null;
```
index should be local - here index is the same as i

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,96 @@
...
+                values[index] = value;
+            } else {
+                saveStorage(key, value);
+            }
+            index = 0;
+        } else if (isKey(key)) {
+            values[index] = value;
+        } else {
+            saveStorage(null, value);
+        }
```
aaananas, line 32: too complicated, use something similar as in my comment to get() method
iterate over the array and replace if keys are equals
after the loop - place a new key-pair value on size index (remember size is also the next available place)

aaananas (changes requested):

Ivan95kos (approved):
Great job

PR #942, opened 2023-02-05
solved task

slaybrute:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_STORAGE_VALUE = 10;
+    private int createdStorages;
+
+    private ArrayList<K> keyArrayList = new ArrayList<>();
+    private ArrayList<V> valueArrayList = new ArrayList<>();
```
okuzan, line 12: [This is not allowed](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-use-any-kind-of-list-set-or-map). Create new implementation

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
okuzan, line 7: delete empty line

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
slaybrute: Okay, you mean It is better to create array with type Object, instead of ArrayList ?

slaybrute: I've already altered, but applying ArrayList was too much convenient than Array. I can see, than I have added extra code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+    }
+
+    private int indexOf(Object[] array, Object o) {
+        for (int i = 0; i < createdStorages; i++) {
+            if (Objects.equals(array[i], o)) {
+                return i;
+            }
+        }
+        return -1;
+    }
```
Ivan95kos, line 31: All private methods should be below public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+    private V value;
+
+    private boolean contains(Object[] array, Object o) {
+        for (int i = 0; i < createdStorages; i++) {
+            if (Objects.equals(array[i], o)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
Ivan95kos, line 22: do not use such naming(o), how to guess what it stores?
```suggestion
    private boolean contains(Object key) {
        for (int i = 0; i < size; i++) {
            if (Objects.equals(keys[i], key)) {
                return true;
            }
        }
        return false;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+
+    private int indexOf(Object[] array, Object o) {
```
Ivan95kos, line 24: do the same as for "private boolean contains(Object key)"

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_VALUE = 10;
+    private int createdStorages;
+    private Object[] keyArrayList = new Object[MAX_STORAGE_VALUE];
+    private Object[] valueArrayList = new Object[MAX_STORAGE_VALUE];
+
+    private K key;
+    private V value;
```
Ivan95kos, line 13: useless variables
```suggestion
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+        }
+        return (V) values[indexOf(keys, key)];
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private boolean contains(Object[] array, Object key) {
```
kozhukhovsky, line 39: These two methods are similar, try using one.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_VALUE = 10;
+    private int size;
+    private Object[] keys = new Object[MAX_STORAGE_VALUE];
+    private Object[] values = new Object[MAX_STORAGE_VALUE];
+
     @Override
     public void put(K key, V value) {
+        if (size < MAX_STORAGE_VALUE) {
+            if (contains(keys, key)) {
```
kozhukhovsky, line 15: You can use indexOf checking here. If the index is not found, it returns -1. That way you can remove the contains method.

kozhukhovsky (changes requested):

kozhukhovsky (approved):

PR #943, opened 2023-02-06
internal test passed

iamchgi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_CAPACITY = 10;
+    private final Object[] indexes;
+    private final Object[] values;
+
+    private int storageCapacity = 0;
```
```suggestion
    private int size;
```
class fields have default values (local variables are different - they have to be initialized)

aaananas (approved):
great job

PR #948, opened 2023-02-12
relized task

kromch:


mishok321 (changes requested):
1. do not push to the master branch, create another one
2. read common mistakes before you get to work

PR #950, opened 2023-02-13
relized task

kromch:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,23 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final HashMap<K, V> storage = new HashMap<>();
```
potrap, line 7: the checklist says that in this task we shouldn`t use any kind of List, Set or Map

potrap (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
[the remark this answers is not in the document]
kromch: re-done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_RANGE = 10;
+    private K[] keysRange;
+    private V[] valuesRange;
+
```
sarakhmen, line 9: Let's introduce a new class field 'index', indicating where the next element will be placed. Then you put() method logic will be the following: if a key is present in a key array then replace the value, if not then add a new key and value into appropriate arrays. Also be careful not to iterate to the maximum capacity, but to the 'index' value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    public StorageImpl() {
+        keysRange = (K[]) new Object[MAX_RANGE];
+        valuesRange = (V[]) new Object[MAX_RANGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int i = 0;
+        for (; i < MAX_RANGE; i++) {
```
sarakhmen, line 18: Why don't you initialize 'i' in the dedicated loop section?

sarakhmen (changes requested):
Please redo the work according to my comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+            if (keysRange[i] == key || keysRange[i] == valuesRange[i]
+                    || (keysRange[i] != null && keysRange[i].equals(key))) {
```
outref, line 20: If you already have a pair in your Strorage that has same key and value (like key "123" and value "123"), 
keysRange[i] == valuesRange[i] would return true and your method will not work as intended.
Same in get().

outref (commented):
kind suggestions)

src/main/java/core/basesyntax/impl/StorageImpl.java
kromch: refactored

src/main/java/core/basesyntax/impl/StorageImpl.java
kromch: without this conditions test are fail. i did try many variants

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+            }
+        }
+        return valuesRange[i];
     }
 
     @Override
     public int size() {
-        return -1;
+        int i = 0;
+        for (;i < MAX_RANGE; i++) {
```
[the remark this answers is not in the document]
kromch: in last commit is to display this logic has explicit

src/main/java/core/basesyntax/impl/StorageImpl.java
kromch: in my first local variant was be same idea but later i did refuse her.  i returned this variable again

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_RANGE = 10;
+    private K[] keysRange;
+    private V[] valuesRange;
+    private int index;
+
+    public StorageImpl() {
+        keysRange = (K[]) new Object[MAX_RANGE];
+        valuesRange = (V[]) new Object[MAX_RANGE];
+        index = 0;
```
petro-ushchuk, line 14: `index` already equals `0`

petro-ushchuk (approved):

petro-ushchuk (approved):

petro-ushchuk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int i = 0;
+        for (; i < index; i++) {
+            if (keysRange[i] == key || (keysRange[i] != null && keysRange[i].equals(key))) {
```
yurihhhT, line 21: this equals you can make in other method
like
``` 
private boolean compareKeys(K key1, K key2) {
        return key1 == key2 || (key1 != null && key1.equals(key2));
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                valuesRange[i] = value;
```
yurihhhT, line 22:
```suggestion
        for (; i < index; i++) {
            if (compareKeys(keysRange[i], key) {
                valuesRange[i] = value;
      ```

you can make this equals in different metod

private boolean compareKeys(K key1, K key2) {
        return key1 == key2 || (key1 != null && key1.equals(key2));
    }

yurihhhT: you can make this equals in different metod

private boolean compareKeys(K key1, K key2) {
        return key1 == key2 || (key1 != null && key1.equals(key2));
    }

yurihhhT (approved):

yurihhhT (approved):

Hlib13 (commented):
.

Hlib13 (approved):

PR #952, opened 2023-02-14
I implement class StorageImpl

NazarZva:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH = 10;
+    private static final int FIRST_INDEX = 0;
```
GalubaGSV, line 7: You don't need this constant.

GalubaGSV (changes requested):
A small flaw

GalubaGSV (approved):
Good job!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+
     @Override
     public void put(K key, V value) {
+        boolean keyWas = false;
+        for (int i = 0; i < index; i++) {
+            if (compare(this.key[i], key)) {
+                this.value[i] = value;
+                keyWas = true;
+            }
+        }
```
sarakhmen, line 24: I'd recommend something like this
```suggestion
        for (int i = 0; i < index; i++) {
            if (compare(this.key[i], key)) {
                this.value[i] = value;
                return
            }
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        if (keyWas == false) {
+            this.key[index] = key;
+            this.value[index] = value;
+            index++;
+        }
+    }
+
+    private boolean compare(Object first, Object second) {
```
sarakhmen, line 32: This method is used only for keys comparison, so could you change the method signature considering it?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        if (first == null || second == null) {
+            if (first == second) {
+                return true;
+            } else {
+                return false;
+            }
+        } else {
+            return first.equals(second);
+        }
```
sarakhmen, line 41: let's simplify it
```suggestion
     return key1 == key2 || (key1 != null && key1.equals(key2));
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
     }
 
     @Override
     public V get(K key) {
+        int indexResult = 0;
+        for (int i = 0; i < index; i++) {
+            if (compare(this.key[i], key)) {
```
sarakhmen, line 48: 'this' referencing is redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+                return false;
+            }
+        } else {
+            return first.equals(second);
+        }
     }
 
     @Override
     public V get(K key) {
+        int indexResult = 0;
```
sarakhmen, line 46: Remove this unused variable

sarakhmen (changes requested):
Good job, let's improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
NazarZva: why? we have two key in this method, so if i will delete 'this', compilator woduldn`t know what key must we it use

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH = 10;
```
karma-o, line 6:
```suggestion
    private static final int INITIAL_LENGTH = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            }
+        }
+        this.key[index] = key;
+        this.value[index] = value;
+        index++;
+    }
+
+    private boolean compareKeys(Object first, Object second) {
+        return (first == null || second == null)
+                ? ((first == second) ? true : false) : first.equals(second);
```
karma-o, line 31:
```suggestion
        return first == second || first != null && first.equals(second);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH = 10;
+    private V[] value;
+    private K[] key;
+    private int index;
```
karma-o, line 9:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                this.value[i] = value;
+                return;
+            }
+        }
+        this.key[index] = key;
+        this.value[index] = value;
+        index++;
+    }
+
+    private boolean compareKeys(Object first, Object second) {
```
karma-o, line 29: private methods should go after public ones

karma-o (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (compareKeys(this.keys[i], key)) {
+                this.values[i] = value;
+                return;
+            }
+        }
+        this.keys[size] = key;
+        this.values[size] = value;
```
KseniiaMakarova, line 25:
```suggestion
            if (compareKeys(keys[i], key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
```
avoid `this` if there is no necessity in it
code is easier to read

PR #954, opened 2023-02-14
Storage hw

Dominskyi93:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+                if (array[i].getKey().equals(key)) {
+                    array[i] = instanceOfKeyValue;
+                    isExist = true;
+                    break;
+                }
+            } else if (array[i].getKey() == key) {
+                array[i] = instanceOfKeyValue;
+                isExist = true;
+                break;
+            }
```
MaksymHorielikov, line 25:
```suggestion
            if (array[i].getKey() == key || array[i].getKey() != null 
                    && array[i].getKey().equals(key)) {
                array[i] = instanceOfKeyValue;
                isExist = true;
                break;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        }
+        if (!isExist) {
+            array[size] = instanceOfKeyValue;
+            size++;
```
MaksymHorielikov, line 29: what happens if the size exceeds START_CAPACITY? You should probably throw an exception.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (array[i].getKey() != null) {
+                if (array[i].getKey().equals(key)) {
+                    return array[i].getValue();
+                }
+            } else if (array[i].getKey() == key) {
+                return array[i].getValue();
+            }
```
MaksymHorielikov, line 42:
```suggestion
            if (array[i].getKey() == key || array[i],getKey() != null 
                     && array[i].getKey().equals(key)) {
                return array[i].getValue();                
            }
```

MaksymHorielikov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_CAPACITY = 10;
+    private final KeyValuePair<K, V>[] array = new KeyValuePair[START_CAPACITY];
+    private int size = 0;
+
     @Override
     public void put(K key, V value) {
+        if (size == 10) {
+            throw new RuntimeException("The storage is full, you can replace an existing element!");
+        }
```
MaksymHorielikov, line 14: Here you need to use a constant instead of 10 and this check should be used later, before trying to write a new key with a value to the store (line 26). Your option does not allow you to change the value of an existing key in the store, in the case when the size is already equal to 10

MaksymHorielikov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Dominskyi93: I think it will be more effective to first check the storage size, and then try to put elements there. And if there are already 10 elements, throw an exception

MaksymHorielikov: In this case, we are not talking about efficiency.
Method algorithm:
first you check if there is such a key in the storage, if there is, change its value, otherwise
check whether it is possible to add a new key, if not then an exception, if yes then add.

MaksymHorielikov (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                    && array[i].getKey().equals(key)) {
+                array[i] = instanceOfKeyValue;
+                isExist = true;
+                break;
+            }
+        }
+        if (!isExist) {
+            if (size == START_CAPACITY) {
+                throw new RuntimeException(
+                        "The storage is full, you can replace an existing element!");
```
petro-ushchuk, line 25: Inline it
```suggestion
                throw new RuntimeException("The storage is full, you can replace an existing element!");
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_CAPACITY = 10;
+    private final KeyValuePair<K, V>[] array = new KeyValuePair[START_CAPACITY];
```
petro-ushchuk, line 7: Initialize `new KeyValuePair` in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                break;
+            }
+        }
+        if (!isExist) {
+            if (size == START_CAPACITY) {
+                throw new RuntimeException(
+                        "The storage is full, you can replace an existing element!");
+            }
+            array[size] = instanceOfKeyValue;
+            size++;
```
petro-ushchuk, line 28:
```suggestion
            array[size++] = instanceOfKeyValue;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     @Override
     public void put(K key, V value) {
+        boolean isExist = false;
+        KeyValuePair<K, V> instanceOfKeyValue = new KeyValuePair<>(key, value);
+        for (int i = 0; i < size; i++) {
+            if (array[i].getKey() == key || array[i].getKey() != null
+                    && array[i].getKey().equals(key)) {
+                array[i] = instanceOfKeyValue;
+                isExist = true;
+                break;
```
petro-ushchuk, line 19: if it exists you can return from here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            }
+        }
+        if (!isExist) {
```
petro-ushchuk, line 22: if you'll return from loop, this block and variable `isExist` are useless

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    public StorageImpl() {
+        this.array = (KeyValuePair<K, V>[]) new KeyValuePair[START_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        KeyValuePair<K, V> instanceOfKeyValue = new KeyValuePair<>(key, value);
+        for (int i = 0; i < size; i++) {
+            if (array[i].getKey() == key || array[i].getKey() != null
+                    && array[i].getKey().equals(key)) {
```
nacenik, line 19: [extract this check into the private method and reuse it. Cause you have 2 of the same block of code  ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-create-repeating-code)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_CAPACITY = 10;
+    private KeyValuePair<K, V>[] array;
```
nacenik, line 7: Make this variable final too. If we use the final keyword with an object, it means that the reference cannot be changed, but its state (instance variables) can be changed.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private int size = 0;
+
+    public StorageImpl() {
+        this.array = (KeyValuePair<K, V>[]) new KeyValuePair[START_CAPACITY];
```
nacenik, line 11: that's enough to do simple `new KeyValuePair<>[START_CAPACITY];` without casting

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_CAPACITY = 10;
+    private KeyValuePair<K, V>[] array;
+    private int size = 0;
```
nacenik, line 8: you don't need to initialize the int variable, cause class fields have default values (local variables are different - they have to be initialized)

nacenik (changes requested):

src/main/java/core/basesyntax/impl/KeyValuePair.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax.impl;
+
+public class KeyValuePair<K, V> {
```
karma-o, line 3: I would make it an inner static class, without the getters and setters because we could have been referencing the fields directly

karma-o (approved):
GJ, left some general comment

PR #955, opened 2023-02-15
Implemented the Storage of key-value type

Yehor-Cherkashyn:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX];
+        values = (V[]) new Object[MAX];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
potrap, line 19:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```
it looks prettier)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
```
potrap, line 27: i think you should throw RunntimeException to indicate that your storage is overflowed

potrap (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Yehor-Cherkashyn: I don't think it will affect the logic of the program in any way

src/main/java/core/basesyntax/impl/StorageImpl.java
Yehor-Cherkashyn: task specifies the maximum length of the array. if it were not specified in the task, then we could talk about throwing an exception

potrap (approved):

petro-ushchuk (approved):

PR #956, opened 2023-02-15
create class StorageImpl

MaksymHorielikov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (size == MAX_SIZE) {
```
IlkoGood, line 19: This is unnecessary check according to the task, but it completely falls in program logic, so let it be.

IlkoGood (approved):
Good job!

src/main/java/core/basesyntax/impl/StorageImpl.java
MaksymHorielikov: What do you suggest to do when trying to add a new entry to a full repository? The method is void and if you don't throw an exception, the user may think the method succeeded when it didn't.

IlkoGood: You are absolutely right, except the fact it's impossible in current task, as said in readme.
But you had sound plan and I approved it with no doubts!

MaksymHorielikov: The task description doesn't say what the put method should do if you try to add a new key with a new value, so it's up to the developer. Exceptions must be used when the program does not work correctly, and this is just our case)

petro-ushchuk (approved):

PR #958, opened 2023-02-15
implemented StotageImpl class

outref:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STARTING_CAPACITY = 10;
+    private K[] keys = (K[]) new Object[STARTING_CAPACITY];
```
NazarZva, line 7:
```suggestion
    private K[] keys;
```
we need do this in constuctor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    private V[] values = (V[]) new Object[STARTING_CAPACITY];
```
NazarZva, line 8:
```suggestion
    private V[] values;
```
the same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        if (keyIndex == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[keyIndex] = value;
+        }
+        if (keys.length == size) {
+            increaseCapacity();
+        }
```
NazarZva, line 23:
```suggestion
        if (keys.length == size) {
            increaseCapacity();
        }
```
we shouldn`t have this condition

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    //doubles arrays length
+    private void increaseCapacity() {
+        K[] newKeysArray = (K[]) new Object[keys.length * 2];
+        System.arraycopy(keys, 0, newKeysArray, 0, keys.length);
+        keys = newKeysArray;
+
+        V[] newValuesArray = (V[]) new Object[values.length * 2];
+        System.arraycopy(values, 0, newValuesArray, 0, values.length);
+        values = newValuesArray;
+    }
```
NazarZva, line 62:
```suggestion
    //doubles arrays length
    private void increaseCapacity() {
        K[] newKeysArray = (K[]) new Object[keys.length * 2];
        System.arraycopy(keys, 0, newKeysArray, 0, keys.length);
        keys = newKeysArray;

        V[] newValuesArray = (V[]) new Object[values.length * 2];
        System.arraycopy(values, 0, newValuesArray, 0, values.length);
        values = newValuesArray;
    }
```
we shouldn`t use this method, bacause we assume that the maximum number of elements in our Storage is 10

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        return size;
+    }
+
+    //returns -1 if no such key was found
+    private int getKeyIndex(K obj) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(obj)) {
+                return i;
+            }
+            if (keys[i] == null && obj == null) {
```
VitaliiPotseluiko, line 46: first of all. it wil be better) And then You shoud join line 43 and 46. Think what kind of operator you have to use)
```suggestion
            if (obj == keys[i]) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                return i;
+            }
```
NazarZva, line 48:
```suggestion
            if (keys[i] == obj || keys[i] != null && keys[i].equals(obj)) {
                return i;
            }
```
it will be better

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STARTING_CAPACITY = 10;
+    private K[] keys = (K[]) new Object[STARTING_CAPACITY];
+    private V[] values = (V[]) new Object[STARTING_CAPACITY];
+    private int size = 0;
```
VitaliiPotseluiko, line 9: There is no necessary to initialize this field. Because default value is 0 as well.

NazarZva (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
VitaliiPotseluiko, line 7: I my opinion fields keys and values you should initialize in constructor. Just my recommendation)

src/main/java/core/basesyntax/impl/StorageImpl.java
outref: Added it just in case. I think it doesnt't hurt to have it)

src/main/java/core/basesyntax/impl/StorageImpl.java
outref: Added this and all other changes. Approve if you agree.

src/main/java/core/basesyntax/impl/StorageImpl.java
outref: Added this and all other changes. Approve if you agree.

VitaliiPotseluiko (approved):

NazarZva (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STARTING_CAPACITY];
+        values = (V[]) new Object[STARTING_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        int keyIndex = getKeyIndex(key);
+        if (keyIndex == -1) {
```
maxline, line 19: -1 is magic number.  
Also you can think about creating a boolean method that check if values already present in array. And call the new method in all places that check if `keyIndex == -1`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    //returns -1 if no such key was found
```
maxline, line 45: please remove comments from the code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] != null && keys[i].equals(obj))
+                    || (keys[i] == obj)) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    //doubles arrays length
```
maxline, line 56: please remove too

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline: not fixed `-1` magic numbers

maxline (changes requested):
one occurence with magic value missed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+    }
+
+    private void increaseCapacity() {
+        K[] newKeysArray = (K[]) new Object[keys.length * 2];
+        System.arraycopy(keys, 0, newKeysArray, 0, keys.length);
+        keys = newKeysArray;
+
+        V[] newValuesArray = (V[]) new Object[values.length * 2];
+        System.arraycopy(values, 0, newValuesArray, 0, values.length);
+        values = newValuesArray;
```
nacenik, line 63: According to the task, the max size of the storage is 10 elements. Delete this method
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,63 @@
...
+        if (keyIndex == NOT_FOUND_INDEX) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[keyIndex] = value;
+        }
+        if (keys.length == size) {
+            increaseCapacity();
+        }
```
nacenik, line 29: This condition is useless

nacenik (changes requested):
Just delete the useless blocks of code, and the task will be approved 😉

maxline (approved):
Good job! Approved.

PR #959, opened 2023-02-15
Implemented StorageImpl class and its methods.

VitaliiPotseluiko:


outref (approved):
In my opinion everything looks perfect)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null) {
```
KseniiaMakarova, line 41:
```suggestion
            if (key == keys[i]) {
```
it is a nice way to compare nulls

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            size++;
+        } else {
+            for (int i = 0; i < size && size <= MAX_CAPACITY; i++) {
+                if (key == null && keys[i] == null) {
+                    values[i] = value;
+                    return;
+                } else if (keys[i] != null && keys[i].equals(key)) {
+                    values[i] = value;
+                    return;
+                }
```
KseniiaMakarova, line 30: same comments here as above

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        keys = new Object[MAX_CAPACITY];
+        values = new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == 0) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
KseniiaMakarova, line 21: do you need this check?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
VitaliiPotseluiko: Now I don't need this check because I saw my mistake, it wil be rest)))

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_CAPACITY];
+        values = new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size && size <= MAX_CAPACITY; i++) {
```
maxline, line 18: please extract the new method e.g. `contains(key)` and call it from `put()` and from `get()` methods to avoid code duplication.
also it is good idea create method `int indexOf(key)` to avoid duplication in finding index of the element.

maxline: `size` should not be greater as MAX_CAPACITY

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+            if (key == keys[i] || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
```
maxline, line 26: we don't need increment `size` if reach MAX_CAPACITY

maxline (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline: According to requirements of this task, checking MAX_CAPACITY is not necessary.  Ignore this comment.

PR #966, opened 2023-02-16
Hw-storage

KosMarch:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
```
EliSoll, line 26: As for me, incrementing size inside box-shaped brackets looks more structured with less coding, but I really do not insist on such changes (maybe it isn't the best practice)
```suggestion
        values[size++] = value;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[ARRAY_SIZE];
+        values = (V[]) new Object[ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
EliSoll, line 19: Good job! We can extract this statement inside if brackets to separate private method like "boolean checkKeys" or smth like that. Then you can avoid duplicating code here and in your get method. )

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
EliSoll, line 32: The same comment: you can avoid duplications here and in your put method by extracting checking to separate private method.

EliSoll (approved):
Good job! I've left some comments about code style, but these changes are not so crucial for me to reject overall your pull request.

src/main/java/core/basesyntax/impl/StorageImpl.java
Andriy-dz: I agree with the comment above

Andriy-dz, line 19:
```suggestion
            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            if (checkKeys(key, i)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
+    boolean checkKeys(K key, int i) {
+        return keys[i] == key || keys[i] != null && keys[i].equals(key);
+    }
```
Andriy-dz, line 41: you missed the access modifier and it's better to add parentheses, so it's more readable. don't forget to order methods by access modifiers

Andriy-dz (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     public V get(K key) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (checkKeys(key, i)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
+    private boolean checkKeys(K key, int i) {
```
KyslyiViacheslav, line 39: I sink , if you want to check keys, better to pass the key and keys[i] to method checkKeys(K key, K keys[i]), instead (K key, int i). Then code will be more readable.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[ARRAY_SIZE];
+        values = (V[]) new Object[ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (checkKeys(key, i)) {
```
KyslyiViacheslav, line 19: I sink , if you want to check keys, better to pass the key and keys[i] to method checkKeys(K key, K keys[i]), instead (K key, int i). Then code will be more readable.

KyslyiViacheslav (approved):

Nikitos787 (approved):
I think everything is good

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     public V get(K key) {
+        for (int i = 0; i < ARRAY_SIZE; i++) {
+            if (checkKeys(key, keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
 
+    private boolean checkKeys(K key, K keys) {
```
maxline, line 39: A little suggestion about naming. It would be more accurate to name method as `compareKeys()`. And second parameter keys should not be in plural, cause it reads like an array or list.

maxline (approved):
Good job. Approved with small suggestion about naming. You can take note for future.

PR #968, opened 2023-02-17
implemented Storage with service methods

EliSoll:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                values[i] = value;
+                return;
+            }
+        }
+        if (isStorageFull()) {
+            throw new RuntimeException("Can't add key: " + key + ", value: " + value + "."
+                    + " Storage is full");
+        }
+        keys[size] = key;
+        values[size++] = value;
```
denys-mg, line 29: probably it's better to write size increment operation in a new line, to improve code readability

denys-mg (approved):
Seems that everything is good)

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline: yes, `size++` in a new line will be more readable.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (isKeyIdentical(key, i)) {
```
maxline, line 19: The method name `isKeyIdentical()` implies that method compares two values that have the same type. It is looks not consistent enough when one method parameter is object and second is index. I'd suggest to move the logic with `for` to a separate method. And also avoid code duplication in `put()` and `get()` methods. E.g. 1) private `int indexOf(K key)` that returns index if element was found or `-1` if not , 2) method `contains(K key)` that just check if `indexOf(key) >= 0;`.

maxline (changes requested):
Good job! Let's improve the solution

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private boolean checkIndex(int index) {
+        return index != -1;
+    }
```
nacenik, line 52: According to the clean code: all private methods should be after all public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
-        return -1;
+        return size;
+    }
+
+    private boolean isStorageFull() {
+        return size == MAX_CAPACITY;
     }
 }
+
+
```
nacenik, line 64: Delete redundant empty lines

nacenik (approved):
Good job! But try to avoid mistakes that I mentioned

PR #970, opened 2023-02-17
implemented storage logic

denys-mg:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final StorageItem[] storageItems;
+    private int indexCounter = 0;
```
yurihhhT, line 8: good work)

yurihhhT (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o, line 8: * this is kind of obvious that it is a number :)
* primitive class fields have default values
```suggestion
    private static final int MAX_ITEMS = 10;
    private final Pair[] pairs;
    private int size;
```

karma-o (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
     public void put(K key, V value) {
+        Pair<K, V> existingStorageItem = getStorageItemByKey(key);
+        if (existingStorageItem != null) {
+            existingStorageItem.value = value;
+            return;
+        }
+        Pair<K, V> storageItem = new Pair<>(key, value);
+        if (size == pairs.length) {
+            throw new RuntimeException("Storage is FULL, you can't add new pair");
+        }
```
nacenik, line 24: make check on size before create new Pair
```suggestion
        if (size == pairs.length) {
            throw new RuntimeException("Storage is FULL, you can't add new pair");
        }
        Pair<K, V> storageItem = new Pair<>(key, value);
```

nacenik (approved):
Good job! 🔥

PR #971, opened 2023-02-17
implemented storage KeyValue-logic

ArtemGrunin:


jaroslavcyk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ROWS_NUMBER = 10;
+    private static final int MAX_COLUMN_NUMBER = 2;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private Object[][] keyValuePairs;
+    private int storageSize;
```
karma-o, line 11: if you wanted to do it as a pair, you should have created an inner class `Pair` with 2 fields: `key` and `value` of type `K` and `V` and here have it as `Pair[]` instead of `Object[][]` or just create 2 arrays of type `K` and `V` and have them separate here
\+ most of the constants will become redundant with those changes so leave only the 10 and name it correctly

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        }
+        if (isStorageFull()) {
+            throw new RuntimeException("Storage is full! Can't add key: "
+                    + key + ", value: " + value);
+        }
+        if (keyIsNew) {
+            keyValuePairs[storageSize][KEY_INDEX] = key;
+            keyValuePairs[storageSize][VALUE_INDEX] = value;
+            storageSize++;
+        }
```
karma-o, line 35:
```suggestion
                return;
            }
        }
        if (isStorageFull()) {
            throw new RuntimeException("Storage is full! Can't add key: "
                    + key + ", value: " + value);
        }
        keyValuePairs[storageSize][KEY_INDEX] = key;
        keyValuePairs[storageSize][VALUE_INDEX] = value;
        storageSize++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+    private boolean isStorageFull() {
+        return storageSize == MAX_ROWS_NUMBER;
+    }
+
+    private boolean compareObjects(Object obj1, Object obj2) {
+        return (obj1 == obj2) || obj1 != null && obj1.equals(obj2);
     }
 
     @Override
     public V get(K key) {
```
karma-o, line 47: public methods should be on top of the private ones

karma-o (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ArtemGrunin: ok) thanks for your comments) I seem to have refactored the code. With 2 arrays it turned out to be a little more informative and I did not need additional constants)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH = 10;
+    private K[] key;
+    private V[] value;
```
sarakhmen, line 8: It's better to rename these arrays to 'keys' and 'values' and this way you could avoid using 'this' reference in the further methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
-        return -1;
+        return storageSize;
+    }
+
+    private boolean isStorageFull() {
+        return storageSize == MAX_LENGTH;
+    }
+
+    private boolean compare(Object obj1, Object obj2) {
+        return (obj1 == obj2) || obj1 != null && obj1.equals(obj2);
```
sarakhmen, line 53: you can remove redundant parentheses 
```suggestion
        return obj1 == obj2 || obj1 != null && obj1.equals(obj2);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     public int size() {
-        return -1;
+        return storageSize;
+    }
+
+    private boolean isStorageFull() {
+        return storageSize == MAX_LENGTH;
+    }
+
+    private boolean compare(Object obj1, Object obj2) {
```
sarakhmen, line 52: this method is used to compare keys, so let's parametrize and rename your 'keys' 
```suggestion
    private boolean compare(K key1, K key2) {
```

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #972, opened 2023-02-17
Solved task

KyslyiViacheslav:
Implemented storage of key-value type;
Implemented methods put,get,size.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -4,13 +4,13 @@ import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     public static final int MAX_STORAGE_PLACE = 10;
-    private final Object[] value;
-    private final Object[] key;
+    private final V[] value;
+    private final K[] key;
     private int currentSize;
 
     public StorageImpl() {
-        key = new Object[MAX_STORAGE_PLACE];
-        value = new Object[MAX_STORAGE_PLACE];
+        key = (K[]) new Object[MAX_STORAGE_PLACE];
+        value = (V[]) new Object[MAX_STORAGE_PLACE];
     }
 
     @Override
@@ -30,7 +30,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
     public V get(K key) {
         for (int i = 0; i < currentSize; i++) {
             if (equal(key,this.key[i])) {
-                return (V) this.value[i];
+                return this.value[i];
             }
         }
         return null;
@@ -42,12 +42,6 @@ public class StorageImpl<K, V> implements Storage<K, V> {
     }
 
     private boolean equal(Object a, Object b) {
-        if (a == b) {
-            return true;
-        }
-        if (a == null || b == null || a.getClass() != b.getClass()) {
-            return false;
-        }
         return (a == b) || (a != null && a.equals(b));
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_STORAGE_PLACE = 10;
+    public static final int NUMBER_OF_ITEMS = 2;
+    private final Object[][] valueStorage;
```
Skrudj, line 8: rewrite the logic of storage. value and key may have different classes, but using the two-dimensional array you make that key and value must be the same class. create two arrays: one for values and another for keys. Others will be the same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private int currentSize;
+
+    public StorageImpl() {
+        valueStorage = new Object[MAX_STORAGE_PLACE][NUMBER_OF_ITEMS];
```
Skrudj, line 12: here you don't make values and keys as classes given as parametrs

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currentSize; i++) {
+            if (equal(key,valueStorage[i][0])) {
```
Skrudj, line 18: and also it will be better to use constants for indexes. index of key and index of value

Skrudj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: I have constant for **MAX_STORAGE_PLACE**, I think  for index we don't need constant, I have changeable field **currentSize**   in which you write current store size. Every time when we put new element, we must check our Input key with any  key in **valueStorage** and replace _value_ if _keys_ equal , for It we need unconstant index

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: I don't agree with you, my valueStorage should store any class, no matter what type of object you want to store.  
Also the task states what we **should solve this task using one or two arrays (this depends on your implementation)**

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: I don't agree with you, my valueStorage should store any class, no matter what type of object you want to store.  
Also the task states what we **should solve this task using one or two arrays (this depends on your implementation)**

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -4,32 +4,33 @@
...
-    public static final int NUMBER_OF_ITEMS = 2;
-    private final Object[][] valueStorage;
+    private final Object[] value;
+    private final Object[] key;
     private int currentSize;
 
     public StorageImpl() {
-        valueStorage = new Object[MAX_STORAGE_PLACE][NUMBER_OF_ITEMS];
+        key = new Object[MAX_STORAGE_PLACE];
+        value = new Object[MAX_STORAGE_PLACE];
```
Skrudj, line 13: you still can put whatever in this arrays. you should write 
key = <K[]> new Object[MAX_STORAGE_PLACE] and such for value

Skrudj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Skrudj: if it would store any class, why do we need generics?

src/main/java/core/basesyntax/impl/StorageImpl.java
Skrudj: i meant those where you write 0. 0 it's key, 1 it's value. valueStorage[i][KEY_INDEX] - what i meant

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: You don"t need do it. When you input data to storage you do it by method put, this method have only data wich you declarative when you create new StorageImpl. If you try input other data types, you will have compile error.

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: sorry. I didn't understand what you meant. it good idea

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: When you input data to storage, you do it by method put, this method has only data that you declare when you create new StorageImpl<K,V> .If you try input other data types, you will have compile error. If you open class ArrayList you will see that in ArrayList data store in simple _Object[] elementData_ without any typecasting.

Skrudj (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: @Skrudj is right, the point of generics is not only to restrict the put method types but also to store that values with the correct type, this is called **type safety**.
The fact that you have these fields private and accessed only through `put()` method does not mean much, we can easily set anything we want to that field with a reflection API or even just get this element with `get()` method, cast it to object and change the link to something completely different, regardless of the above written.
Regarding the original ArrayList, it is much more complex, has a lot more functionality and optimizations, and is not really viable for our task of just 2 methods which is more about the generics theory explained rather than the `List` or `Map` implementations(we will cover them later on),
So our array fields and parameters that represent them should to be parametrized

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    }
+
+    private boolean equal(Object a, Object b) {
+        if (a == b) {
+            return true;
+        }
+        if (a == null || b == null || a.getClass() != b.getClass()) {
+            return false;
+        }
+        return (a == b) || (a != null && a.equals(b));
```
karma-o, line 51:
```suggestion
        return a == b || a != null && a.equals(b);
```

karma-o (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: 1 array was about implementing a `Pair` class with `key` and `value` fields

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: +

src/main/java/core/basesyntax/impl/StorageImpl.java
KyslyiViacheslav: +

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
+    public StorageImpl() {
+        key = (K[]) new Object[MAX_STORAGE_PLACE];
+        value = (V[]) new Object[MAX_STORAGE_PLACE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currentSize; i++) {
+            if (equal(key,this.key[i])) {
```
Andriy-dz, line 19:
```suggestion
            if (equal(key, this.key[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        }
+        this.key[currentSize] = key;
+        this.value[currentSize] = value;
+        currentSize++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentSize; i++) {
+            if (equal(key,this.key[i])) {
```
Andriy-dz, line 32:
```suggestion
            if (equal(key, this.key[i])) {
```

Andriy-dz (approved):

PR #973, opened 2023-02-17
Storage Generic pull request

jaroslavcyk:
Implement class StorageImpl create two arrays with max size 10. Implement method put() which add new pair key-value and replace value if array have this key. Implement method get() which get value by a key.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,41 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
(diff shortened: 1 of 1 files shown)


PlagueEbola (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```
Andriy-dz, line 32: it is better to put this code in a separate method, because you are using it twice

Andriy-dz (changes requested):

petro-ushchuk (changes requested):
Good work, but you should fix your issues with build before asking review.

image

petro-ushchuk (approved):
Good work! 🔥

PR #974, opened 2023-02-17
created fields to Storage class which contain keys, values and size o…

Skrudj:
…f storage, implemented logic of putting, getting and rewriting data to storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,41 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_ARRAYS_LENGTH = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[STORAGE_ARRAYS_LENGTH];
+        this.values = (V[]) new Object[STORAGE_ARRAYS_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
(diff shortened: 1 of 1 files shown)


NadiaHuryk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    public StorageImpl() {
+        this.keys = (K[]) new Object[STORAGE_ARRAYS_LENGTH];
+        this.values = (V[]) new Object[STORAGE_ARRAYS_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
+                values[i] = value;
```
maxline, line 20: Code duplication in `put()` and `get()` methods. I'd suggest to extract the logic with for to a separate method.  E.g. 1) new method  `private int indexOf(K key)` that returns index if element was found or `-1` if not , 2) method `contains(K key)` that just check if `indexOf(key) >= 0;`. Do not forget add constants for magic values.

maxline (changes requested):
Good job! let's improve the solution.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
+                return i;
+            }
+        }
+        return NOT_FOUND_INDEX;
+    }
+
+    private boolean contains(K key) {
+        return indexOf(key) >= 0;
```
nacenik, line 49: According to the clean code: all private methods should be after all public methods

nacenik (approved):
Good job! Try to avoid mentioned mistake in the future

PR #975, opened 2023-02-18
Added storage implementation.

De1eF:
Maybe there is another way to do a null check on cells in put & get. I am ready for critiques)

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -2,18 +2,48 @@ package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
 
+import java.util.ArrayList;
+import java.util.List;
+
 public class StorageImpl<K, V> implements Storage<K, V> {
+    List<K> keyList = new ArrayList<>();
+    List<V> valueList = new ArrayList<>();
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyList.size(); i++) {
+            if (key == null && keyList.get(i) == null) {
+                valueList.set(i, value);
+                return;
+            }
+            if (keyList.get(i) != null && keyList.get(i).equals(key)) {
+                valueList.set(i, value);
+                return;
+            }
+        }
+        addCell(key, value);
+    }
+
+    private void addCell(K key, V value) {
+        keyList.add(key);
+        valueList.add(value);
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyList.size(); i++) {
+            if (key == null && keyList.get(i) == null) {
+                return valueList.get(i);
+            }
+            if (keyList.get(i) != null && keyList.get(i).equals(key)) {
+                return valueList.get(i);
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return keyList.size();
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,48 +2,68 @@
...
     public void put(K key, V value) {
-        for (int i = 0; i < keyList.size(); i++) {
-            if (key == null && keyList.get(i) == null) {
-                valueList.set(i, value);
-                return;
-            }
-            if (keyList.get(i) != null && keyList.get(i).equals(key)) {
-                valueList.set(i, value);
-                return;
+        if (keyList != null) {
```
sanya-1xx8, line 11: if ((key == keyList[i]) || key != null && key.equals(keyList[i])) { valueList[i] = value;
                    return; }

sanya-1xx8: this check is enough

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,48 +2,68 @@
...
+        System.arraycopy(original, 0, newArr, 0, original.length);
+        newArr[newArr.length - 1] = toAdd;
+        return newArr;
+    }
+
+    private V[] appendValuesArray(V[] original, V toAdd) {
+        V[] newArr = (V[])new Object[original.length + 1];
+        System.arraycopy(original, 0, newArr, 0, original.length);
+        newArr[newArr.length - 1] = toAdd;
+        return newArr;
```
sanya-1xx8, line 49: since the size of the array in the condition is 10, it is not necessary to create a new array every time. It is better to leave this question for the mentor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,48 +2,68 @@
...
             }
         }
         addCell(key, value);
     }
 
     private void addCell(K key, V value) {
-        keyList.add(key);
-        valueList.add(value);
+        if (keyList == null) {
+            keyList = (K[])new Object[1];
```
sanya-1xx8, line 28: Your code should be easy to read. Let’s move all hardcoded values to constant fields and give them informative names.

src/main/java/core/basesyntax/impl/StorageImpl.java
sanya-1xx8: we can create two arrays of size 10 in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,6 +2,8 @@
 
 import core.basesyntax.Storage;
 
+import java.util.Objects;
```
sanya-1xx8, line 5: we cannot use the class Objects in this task. Therefore,Maven check does not pass.
Try to leave null check as I wrote below

sanya-1xx8 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
sanya-1xx8: if ((key == keyList[i]) || key != null && key.equals(keyList[i])) { valueList[i] = value;
return; }

sanya-1xx8 (approved):
Good job)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keyList;
+    private V[] valueList;
```
karma-o, line 7: * initialize these fields in the constructor and remove all of the redundant arrays null checks after this change
*  add a `size` variable which will represent the number of elements that we have put into our storage \+ a default max capacity constant
* this is not a `List` so let's not name it like it is
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (keyList != null) {
+            for (int i = 0; i < keyList.length; i++) {
+                if ((key == keyList[i])
+                        || key != null && key.equals(keyList[i])) {
```
karma-o, line 14: redundant braces

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keyList;
+    private V[] valueList;
+
     @Override
     public void put(K key, V value) {
+        if (keyList != null) {
+            for (int i = 0; i < keyList.length; i++) {
```
karma-o, line 12: iterate up to the `size`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+        newArr[newArr.length - 1] = toAdd;
+        return newArr;
+    }
+
+    private V[] appendValuesArray(V[] original, V toAdd) {
+        V[] newArr = (V[])new Object[original.length + 1];
+        System.arraycopy(original, 0, newArr, 0, original.length);
+        newArr[newArr.length - 1] = toAdd;
+        return newArr;
     }
```
karma-o, line 47: all of this can be replaced with:
```
keys[size] = key;
values[size] = value;
size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyList.length; i++) {
+            if ((key == keyList[i])
+                    || key != null && key.equals(keyList[i])) {
```
karma-o, line 53: also redundant braces

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+    private V[] appendValuesArray(V[] original, V toAdd) {
+        V[] newArr = (V[])new Object[original.length + 1];
+        System.arraycopy(original, 0, newArr, 0, original.length);
+        newArr[newArr.length - 1] = toAdd;
+        return newArr;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyList.length; i++) {
```
karma-o, line 51: loop only up to the size

karma-o (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys = (K[]) new Object[10];
```
sarakhmen, line 6: Let's introduce a constant for this magic number

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+        values[size] = value;
+        size++;
+        //addCell(key, value);
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == this.keys[i]
+                    || key != null && key.equals(this.keys[i])) {
```
sarakhmen, line 31: move this comparison to the separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys = (K[]) new Object[10];
+    private V[] values = (V[]) new Object[10];
```
sarakhmen, line 7: also, let's initialize this arrays within the constructora

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        if (this.keys != null) {
+            for (int i = 0; i < size; i++) {
+                if (key == this.keys[i]
+                        || key != null && key.equals(this.keys[i])) {
+                    this.values[i] = value;
```
sarakhmen, line 16: 'this' referencing is redundant

sarakhmen (changes requested):
Good job, let's improve your solution :)

petro-ushchuk (approved):
Good work 🔥

PR #976, opened 2023-02-18
implementation of storage task

Nikitos787:


KyrylH (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
+        this.values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (compare(key, i)) {
```
maxline, line 19: If method name is `compare `it implies that method compares two values that have one type.  It is looks not consistent enough when one method parameter is object and second is index. I'd suggest to move the logic with for to a separate method. And also avoid code duplication in `put()` and `get()` methods. E.g.  1) `private int indexOf(K key) ` that returns index if element was found or -1 if not , 2) method `contains(K key)` that just check  if ` indexOf(key) >= 0; `.

maxline (changes requested):
Good job! a little suggestion was added.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int NOT_EXIST_INDEX = -1;
```
sarakhmen, line 7:
```suggestion
    private static final int NOT_EXISTS_INDEX = -1;
```

sarakhmen (approved):
Good job :)

PR #977, opened 2023-02-18
implemented class StorageImpl

PlagueEbola:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_STORAGE = 10;
+    private final K[] key;
+    private final V[] value;
```
redmanO-o, line 8: better to give different names to the arrays than the names of the input data then you will be able to remove unnecessary "this." from the code.
```suggestion
    private final V[] differentValue;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+
+    public StorageImpl() {
+        key = (K[]) new Object[MAX_SIZE_STORAGE];
+        value = (V[]) new Object[MAX_SIZE_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((this.key[i] == key)
```
redmanO-o, line 19:
```suggestion
            if ((key[i] == key)
```
and you can remove other "this." in your code. This will improve readability

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+        this.key[size] = key;
+        this.value[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((this.key[i] == key)
+                    || (this.key[i] != null && this.key[i].equals(key))) {
```
redmanO-o, line 34: better to add a separate method for comparing keys

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    public StorageImpl() {
+        key = (K[]) new Object[MAX_SIZE_STORAGE];
+        value = (V[]) new Object[MAX_SIZE_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((this.key[i] == key)
+                    || (this.key[i] != null && this.key[i].equals(key))) {
```
redmanO-o, line 20: and here.
better to add a separate method for comparing keys

redmanO-o (changes requested):

redmanO-o (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE_STORAGE];
+        values = (V[]) new Object[MAX_SIZE_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (compare(keys[i], key)) {
+                values[i] = value;
```
maxline, line 20: The method name `compare()` implies that method compares two values that have the same type. It is looks not consistent enough when one method parameter is object and second is index. I'd suggest to move the logic with `for` to a separate method. And also avoid code duplication in `put()` and `get()` methods. E.g. 1) new method `private int indexOf(K key)` that returns index if element was found or `-1` if not , 2) method `boolean contains(K key)` that just check if `indexOf(key) >= 0;`.

maxline (changes requested):
good job! let's improve the solution.

sarakhmen (approved):
Good job :)

PR #981, opened 2023-02-19
implemented logic in StorageImpl class

OlehSarapuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_ELEMENTS = 10;
+    private Object[] keys;
+    private Object[] values;
```
pechivo, line 8:
```suggestion
    private Object[] keys;
    private Object[] values;
```
You can here already make your arrays type K and V  and not casting then the result of get method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    public StorageImpl() {
+        keys = new Object[MAX_STORAGE_ELEMENTS];
+        values = new Object[MAX_STORAGE_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        storageContainsKey = false;
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```
pechivo, line 20: > if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) 

this part of the code repeat twice, so it's better to create a method for it like compare method and call it in necessary methods

pechivo, line 20: this part of the code repeat twice, so it's better to create a method for it like compare method and call it in necessary methods

pechivo (changes requested):

OlehSarapuk:
thanks for your remarks
i've tried to implement them and i refactored my code  a little bit)

pechivo (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            if (isEqual(key, keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
 
+    private boolean isEqual(K key1, K key2) {
+        return (key1 == key2) || (key1 != null && key1.equals(key2));
+    }
```
nacenik, line 42: According to the clean code: all private methods should be after public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_ELEMENTS = 10;
```
```suggestion
    private static final int MAX_STORAGE_SIZE = 10;
```

nacenik (changes requested):
Almost done, just fix little mistakes 😉

OlehSarapuk:
thanks)
all done

PR #989, opened 2023-02-19
Implement the Storage

copypasterr:
…o types of data

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,46 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private K[] keys;
+    private V[] values;
+    private int space;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+    }
+
+    public boolean verify(K k, K v) {
+        return (k != null && k.equals(v) || k == v);
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < space; i++) {
+            if (verify(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[space] = key;
+        values[space] = value;
+        space++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < space; i++) {
+            if (verify(keys[i], key)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return space;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < space; i++) {
+            if (verify(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[space] = key;
+        values[space] = value;
+        space++;
```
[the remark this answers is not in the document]
nikolaichuk-artem: better look

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private K[] keys;
+    private V[] values;
```
nikolaichuk-artem, line 9: I think we should make this variable final 
If we use the final keyword with an object, it means that the reference cannot be changed, but its state (instance variables) can be changed.

src/main/java/core/basesyntax/impl/StorageImpl.java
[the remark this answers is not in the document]
copypasterr: thanks, I'll fix it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private final K[] keys;
+    private final V[] values;
+    private int space;
```
sarakhmen, line 11: let's remove it to 'size' or 'index'

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) Array.newInstance(Object.class, STORAGE_SIZE);
+        values = (V[]) Array.newInstance(Object.class, STORAGE_SIZE);
+    }
+
+    public boolean verify(K k, K v) {
+        return (k != null && k.equals(v) || k == v);
+    }
```
sarakhmen, line 20:
```suggestion
    private boolean areKeysEqual(K key1, K key2) {
        return key1 == key2 || key1 != null && key1.equals(key2);
    }
```
Let's declare this method after public ones.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
```
sarakhmen, line 4: remove this import and initialize arrays as I commented below

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+    }
+
+    private boolean areKeysEqual(K key1, K key2) {
+        return key1 == key2 || key1 != null && key1.equals(key2);
+    }
```
nacenik, line 18: According to clean code: all private method should be after all public methods

PR #990, opened 2023-02-19
Storage of keys and values 

s-hrynus:


Hlib13 (commented):
all on top

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean comparator(K key1, K key2) {
```
karma-o, line 44: There is the actual `Comparator` in JDK, which we will learn later on, but for now, let's rename this, do not forget about the methods naming convention
```suggestion
    private boolean compare(K key1, K key2) {
```

karma-o (approved):
Good job!

PR #993, opened 2023-02-19
implemented StorageImpl class (put, get and size methods)

AleksandrLts:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_DATA_LENGTH = 10;
+    private final K[] keys = (K[]) new Object[MAX_DATA_LENGTH];
+    private final V[] values = (V[]) new Object[MAX_DATA_LENGTH];
```
V-Vasylchuk, line 8: I think it's a bad practice to initialize variables right away, it's better to do it in a separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
...
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```
V-Vasylchuk, line 14: first  time comparable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,36 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```
V-Vasylchuk, line 27: second time comparable. When you do the same things more than one time, much better to have a method for that like:
`public boolean compareKeys(K firstKey, K secondKey) {
return /*comparing*/;
}`

V-Vasylchuk (changes requested):

V-Vasylchuk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_DATA_LENGTH = 10;
```
```suggestion
    private static final int MAX_ARRAY_SIZE = 10;
```
OR
```suggestion
    private static final int DEFAULT_SIZE = 10;
```

nacenik (approved):
Perfect! 🔥

PR #1003, opened 2023-02-20
add decision storage 1.0

vfedychkanych:


README.md
```diff
@@ -5,14 +5,6 @@ You should be able to put new key-value pair and to get value by key.
...
-```java
-Storage<Integer, Box> storage = new StorageImpl<>(); 
-Box box = new Box();
-storage.put(22, box); 
-Box value = storage.get(22); // returns the Box
-int size = storage.size(); // returns storage size
-```
 You should solve this task using one or two arrays (this depends on your implementation). 
 If you use arrays for your solution, we assume that the maximum number of elements in our `Storage` is 10.
 #### [Try to avoid these common mistakes while solving task](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html)
```
petro-ushchuk, line 10: Remove this file from your commit

petro-ushchuk (approved):
Good work, but don't commit external files.

PR #1006, opened 2023-02-20
 BUILD SUCCESS

yuriy49:
create private fields
 add constructor and cast to V/K[] type objects
 implement methods below

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,56 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
     }
 
-    @Override
+    public void put(K key, V value) {
+        int index = findIndex(key);
+        if (index >= 0) {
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
-    @Override
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
Nimakel, line 15: Your variable `size` alredy 0 by default, when you initialize this variable in class field. You can remove that

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+    }
+
+    private int findIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (notNull(keys[i],key)) {
+                return i;
+            }
+        }
         return -1;
     }
```
Nimakel, line 51: You can simplify this by using `for()` loop in `put()` method without using Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+            size++;
+        }
     }
 
-    @Override
     public V get(K key) {
+        int index = findIndex(key);
+        if (index >= 0) {
+            return values[index];
+        }
```
Nimakel, line 36: Also can simplify this by using `for()` loop without using Objects

Nimakel (changes requested):

Nimakel (approved):
Great work!

nacenik (approved):
Perfect! 🔥

PR #1011, opened 2023-02-28
Initial commit with solution

ArtemFedoseev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int STARTING_SIZE = 0;
```
kozhukhovsky, line 7: Redundant variable. By default it will be 0 anyway

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+    private final K[] keys = (K[]) new Object[MAX_SIZE];
+    private final V[] values = (V[]) new Object[MAX_SIZE];
+    private int currentSize = STARTING_SIZE;
```
kozhukhovsky, line 10:
```suggestion
    private int currentSize;
```

kozhukhovsky (changes requested):
You could make it a little simpler, but it's an interesting solution, and it's a good idea. Well done!

kozhukhovsky (approved):

PR #1015, opened 2023-03-04
first commit

Rolik-Oleksandr:
implement solution

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,52 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private int size;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (findKeyIndex(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (findKeyIndex(keys[i], key)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean findKeyIndex(Object firstKey, Object secondKey) {
```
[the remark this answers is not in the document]
uliana-andrushchuk: Don't use `Object` for method arguments.

uliana-andrushchuk (changes requested):

Rolik-Oleksandr:
fix all issues

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean areKeysEquals(K firstKey, V secondKey) {
```
olekskov, line 45: this method seems to be for internal use, please correct access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        if (firstKey == null && secondKey == null) {
+            return true;
+        }
+        if (firstKey == null || secondKey == null) {
+            return false;
+        }
+        return firstKey.equals(secondKey);
```
olekskov, line 52: There is common approach for equals check that is based on fact that null==null:
```suggestion
        return (firstKey == secondKey || firstKey != null && firstKey.equals(secondKey));
```

olekskov (changes requested):

Rolik-Oleksandr:
fixed corrections

olekskov (changes requested):
Please recheck comment regarding method access modifier, we shouldn't make accessible logic designed for internal use

Rolik-Oleksandr:
fix method access modifier

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private int size;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
okuzan, line 14:
```suggestion
```
0 is default for int

okuzan (approved):
Good job

PR #1020, opened 2023-03-08
Implemented the system of storage that have two parameters key and value

vromanyuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_INDEX = 0;
+    private K key;
+    private V value;
+    private StorageImpl<K, V>[] arrayElements;
+    private int size;
+
+    public StorageImpl() {
+        arrayElements = new StorageImpl[10];
```
kozhukhovsky, line 13: move 10 to the contsnat, like MAX_SIZE

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_INDEX = 0;
+    private K key;
+    private V value;
+    private StorageImpl<K, V>[] arrayElements;
```
```suggestion
    private K[] keys;
    private V[] values;
```

You don't need to store the object itself. It is enough to make an array of keys and values.

src/main/java/core/basesyntax/impl/StorageImpl.java
kozhukhovsky, line 13: init keys and values here

kozhukhovsky, line 13:
```suggestion

```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+            if (index != -1) {
+                arrayElements[index].setValue(value);
+            } else {
+                arrayElements[size] = new StorageImpl<>(key, value);
+                size++;
+            }
+        }
+    }
+
+    private int findIndex(K element) {
```
kozhukhovsky, line 50: private methods should be below all public ones

kozhukhovsky (changes requested):
Need to change the logic to an array of keys and values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_INDEX = 0;
+    private static final int MAX_SIZE = 10;
+    private final K[] arrayKey;
+    private final V[] arrayValue;
```
uliana-andrushchuk, line 9: You can simplify variable names to `keys`, `values`

src/main/java/core/basesyntax/impl/StorageImpl.java
uliana-andrushchuk: not fixed. please, move this method below all public methods

uliana-andrushchuk (changes requested):

olekskov (approved):
Good job!

PR #1021, opened 2023-03-08
HW-jv-storage-solved

OlStepp:


olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private int size = 0;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_AMOUNT];
+        this.values = (V[]) new Object[MAX_AMOUNT];
+    }
+
+    public boolean keyEquals(K keyFirst, K keySecond) {
```
okuzan, line 16: this method should be private, and therefore go after all the public ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_AMOUNT = 10;
+    private int size = 0;
```
okuzan, line 7: redundant initialization, 0 is default for int

okuzan (changes requested):
Good job so far!

aaananas (approved):

PR #1023, opened 2023-03-12
solution of the task

Anestheziv:


olekskov (approved):
Nice job! Approved. But comment in line 31 is unnecessary :)

PR #1024, opened 2023-03-14
completed task

BakhmetIvan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        }
+        keysArr[lastArrayIndex] = key;
+        valuesArr[lastArrayIndex] = value;
+        lastArrayIndex++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < lastArrayIndex; i++) {
+            if (keysArr[i] == key || keysArr[i] != null && keysArr[i].equals(key)) {
```
uliana-andrushchuk, line 35: You have almost the same logic here and in `put` method. try to extract it to separate method to avoid code duplication

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return lastArrayIndex;
+        //кількість елементів у масиві
```
uliana-andrushchuk, line 47: Please, remove all comments

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
BakhmetIvan: I'm a little confused about how to make it a separate method, because they have differences.

BakhmetIvan: in the first case we just exit the method and in the second case we return the generic value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private Object[] keysArr;
+    private Object[] valuesArr;
+    private int lastArrayIndex;
+
+    public StorageImpl() {
+        keysArr = new Object[MAX_ARRAY_LENGTH];
+        valuesArr = new Object[MAX_ARRAY_LENGTH];
+        lastArrayIndex = 0;
```
Ivan95kos, line 14: int has default value 0
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        int index = getIndex(key);
+        if (index == -1) {
+            return null;
+        } else {
+            V searchResult = (V) valuesArr[index];
+            return searchResult;
+        }
+    }
+
+    private int getIndex(K key) {
```
Ivan95kos, line 40: All private methods should be below public methods

Ivan95kos (changes requested):

kozhukhovsky (approved):

PR #1026, opened 2023-03-15
Ready for check

AnatolyNz:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < arrayKey.length; i++) {
+            if (key == arrayKey[i] || key != null && key.equals(arrayKey[i])) {
+                if (key == null && arrayKey[i] == null && arrayValue[i] == null) {
+                    size++;
+                }
+                arrayValue[i] = value;
+                break;
```
kozhukhovsky, line 24:
```suggestion
                return;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+            if (arrayKey[i] == null && arrayValue[i] == null) {
+                arrayKey[i] = key;
+                arrayValue[i] = value;
+                size++;
+                break;
+            }
```
kozhukhovsky, line 31:
```suggestion
                arrayKey[size] = key;
                arrayValue[size] = value;
                size++;
           
```

kozhukhovsky: and move this out of loop

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+        }
+        arrayKey[size] = key;
+        arrayValue[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < arrayKey.length; i++) {
+            if (key == arrayKey[i] || key != null && key.equals(arrayKey[i])) {
```
okuzan, line 35: you have the same logic in get and put, extract it to an external method

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    private int size;
+    private K[] arrayKey;
+    private V[] arrayValue;
+
+    public StorageImpl() {
+        arrayKey = (K[]) new Object[MAX_SIZE];
+        arrayValue = (V[]) new Object[MAX_SIZE];
+    }
+
+    private int findValue(K key) {
```
Ivan95kos, line 16: For the future, All private methods should be below public methods

Ivan95kos (approved):

PR #1029, opened 2023-03-17
jv-storage

VolodymyrDavydov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int SIZE_OF_STORAGE = 10;
+    private final K[] keysOfStorage;
+    private final V[] valuesOfStorage;
+    private int sizeOfStorage;
```
Ivan95kos, line 9: Variables should not have a class name
```suggestion
    private final K[] keys;
    private final V[] values;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+
+    public StorageImpl() {
+        this.keysOfStorage = (K[]) new Object[SIZE_OF_STORAGE];
+        this.valuesOfStorage = (V[]) new Object[SIZE_OF_STORAGE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < sizeOfStorage;i++) {
+            if (key == keysOfStorage[i] || key != null && key.equals(keysOfStorage[i])) {
```
Ivan95kos, line 19: You can move this check into a private method, you use it twice

Ivan95kos (approved):

PR #1034, opened 2023-03-27
implementation a Key-Value Store that is parameterized with two data …

TemkaHillfinger:
…types.

Added method implementation put,get

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,41 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_ARRAY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE_ARRAY];
+        this.values = (V[]) new Object[MAX_SIZE_ARRAY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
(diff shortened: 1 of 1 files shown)


okuzan (changes requested):
Good job, but you have the same logic in get and put, which can be extracted to an external method (iterating from 0 to size, comparing by equals). This method for example can return index of a match, or -1 if match not found

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[SIZE];
+        this.values = (V[]) new Object[SIZE];
+    }
+
+    public int getIndex(K key) {
```
okuzan, line 16: should be private and placed after the public ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
     }
 
     @Override
     public V get(K key) {
+        int index = getIndex(key);
+        if (index != -1) {
+            return values[index];
+        }
         return null;
```
```suggestion
        int index = getIndex(key);
        return index != -1 ? values[index] : null;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            values[index] = value;
+        } else {
+            if (size == SIZE) {
+                throw new RuntimeException("Array is full");
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+
```
okuzan, line 38: delete the line

okuzan (changes requested):

Ivan95kos (approved):
Great job

PR #1037, opened 2023-03-30
Storage solution

z1row:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keyStorage[i], key)) {
```
uliana-andrushchuk, line 44: Please, don’t use class Objects 
see [common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-use-class-objects)

uliana-andrushchuk (changes requested):

PR #1038, opened 2023-04-02
Work

Volodymyr-Romanyshyn:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private int size = 0;
+    private K[] keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+    private V[] values = (V[]) new Object[MAX_ARRAY_LENGTH];
```
uliana-andrushchuk, line 9: Please, initialize these fields in the constructor
[Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
     @Override
     public int size() {
-        return -1;
+        int valueCounter = 0;
+        for (V value : values) {
+            if (value != null) {
+                valueCounter++;
+            }
+        }
+        return valueCounter;
```
uliana-andrushchuk, line 42: You can simply return `size`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private int size = 0;
+    private K[] keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+    private V[] values = (V[]) new Object[MAX_ARRAY_LENGTH];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null || (key != null && key.equals(keys[i]))) {
```
uliana-andrushchuk, line 14: Please, extract this condition to separate method

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private int size;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        size = 0;
```
kozhukhovsky, line 12: 0 is default value of integer

kozhukhovsky, line 12:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+        values = (V[]) new Object[MAX_ARRAY_LENGTH];
+    }
+
+    private boolean isKeysEqual(K key, K key1) {
```
kozhukhovsky, line 17: private methods should come after public ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        return (key == null && key1 == null || (key != null && key.equals(key1)));
```
kozhukhovsky, line 18:
```suggestion
        return (key == null || (key != null && key.equals(key1)));
```

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Volodymyr-Romanyshyn: if we don't add this expression "&& key1 == null" we have failure with two keys with "null"

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (keys[i]) != null && keys[i].equals(key)) {
```
olekskov, line 32:
```suggestion
            if (isKeysEqual(key, keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean isKeysEqual(K key, K key1) {
+        return (key == null && key1 == null || (key != null && key.equals(key1)));
```
olekskov, line 45:
```suggestion
        return (key == key1 || (key != null && key.equals(key1)));
```

olekskov (changes requested):
I believe that Anton meant that for our comparison
key == null && key1 == null
will give us same result as
key == key1
so we can use shorter form and thus use isKeysEqual() method in both conditional statements are they are pursuing same purpose

olekskov (approved):
Good job, approved.
For further tasks you may use shorter form as below:

PR #1039, opened 2023-04-02
home-work/jv-storage

andrii-subash:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
uliana-andrushchuk, line 49: Don’t use class Objects
see [common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor)

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
+        this.values = (V[]) new Object[MAX_SIZE];
+        this.size = 0;
```
olekskov, line 14: 0 is default int value, there is no need to explicitly set it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size == MAX_SIZE) {
```
olekskov, line 23: I'd recommend to use >= in such cases despite it's impossible now to go over storage size, just to avoid possible problems in case more and more methods will be created if future

olekskov (approved):
Good job, approved, please pay attention to comments below for further tasks

PR #1040, opened 2023-04-03
Storage implemented.

TymurProkhorov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
...
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int currentSize;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 11:
```suggestion
    private Object[] keys;
    private Object[] values;
```
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
...
+        values[currentSize] = value;
+        currentSize++;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V value = null;
+        for (int i = 0; i < currentSize; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 36: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
...
+        result = 31 * result + Arrays.hashCode(keys);
+        result = 31 * result + Arrays.hashCode(values);
+        return result;
+    }
+
+    private void sizeCheck() {
+        if (currentSize <= MAX_SIZE) {
+            return;
+        }
+        throw new ArrayIndexOutOfBoundsException("Error. Storage is full.");
```
okuzan, line 74: print currentSize too

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int currentSize;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
...
+                && Arrays.equals(values, storage.values);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(currentSize);
+        result = 31 * result + Arrays.hashCode(keys);
+        result = 31 * result + Arrays.hashCode(values);
+        return result;
+    }
```
okuzan, line 68: this is not necessary

okuzan (changes requested):

okuzan (changes requested):
You have repeated logic in get and put methods, you need to extract it to an external method. For example, you can implement it like that: method takes a key as an input and iterates from 0 to size till the match is found, and returns the index of a match or a -1, if the match is not found

uliana-andrushchuk (approved):

PR #1042, opened 2023-04-06
Solution

MTM1963:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 16: 0 is default for int, it's redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 22: image

okuzan (changes requested):
Replace objects.equals with your own logic. You should extract that repeated logic (iterating in a for loop and comparing) to an external method

src/main/java/core/basesyntax/impl/StorageImpl.java
MTM1963: Done.

src/main/java/core/basesyntax/impl/StorageImpl.java
MTM1963: Removed

uliana-andrushchuk (approved):

PR #1043, opened 2023-04-08
storage solution

ArtyomK75:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            }
+            if (storage[i].getKey().equals(key)) {
+                storage[i].setValue(value);
+                return;
+            }
+        }
+        Pair<K, V>[] localStorage = new Pair[lengthStorage + 1];
+        for (int i = 0; i < lengthStorage; i++) {
+            localStorage[i] = storage[i];
+        }
```
olekskov, line 30: 1. I don't think we will benefit from growing array, especially consider that it never becomes shorter.
2. I can't see logic that will represent requirement "we assume that the maximum number of elements in our Storage is 10."
Please instantiate array with length required by task, simplify overall solution accordingly and add logic for situation when storage is full

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    public StorageImpl() {
+        storage = new Pair[0];
+    }
+
     @Override
     public void put(K key, V value) {
+        int lengthStorage = storage.length;
+        for (int i = 0; i < lengthStorage; i++) {
+            if (storage[i].getKey() == null && key == null) {
```
olekskov, line 16: if they are both null then == will return true
```suggestion
            if (storage[i].getKey() == key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     public V get(K key) {
+        for (int i = 0; i < storage.length; i++) {
+            if (storage[i].getKey() == null && key == null) {
+                return storage[i].getValue();
+            } else if (storage[i].getKey() == null || key == null) {
+                continue;
+            }
+            if (storage[i].getKey().equals(key)) {
+                return storage[i].getValue();
+            }
```
olekskov, line 45: based on comment above this and previous similar construction can be much more simple:
```suggestion
            if (storage[i].getKey() == key || storage[i].getKey() != null && storage[i].getKey().equals(key)) {
                return storage[i].getValue();
            } 
```

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Pair<K, V>[] storage;
+    private int indexFreePosition;
+
+    public StorageImpl() {
+        storage = new Pair[10];
```
olekskov, line 10: magic numbers should be class constants with proper naming

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < indexFreePosition; i++) {
+            if (storage[i].getKey() == key
+                    || storage[i].getKey() != null
+                    && storage[i].getKey().equals(key)) {
+                storage[i].setValue(value);
+                return;
+            }
+            if (storage[i] == null) {
```
olekskov, line 23: it doesn't seem to be possible situation, you called storage[i].getKey() in previous if() block so if storage[i] is null you already have NPE

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                break;
+            }
+        }
+
+        storage[indexFreePosition] = new Pair<>(key, value);
```
olekskov, line 28: Please throw some relevant exception when user tries to put something to storage that is already full, because now you just replace last position in the storage what is not correct

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            }
+        }
+        try {
+            storage[indexFreePosition] = new Pair<>(key, value);
+        } catch (IndexOutOfBoundsException e) {
+            throw new RuntimeException("In array not enough"
+                    + " space for this operation", e);
+        }
+        indexFreePosition++;
     }
```
olekskov, line 33:
```suggestion
    @Override
    public void put(K key, V value) {
        if (indexFreePosition >= storage.length) {
            throw new RuntimeException("Storage is full");
        }
        for (int i = 0; i < indexFreePosition; i++) {
            if (storage[i].getKey() == key
                    || storage[i].getKey() != null
                    && storage[i].getKey().equals(key)) {
                storage[i].setValue(value);
                return;
            }
        }
    }
```

olekskov (approved):
Approved. For further please consider making checks before main logic in order not to perform additional actions in case there we may just skip them.
Example is below:

PR #1049, opened 2023-04-20
Implemented the Storage of key-value type that is parameterized with

Stipaxa-hub:
two types of data.

Implemented put method, which can put value by key and replace value if key is occupied. Implemented method get, which return value by key. Implemented method size, which return size storage. Added indexOf method, which search unique key and return the key.


okuzan (approved):
Nice job!

PR #1050, opened 2023-04-21
solved task

OlehPryshchepa:


rovein (approved):
Good job! Keep the rhythm!

Try to solve this task once more time, but using two arrays (without linked nodes).

PR #1051, opened 2023-04-22
completed task

sZlishchev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 
     @Override
     public V get(K key) {
+        int index = 0;
+        for (K storageKey : keys) {
+            if (storageEquals(key, storageKey)) {
+                return values[index];
+            }
+            index++;
+        }
```
sarakhmen, line 41: why not just use a for loop here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private static boolean storageEquals(Object first, Object second) {
```
sarakhmen, line 50:
```suggestion
    private boolean keyEquals(K first, K second) {
```
I'd recommend declaring it this way

PR #1052, opened 2023-04-24
Solution

ArtEg0:
all tests passed.
added logic to Override methods: put, get.
added new methods: findIndex and StorageImpl.
added SuppressWarnings("unchecked") for public StorageImpl.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,53 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
+    private int findIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null || keys[i] != null && keys[i].equals(key)) {
```
olekskov, line 20:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

olekskov (approved):
Good job!
Below condition may be simplified:

PR #1053, opened 2023-04-24
implemented the Storage class parameterized by two data types in whic…

V-Solianyk:
…h added the necessary fields, constructor constants, and also wrote the implementation of three methods.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,48 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_SIZE = 0;
+    private static final int CAPACITY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[CAPACITY];
+        values = (V[]) new Object[CAPACITY];
+        size = INITIAL_SIZE;
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean noEqualsElement = true;
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                noEqualsElement = false;
+            }
+        }
+        if (noEqualsElement) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key != null && keys[i] != null && (key.equals(keys[i]))) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        keys = (K[]) new Object[CAPACITY];
+        values = (V[]) new Object[CAPACITY];
+        size = INITIAL_SIZE;
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean noEqualsElement = true;
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || key != null && key.equals(keys[i])) {
```
sarakhmen, line 22:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```
also this check could be moved to a separate method and reused in your program

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+                values[i] = value;
+                noEqualsElement = false;
```
sarakhmen, line 24: let's directly return from the method instead of using this boolean variable here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_SIZE = 0;
```
sarakhmen, line 6: redundant here actually

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private static final int CAPACITY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[CAPACITY];
+        values = (V[]) new Object[CAPACITY];
+        size = INITIAL_SIZE;
```
sarakhmen, line 15: is initialized to 0 by default

sarakhmen (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
V-Solianyk: thank you, but I invented my "bicycle", but it was possible to make a method of comparison simply based on the image of OBJECTS

src/main/java/core/basesyntax/impl/StorageImpl.java
V-Solianyk: almost the same story, as the previous comment, I constantly complicate the solution and code for myself

sarakhmen (approved):
Good job ;)

PR #1056, opened 2023-04-25
task solved

MykytaKat:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[10];
```
sarakhmen, line 11: 10 is a magic number here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        this.values = (V[]) new Object[10];
+        this.size = 0;
```
sarakhmen, line 13: size is initialized with 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
```
sarakhmen, line 32: code duplication. Could you merge these two loops and avoid using if-else?

sarakhmen (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((key == keys[i]) || (key != null && key.equals(keys[i]))) {
```
sarakhmen, line 33: I meant to reuse that method here :)
Also you can remove redundant parentheses:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean isMatch(K key, int index) {
```
sarakhmen, line 44: It'd be better to have 2 keys as parameters

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        return (key == null) ? (keys[index] == null) : key.equals(keys[index]);
```
sarakhmen, line 45:
```suggestion
        return key == null ? keys[index] == null : key.equals(keys[index]);
```

sarakhmen (approved):
Good job :)

PR #1057, opened 2023-04-28
Solution of this task

IvanovYurii:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_COUNT = 10;
+    private final Object[] keyArray = new Object[ARRAY_COUNT];
+    private final Object[] valueArray = new Object[ARRAY_COUNT];
+    private int size = 0;
```
kozhukhovsky, line 9: 0 is default value

```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_COUNT = 10;
```
kozhukhovsky, line 6:
```suggestion
    private static final int MAX_LENGTH = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private final Object[] keyArray = new Object[ARRAY_COUNT];
+    private final Object[] valueArray = new Object[ARRAY_COUNT];
```
kozhukhovsky, line 8:
```suggestion
    private final Object[] keys = new Object[ARRAY_COUNT];
    private final Object[] values = new Object[ARRAY_COUNT];
```

kozhukhovsky (changes requested):
Solution with overengineering. You do not need to use a hashcode, in the put method it is enough to check whether there is such a key or not, if there is, overwrite the value, if not, write it to a new cell. In the get method, iterate over and return a value if the key matches.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }
+
```
kozhukhovsky, line 47: remove empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH = 10;
+    private final Object[] keys = new Object[MAX_LENGTH];
+    private final Object[] values = new Object[MAX_LENGTH];
+    private int size;
+
+    public boolean isEquel(Object firstObject, Object secondObject) {
```
kozhukhovsky, line 11:
```suggestion
    private boolean isEqual(Object firstObject, Object secondObject) {
```

And move it to the end of the class, first the public methods, then the private ones

kozhukhovsky (changes requested):
Good! Need small fixes

kozhukhovsky (approved):

PR #1059, opened 2023-04-29
Created two arrays and size variable :

RostyslavOnysh:
First array "keys" store keys value;
Second array "values" store value;
I use "size" variable to check how many keys store in array

Override equals method to check does the key exist in the keys array.If the key has been found, I replace its value in the values array.

If the key is not present, I add a new key and its corresponding value to the end of the keys and values arrays, using the variable size to determine the position in the array.

In the "get()" method, I locate the key in the keys array in the same way as in the "put()" method, and return the corresponding value from the values array for the found key.

In the size method, I return size.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,80 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[10];
+        values = (V[]) new Object[10];
+        size = 0;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        StorageImpl current = (StorageImpl) o;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], current.keys[i])) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(size);
+        result = 31 * result + Arrays.hashCode(keys);
+        result = 31 * result + Arrays.hashCode(values);
+        return result;
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean found = false;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+        if (size == keys.length) {
+            keys = Arrays.copyOf(keys, size * 2);
+            values = Arrays.copyOf(values, size * 2);
+        }
+        keys[size] = key;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
+
 }


RostyslavOnysh:
I have a question, I have warnings highlighted with this context:
 "Warning:(14, 17) Unchecked cast: 'java.lang.Object[]' to 'K[]'
Warning:(15, 19) Unchecked cast: 'java.lang.Object[]' to 'V[]'
Warning:(27, 10) Raw use of parameterized class 'StorageImpl'
Warning:(27, 33) Raw use of parameterized class 'StorageImpl'.
And I can't seem to fix them in any way, do you have any advice on how to fix them properly?

Another question is whether it would be appropriate to use a more comprehensive equals method for this task. Here is an example of such a method for this task:  

 public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (o == null || getClass() != o.getClass()) {
            return false;
        }
        if (o.getClass().equals(StorageImpl.class)) {
            StorageImpl current = (StorageImpl) o;
            if (current.keys.length != keys.length){
                return false;
            }for (int i = 0; i < keys.length;i++){
                if(!Objects.equals(keys[i], current.keys[i])){
                    return false;
                }
                if (!Objects.equals(values[i],current.values[i])){
                    return false;
                }
            }
            return true;
        }
        return false;
    }

in this method, I also compare the length and use the Object's.equals() method, which compares two objects for equality, taking into account the possibility of null values. 🙈

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[10];
+        values = (V[]) new Object[10];
```
Ivan95kos, line 14: All magic numbers, strings, symbols in your code should be constants.
[this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+        size = 0;
```
Ivan95kos, line 15: int has default value 0
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+            }
+        }
+        if (!found) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+        if (size == keys.length) {
+            keys = Arrays.copyOf(keys, size * 2);
+            values = Arrays.copyOf(values, size * 2);
```
Ivan95kos, line 60: All magic numbers, strings, symbols in your code should be constants.
[this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+        }
+        keys[size] = key;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
Ivan95kos, line 68: Don’t use class Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(size);
+        result = 31 * result + Arrays.hashCode(keys);
+        result = 31 * result + Arrays.hashCode(values);
+        return result;
+    }
```
Ivan95kos, line 41: For this task, it is not necessary to implement the equals and hash code

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
+        if (!found) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+        if (size == keys.length) {
+            keys = Arrays.copyOf(keys, size * ARRAY_MULTIPLIER);
+            values = Arrays.copyOf(values, size * ARRAY_MULTIPLIER);
+        }
     }
```
olekskov, line 56:
```suggestion
    public void put(K key, V value) {
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || keys != null && key.equals(keys[i])) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
        if (size == keys.length) {
            keys = Arrays.copyOf(keys, size * ARRAY_MULTIPLIER);
            values = Arrays.copyOf(values, size * ARRAY_MULTIPLIER);
        }
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,80 @@
...
 
-    @Override
     public V get(K key) {
+        if (key == null) {
```
olekskov, line 59: Please also rework this method according to comment above

olekskov (changes requested):

kozhukhovsky (approved):

PR #1061, opened 2023-04-30
Done task

ASnazar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                return (V)values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return occupiedValues - START_ITEMS_NUMBER;
```
sarakhmen, line 41: quite unusual approach. Could you replace your 'occupiedValue' with something like 'index' and initialize it with 0 by default (actually it will be initialized with 0 by default if 'index' is your class field) and increment it after a new element is added?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+            occupiedValues++;
+            keys[occupiedValues] = key;
+            values[occupiedValues] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i <= occupiedValues; i++) {
+            if (keys[i] != null && keys[i].equals(key) || key == keys[i]) {
```
sarakhmen, line 32: I think this condition is duplicated in another method (checkKey()). Could you please create a separate method for this comparison and reuse it where needed?

sarakhmen (changes requested):
good job but let's slightly improve your solution :)

PR #1062, opened 2023-05-01
Solved

Poloutinho:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
+        this.values = (V[]) new Object[MAX_SIZE];
+        this.size = 0;
```
sarakhmen, line 14: size is initialized with 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
+        this.values = (V[]) new Object[MAX_SIZE];
```
sarakhmen, line 13: let's not reference 'this', you can just use 'keys'/'values' here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOf(key);
+        if (index >= 0) {
```
sarakhmen, line 20: let's inline this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+            // Key already exists, replace the value
```
sarakhmen, line 21: also let's remove this comment :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+            values[index] = value;
+        } else {
+            // Key doesn't exist, add it
```
sarakhmen, line 24: and this one :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
     public V get(K key) {
-        return null;
+        int index = indexOf(key);
+        if (index >= 0) {
+            // Key exists, return the value
+            return values[index];
+        } else {
+            // Key doesn't exist, return null
+            return null;
+        }
```
sarakhmen, line 43:
```suggestion
        int index = indexOf(key);
        return index >= 0 ? values[index] : null;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
sarakhmen, line 53:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
let's simplify it this way :)

sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline: comment not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (indexOf(key) >= 0) {
+            // Key already exists, replace the value
+            values[indexOf(key)] = value;
```
maxline, line 20: `indexOf()` that has loop inside - called twice, it is better introduce a variable at first call and assing found value.

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline: not fixed

maxline (changes requested):
good job, let's improve the solution

rovein (approved):
Good job! Keep the rhythm!

PR #1063, opened 2023-05-01
jv-storage v0.1

eugenetolkunov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
+    private int size = 0;
```
sarakhmen, line 9: is initialized with 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
```
sarakhmen, line 8: As described in common mistakes, let's initialize these arrays in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
+    private int size = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```
sarakhmen, line 14:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
Also, actually, this comparison is duplicated within another method. Could you please move it to a separate method and call it whenever needed?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,37 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }
+
```
sarakhmen, line 39:
```suggestion
```

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+    }
+
+    public static boolean equalsKey(Object a, Object b) {
```
maxline, line 16: why `public` and `static`? also remove this method to the end of the class

maxline (changes requested):
well done! left just one comment

PR #1064, opened 2023-05-01
implemented StorageImpl class

nechupara:


ViktoriiaSemenchenko (approved):
good job

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[0];
+        values = (V[]) new Object[0];
```
rovein, line 11: From the requirements: **If you use arrays for your solution, we assume that the maximum number of elements in our Storage is 10.**

So, let's change it accordingly.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            }
+        }
         return -1;
     }
+
+    private <T> T[] increaseArraySizeByOne(T[] array) {
+        T[] newArray = (T[]) new Object[array.length + 1];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
+    }
```
rovein, line 54: Good, but you don't need it in that task, you will return to it in ArrayList.

You may need another field like `index` or `size` to track the actual amount of inserted elements.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[0];
+        values = (V[]) new Object[0];
+    }
+
     @Override
     public void put(K key, V value) {
+        int keyIndex = getKeyIndex(key);
+        if (keyIndex == -1) {
```
rovein, line 17: Let's move `-1` to a constant.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            keyIndex = keys.length - 1;
+        }
+        keys[keyIndex] = key;
+        values[keyIndex] = value;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
rovein, line 29: Let's extract it to a separate method to remove code duplication.
```suggestion
            if (keysAreEqual(...) {
```

rovein (changes requested):
Great! But let's improve the solution a little bit.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,53 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private static final int NOT_FOUND_INDEX = -1;
+    private int currentSize;
```
rovein, line 8: Can be simply `size`.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,53 +3,49 @@
...
     private K[] keys;
     private V[] values;
 
     public StorageImpl() {
-        keys = (K[]) new Object[0];
-        values = (V[]) new Object[0];
+        currentSize = 0;
```
rovein, line 13: It will be initialized by '0' by default, so let's remove it.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,53 +3,49 @@
...
-            keys = increaseArraySizeByOne(keys);
-            values = increaseArraySizeByOne(values);
-            keyIndex = keys.length - 1;
+        if (keyIndex == NOT_FOUND_INDEX) {
+            if (currentSize >= 10) {
+                throw new RuntimeException("Can't add this value because the storage if full.");
+            }
+            currentSize++;
+            keyIndex = currentSize - 1;
+            keys[keyIndex] = key;
```
rovein, line 27: keys[currentSize] = key;
currentSize++;

or even 
keys[currentSize++] = key;

Experiment with it. [Reference](https://stackoverflow.com/questions/5413548/java-prefix-postfix-of-increment-decrement-operators)

rovein (changes requested):
Good job! But let's improve your solution ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
nechupara: Since I still need to reassign the _**keyIndex**_ variable, I have decided to stick to a compromise solution like this:
keyIndex = size++;
keys[keyIndex] = key;

sarakhmen (approved):
Good job :)

PR #1065, opened 2023-05-01
Solved task with using Generics

Pkrasota:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
+        this.values = (V[]) new Object[MAX_SIZE];
+        this.size = 0;
```
MykytaKat, line 14: you can delete this line, becouse size is 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexElement(K key) {
+        if (key == null) {
```
MykytaKat, line 49: it would be better if you wrote the method in a loop with only one if block

MykytaKat (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = indexElement(key);
+        if (index >= 0) {
+            return values[index];
+        } else {
```
maxline, line 36: this `else` clause can be simplified because of using `return` in previous line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexElement(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
maxline, line 49: please use for comparing the same approach as for method equals:
  `firstKey == secondKey || firstKey != null && firstKey.equals(secondKey);`
  
 I meant firstly check keys by `==` then `!= null`, then by `equals`

maxline (changes requested):
well done! left a couple small comments

maxline (approved):
well done! approved

PR #1067, opened 2023-05-01
add StorageImpl solution

AnastasiiaVariuchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private final K[] keysArray = (K[]) new Object[MAX_ARRAY_SIZE];
+    private final V[] valuesArray = (V[]) new Object[MAX_ARRAY_SIZE];
```
sarakhmen, line 9: as described in common mistakes let's initialize these arrays in your constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+                valuesArray[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keysArray.length; i++) {
+            if (Objects.equals(key, keysArray[i])) {
```
sarakhmen, line 33: as described in common mistakes let's not use Objects class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+            if (keysArray[i] == null && valuesArray[i] == null) {
+                size++;
+                keysArray[i] = key;
+                valuesArray[i] = value;
+                return;
+            } else if (Objects.equals(key, keysArray[i])) {
+                valuesArray[i] = value;
+                return;
+            }
+        }
```
sarakhmen, line 27: consider how can you use your 'size' field in this method in order to simplify its logic

sarakhmen (changes requested):
Good job but let's improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keysArray = (K[]) new Object[MAX_ARRAY_SIZE];
+        valuesArray = (V[]) new Object[MAX_ARRAY_SIZE];
+    }
+
+    private boolean isEqualsKeys(Object first, Object second) {
+        return (first == second) || (first != null && first.equals(second));
+    }
```
sarakhmen, line 19: let's put your private methods after public ones and also:
```suggestion
    private boolean isEqualsKeys(K first, K second) {
        return first == second || first != null && first.equals(second);
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keysArray.length; i++) {
+            if (keysArray[i] == null && valuesArray[i] == null) {
+                size = size() + 1;
```
sarakhmen, line 25: size++; Actually in my previous comment I meant that you can iterate in a loop to find if key is present and if not - then simply add a new value and key using your 'size' as an index. But let's leave as is :)

src/main/java/core/basesyntax/impl/StorageImpl.java
sarakhmen: lets use your isEqualsKeys here

sarakhmen (changes requested):
Let's slightly improve your solution :)

m-pnsk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean isEqualsKeys(K first, K second) {
```
maxline, line 46:
```suggestion
    private boolean areKeysEqual(K first, K second) {
```

rovein (approved):
Good job! Keep the rhythm!

PR #1069, opened 2023-05-02
implemented the Storage of key-value type that is parameterized with …

ViktoriiaSemenchenko:
…two types of data.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,56 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH = 10;
+    private K[] key;
+    private V[] value;
+    private int size;
+
+    public StorageImpl() {
+        key = (K[]) new Object[MAX_LENGTH];
+        value = (V[]) new Object[MAX_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (isStorageFull()) {
+            throw new RuntimeException("Can not add key: "
+                + key + ", value: " + value + ". Storage is full! ");
+        }
+        for (int i = 0; i < size; i++) {
+            if (compare(this.key[i], key)) {
+                this.value[i] = value;
+                return;
+            }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
-        return -1;
+        return size;
+    }
+
+    private boolean isStorageFull() {
+        return size == MAX_LENGTH;
+    }
+
+    private boolean compare(K key1, K key2) {
+        return Objects.equals(key1, key2);
```
nechupara, line 54: According to the article "Common mistakes (jv-storage)" we are restricted of using Objects class (quotation: "Don’t use class Objects"). So I suggest you to refactor this line.

nechupara (changes requested):
Good job! But let's improve your solution :-D

nechupara (approved):
Looks good to me

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                + key + ", value: " + value + ". Storage is full! ");
+        }
+        for (int i = 0; i < size; i++) {
+            if (compare(this.key[i], key)) {
+                this.value[i] = value;
+                return;
+            }
+        }
+        this.key[size] = key;
+        this.value[size] = value;
```
maxline, line 29: using `this` keyword is not necessary here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH = 10;
```
maxline, line 6: let's rename as `MAX_SIZE` for variables naming consistency

maxline (changes requested):
Well done! Left just a couple of minor remarks.

src/main/java/core/basesyntax/impl/StorageImpl.java
ViktoriiaSemenchenko: but I have error without THIS keyword

maxline (approved):
Well done! Approved

PR #1071, opened 2023-05-02
implemented task

chrisnyrka:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,83 @@
...
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
Gamemod13, line 39: you can simplify that part - cast it to another method, that simplify understanding and readability

Gamemod13 (commented):
Hi! Great job! But let improved it ; )

Gamemod13 (approved):
Cool! Good work)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
...
+        size = 0;
+    }
+
+    public void putKeyIsNull(V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                values[i] = value;
+                return;
+            }
+        }
```
rovein, line 23: Let's extract it into a separate method, e.g like this:
```suggestion
         if (containsKey(key)) {
                values[indexOf(key)] = value;
                return;
        }
```

Don't duplicate the code. Try to create a method that checks that key is already present. Also, try to create a method that returns the index of the key, which will be used to store the value.

It will significantly simplify your code and reduce the number of repetitive lines.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
...
+        keys[size] = null;
+        values[size] = value;
+        size++;
+    }
+
+    public void putKeyIsNotNull(K key, V value) {
```
rovein, line 29: Let's don't create a repetitive code, also let's don't create separate methods "not null" and "is null". Let's have this checks within the main methods.

Check all other cases below.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            if (containsKey()) {
+                values[indexOf(key)] = value;
+                return;
+            }
+            keys[size] = null;
```
maxline, line 42: this `if` can be simplified
 instead of check `key == null` in the method begins, create method that compares keys with `null` check, following the same recommendations that were for overriding method `equals()` task
 ```
 private boolean compareKeys(K first, K second) {
        return (first == second) || first != null && first.equals(second);
    }```

maxline (changes requested):
good job! let's improve the solution. take a notice that you don't fix the all comments from the student review.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -32,9 +48,12 @@ private int indexOf(K key) {
         return 0;
     }
 
-    @Override
-    public void put(K key, V value) {
-        if (key == null) {
+    private boolean compareKeys(K firstKey, K secondKey) {
+        return Objects.equals(firstKey, secondKey);
```
rovein, line 52: Also, from checklist: **Don’t use class Objects**.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
+    private boolean containsKey() {
```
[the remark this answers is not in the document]
rovein: Not moved.

rovein (approved):
Good job! But see some left minor comments, pay attention in the next time :)

PR #1073, opened 2023-05-02
Implemented storage with key-value parameters

viprko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_ARRAY_LENGTH = 10;
```
nechupara, line 6: constants could be private

nechupara (changes requested):
Good job! But let's improve your solution :-D

src/main/java/core/basesyntax/impl/StorageImpl.java
viprko: Agree, my mistake

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    public static final int START_INDEX = 0;
```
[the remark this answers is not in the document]
viprko: In my opinion, in some cases, I used for-loop with a different Start index, not just with "0" and not just in private methods. Hence, the "Start index" should be CONSTANT for more encapsulation and hiding unnecessary information from users.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            size++;
+            return;
+        }
+        values[index] = value;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return (V) lookingForKeyValue(key);
```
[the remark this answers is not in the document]
viprko: +

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        int index = indexOfProperCell(key);
+        if (index == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+            return;
+        }
+        values[index] = value;
```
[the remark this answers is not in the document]
viprko: I can't do this, cos all logic was built for this block (1st - value[SIZES], 2nd - values[INDEX]) and I used "return" for exit from the method. But I will change it to if - else for better readability.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = getIndexForPutting(key);
+        if (index == NO_MATCH_INDEX) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[index] = value;
```
nechupara, line 26: Of course I'm not sure if my variant is good or not, but I just wanted to show it :D
```suggestion
         int index = getIndexForPutting(key);
        if (index == NO_MATCH_INDEX) {
            keys[size] = key;
            index = size;           
            size++;
        } 
        values[index] = value;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private int getIndexForPutting(K key) {
+        for (int i = START_INDEX; i < size; i++) {
+            if ((key == keys[i]) || key != null && key.equals(keys[i])) {
```
nechupara, line 43: Based on mentor's review of my homework I have to add one more change request:
You should extract logical condition from "if" to a separate method. Like this: 
```suggestion
           if (keysAreEqual(...) {
```
Don't ask me why :-D

nechupara (changes requested):
One last time )))

src/main/java/core/basesyntax/impl/StorageImpl.java
viprko: I'm not seen your entire code, but with my solution, if ain't "break" from the method, it's not work

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        for (int i = START_INDEX; i < size; i++) {
+            if (areKeysMatch(key, keys[i])) {
+                return i;
+            }
+        }
+        return NO_MATCH_INDEX;
+    }
+
+    private boolean areKeysMatch(K keyOne, K keyTwo) {
+        return ((keyOne == keyTwo) || keyOne != null && keyOne.equals(keyTwo));
```
nechupara, line 51: About extracting conditional logic from "if" statement to a separate method areKeysMatch() it was my fault, sorry. Let me explain why it happened :D In my task I had such a feedback from mentor but such a case  occurred twice in my code. So in my case it was code duplication. 
Feel free to undo this extraction if you want. But at least you will be aware of such potential cases in future :-D

nechupara (approved):
All good. But please read my last comment  in the code ))

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_LENGTH = 10;
+    private static final int START_INDEX = 0;
```
maxline, line 7: this constant is redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                : values[getIndexForPutting(key)];
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private int getIndexForPutting(K key) {
```
maxline, line 41: rename simple as `findIndex()` or `indexOf()`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        for (int i = START_INDEX; i < size; i++) {
```
maxline, line 42: you can use `0` instead of `START_INDEX`, it is common practice for iterating the array

src/main/java/core/basesyntax/impl/StorageImpl.java
maxline, line 51: please remove redundant braces in `(keyOne == keyTwo)`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            size++;
+        } else {
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return getIndexForPutting(key) == NO_MATCH_INDEX ? null
```
maxline, line 32: please introduce a variable for found index, because method `getIndexForPutting()` has loop inside ant it called twice.

maxline (changes requested):

rovein (approved):
Good job! Keep the rhythm!

PR #1075, opened 2023-05-02
Implement the Storage

DV7Art:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+                        + "\nvalue: "
+                        + value
+                        + "Storage is full");
+        }
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
```
marimarig, line 31: I'd recommend create new methods, what will return index elements

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                return values[i];
+            }
+        }
```
marimarig, line 43: also, create method, what will return index

src/main/java/core/basesyntax/impl/StorageImpl.java
marimarig, line 31: I'd recommend create a new method, what will return index elements

src/main/java/core/basesyntax/impl/StorageImpl.java
marimarig, line 43: also, create method, what will return index

src/main/java/core/basesyntax/impl/StorageImpl.java
marimarig, line 31: I'd recommend create a new method, what will return index elements

src/main/java/core/basesyntax/impl/StorageImpl.java
marimarig, line 43: also, create method, what will return index/ to avoide code duplication

marimarig (commented):
Some recommendations

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == MAX_NUMBER) {
+            throw new RuntimeException("Can't add key: "
+                        + key
+                        + "\nvalue: "
+                        + value
+                        + "Storage is full");
```
rovein, line 24: Good point! Also, you can specify come concrete exception, like `IllegalStateException`.

Also, let's try using `String.format()` here.
```suggestion
            throw new IllegalStateException(String.format("Can't add key: %s, value: %s. Storage is full.", key, value));
```

src/main/java/core/basesyntax/impl/StorageImpl.java
rovein: Let's do it. Create method `indexOf(T key)`.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+        }
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
rovein, line 27: Also, from checklist: **Don’t use class Objects**.
```suggestion
            if (...call here your own method...) {
```

rovein (changes requested):
Great solution! But let's improve it a little bit ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
DV7Art: I can`t understand how I can create here method.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean isEqual(Object a, Object b) {
```
sarakhmen, line 56:
```suggestion
    private boolean isEqual(K a, K b) {
```

PR #1077, opened 2023-05-02
solution jv-storage

mykyta-udav:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
MykytaKat, line 14: you can delete this line, becouse size is zero by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```
MykytaKat, line 20: you can make separate method to avoid code duplication in 20 and 33 lines

MykytaKat (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (isSimilar(key, i)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
+    private boolean isSimilar(K key, int index) {
```
maxline, line 39: 1. please rename as `areEquals()` or `compareKeys()`
2. as second param better to put the `keys[index]` instead of `index`

maxline (changes requested):

rovein (approved):
Good job! Keep the rhythm!

PR #1078, opened 2023-05-02
Complete task and pass all tests.

Fafakepa:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            } else if (areKeysEqual(keys[i], key) && values[i] != null) {
+                values[i] = value;
+                size--;
+            } else if (values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+            }
+        }
+        size++;
     }
```
Tupadzaky, line 31: overcomplicated if else logic. 
You can use your areKeysEqual() method in a loop to find whether the same key already exists, and if it does than you just replace the value that corresponds to that key and exit the method. If key doesn't already exist then you can just put new key and new value at the current size of your storage, with no auxiliary loops or checks needed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean areKeysEqual(K keyFirst, K keySecond) {
+        return (keyFirst != null && keyFirst.equals(keySecond)) || keyFirst == keySecond;
     }
```
Tupadzaky, line 50: you should swap your check by reference to the beginning, because if two keys are equal by reference it means that they are the same objects, and you don't need to check their actual values

Tupadzaky (changes requested):
good job overall, but let's improve your solution

Tupadzaky (approved):
good job

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ARRAY_SIZE];
+        values = (V[]) new Object[MAX_ARRAY_SIZE];
+        size = 0;
```
rovein, line 14: It will be initialized by '0' by default, so let's remove it from here.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int j = 0; j < size(); j++) {
+            if (areKeysEqual(keys[j], key)) {
+                return values[j];
+            }
+        }
```
rovein, line 36: Let's extract it to a separate method.

```suggestion
private int indexOf(key) {
        for (int j = 0; j < size(); j++) {
            if (areKeysEqual(keys[j], key)) {
                return i;
            }
        }
}        
```

rovein (approved):
Great job! See some comments-tips below.

PR #1081, opened 2023-05-03
implement storage interface using two  arrays

ArinaPolna:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEM_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
```
AnastasiiaVariuchuk, line 9: better to use K[] keys and  V[] values instead of Object[] keys and Object[] values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private int index;
+
+    public StorageImpl() {
```
AnastasiiaVariuchuk, line 12: and cast it here like (V[]) new Object[MAX_ITEM_SIZE]

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean equalsObjects(Object first, Object second) {
```
AnastasiiaVariuchuk, line 47: better to use K keys and V values instead of Object keys and Object values(than rename your method like equalsKeys)

AnastasiiaVariuchuk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEM_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+    private int index;
```
maxline, line 10: `index` is a redundant class field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (equalsKeys(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[index] = key;
+        values[index] = value;
```
maxline, line 26:
```suggestion
        keys[size] = key;
        values[size] = value;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        index++;
```
maxline, line 27: `index` variable is redundant

maxline (changes requested):

maxline (approved):
well done! approved

PR #1082, opened 2023-05-03
done hm

vladyslavkryvets:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = indexElement(key);
+        return index >= 0 ? values[index] : null;
+
```
volodymyrruso, line 35: remove empty line

volodymyrruso (changes requested):
Good job! only one little mistake I found

src/main/java/core/basesyntax/impl/StorageImpl.java
vladyslavkryvets: ty, done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexElement(key);
+        if (index >= 0) {
+            this.values[index] = value;
+        } else {
+            if (size == this.keys.length) {
+                throw new OutOfMemoryError("Storage is completely full");
```
rovein, line 23: It seems logically correct, but we never throw any `Error` in our code. `Error` indicates error on the JVM side. So, it would be better to throw something like `IllegalStateException` here.

rovein (approved):
Good job! Left one comment, take a look at it.

PR #1084, opened 2023-05-03
implemented Storage class

maksigors:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keyEquals(key, keys[i])) {
+                return values[i];
+            }
+        }
         return null;
```
marimarig, line 39: suggest following: return (keyEquals(key, keys[i]) ? value[i] : null;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+                keys[i] = key;
+                values[i] = value;
+                size++;
+                return;
+            }
+            if (keyEquals(key, keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
marimarig, line 29: maybe need throw new RuntimeExeption (storage full) if size == keys.length;

marimarig (changes requested):
Good, but i've some remarks

marimarig (approved):
Actually, i don't really agree with this decision, but we'd better call a mentor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,58 @@
...
+    }
+
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean keyEquals(Object first, Object second) {
+        return (first == second) || (first != null && first.equals(second));
```
sarakhmen, line 57:
```suggestion
    private boolean keyEquals(K first, K second) {
        return first == second || first != null && first.equals(second);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,58 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (getKeyIndex(key) >= 0) {
+            return values[getKeyIndex(key)];
+        }
         return null;
```
sarakhmen, line 39: let's call your getKeyIndex() method only once here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,58 @@
...
+                keys[i] = key;
+                values[i] = value;
+                size++;
+                return;
+            }
+            if (keyEquals(key, keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
sarakhmen, line 31: Could you use your getKeyIndex method for this method purposes

sarakhmen (changes requested):
Good job, but let's improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
maksigors: yep, added Exeption and change some code with size against length, thx))

src/main/java/core/basesyntax/impl/StorageImpl.java
maksigors: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
maksigors: ok

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,54 @@
...
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keyEquals(key, keys[i])) {
+                return i;
+            }
+        }
+        return INDEX_NOT_FOUND;
     }
```
sarakhmen, line 45: also let's put your private methods after public ones

sarakhmen (approved):
Good job, you can improve your solution if you want ;)

PR #1088, opened 2023-05-06
part of code

shaman1437:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAXNUMBER = 10;
```
okuzan, line 6: Constant names should be in uppercase and use underscores to separate words, e.g. MAX_NUMBER

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                return (V) values[i];
+            }
+        }
```
okuzan, line 35: if you wish you can extract the common logic pattern - iterating from 0 to size to find equal key - to a separate private method

okuzan (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
shaman1437: ty

PR #1089, opened 2023-05-06
implemented StorageImpl class with two arrays

vaniii1:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && key == null) {
+                count--;
+                return values[i];
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                count--;
```
okuzan, line 41: Remove 'count--' as it decreases the count unnecessarily when retrieving a value.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        this.keys = (K[]) new Object[LENGTH];
+        this.values = (V[]) new Object[LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && key == null) {
+                values[i] = value;
+                break;
```
okuzan, line 21: Add a 'return' statement after updating the value to prevent the key-value pair from being added again.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            }
+        }
+        keys[count] = key;
+        values[count] = value;
+        count++;
```
okuzan, line 30: Move this line inside an 'else' block after the 'for' loop to only increase the count if no existing key is found.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && key == null) {
+                values[i] = value;
+                break;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
```
okuzan, line 25: Add a 'return' statement after updating the value to prevent the key-value pair from being added again.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        keys[count] = key;
+        values[count] = value;
+        count++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == null && key == null) {
+                count--;
```
okuzan, line 37: Remove 'count--' as it decreases the count unnecessarily when retrieving a value.

okuzan, line 37: you have the same body in both if clauses. Combine conditions

src/main/java/core/basesyntax/impl/StorageImpl.java
okuzan, line 25: you have the same body in both if clauses. Combine conditions

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int LENGTH = 10;
```
okuzan, line 6:
```suggestion
    private static final int DEFAULT_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private int count = 0;
```
okuzan, line 7: FYI 0 is default for int

```suggestion
    private int size = 0;
```

okuzan (changes requested):

PR #1098, opened 2023-05-24
Hw storage

kaptetos:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+                    return i;
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
```
olekskov, line 59:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return i;
            }
        }
```

olekskov (changes requested):
Good job overall.
Let's adjust condition to handle both null and not null possibilities simultaniously

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
okuzan, line 49: image
IDEA probably suggested this replacement, but this is checklist issue

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 14: 0 is the default for int, remove this

okuzan (approved):
Good job

PR #1099, opened 2023-05-24
key-value implements

karvatskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_ARRAY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE_ARRAY];
+        values = (V[]) new Object[MAX_SIZE_ARRAY];
+        size = 0;
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size < MAX_SIZE_ARRAY) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new IllegalStateException("Storage is full");
+            }
```
olekskov, line 29: It's better to keep code with minimum tabulations to the right where possible:
```suggestion
            if (size >= MAX_SIZE_ARRAY) {
                throw new IllegalStateException("Storage is full");
            } 
            keys[size] = key;
            values[size] = value;
            size++;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null) {
+                return i;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
```
olekskov, line 54:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                return i;
            }
```

olekskov (changes requested):
Good job overall, some minor comments

olekskov (approved):
Good job!

PR #1101, opened 2023-05-26
implementations storage

GradientAspen:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[DEFAULT_SIZE];
+        values = new Object[DEFAULT_SIZE];
+        size = 0;
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+        Object[] newKeys = new Object[newSize];
+        Object[] newValues = new Object[newSize];
+        System.arraycopy(keys, 0, newKeys, 0, size);
+        System.arraycopy(values, 0, newValues, 0, size);
+        keys = newKeys;
+        values = newValues;
     }
 
     @Override
     public V get(K key) {
```
olekskov, line 69: All public methods should be placed before all private ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+            putForNullKey(value);
+            return;
+        }
+
+        for (int i = 0; i < size; i++) {
+            if (key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
olekskov, line 29: There is no need to split logic for null and notNull flows. There is standard comparison pattern that is suiable for both null and notNull cases:
```suggestion
        for (int i = 0; i < size; i++) {
            if (keys == keys[i] || key != null && key.equals(keys[i])) {
                values[i] = value;
                return;
            }
        }
```
and same for get() operation

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+
+    public StorageImpl() {
+        keys = new Object[DEFAULT_SIZE];
+        values = new Object[DEFAULT_SIZE];
+
+    }
+
     @Override
     public void put(K key, V value) {
+
```
olekskov, line 19: starting method from empty line is a bad habit

src/main/java/core/basesyntax/impl/StorageImpl.java
olekskov: Not fixed, please change order of methods in the class. All public first, then private ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[DEFAULT_SIZE];
+        values = new Object[DEFAULT_SIZE];
+
```
olekskov, line 14: needless empty line

olekskov (changes requested):
Good job overall, style comments left

olekskov (approved):
Good job

PR #1106, opened 2023-06-07
completed task.

LightMyWay99:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
```
```suggestion
        keys = (K[]) new Object[MAX_SIZE];
        values = (V[]) new Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    return i;
+                }
+            } else {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
```
aaananas, line 55:
```suggestion
             if (keys[i] == key || key != null && key.equals(keys[i]) {
                    return i;
                }
```

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        }
     }
 
-    @Override
     public V get(K key) {
-        return null;
+        int index = findKeyIndex(key);
+        if (index >= 0) {
+            return (V) values[index];
+        } else {
```
okuzan, line 35: we don't need else after return :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size < MAX_SIZE) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new IllegalStateException("Storage is full");
+            }
```
```suggestion
            if (size >= MAX_SIZE) {
                throw new IllegalStateException("Storage is full");
            }
            keys[size] = key;
            values[size] = value;
            size++;

```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size < MAX_SIZE) {
```
okuzan, line 21:
```suggestion
            values[index] = value;
            return;
       }
         if (size < MAX_SIZE) {
```

okuzan (changes requested):
Suggested a few changes about code structure involving returns

okuzan (approved):
Looks nice 🪄

PR #1108, opened 2023-06-12
added solution: create getIndexByKey method, added algorithm to put, …

YuraKhilko:
…get, size methods.


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,54 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        int indexByKey = getIndexByKey(key);
+        if (indexByKey != INDEX_NOT_FOUND) {
+            return (V) valuesArray[indexByKey];
+        } else {
+            return null;
+        }
```
```suggestion
        }
        return null;
```

aaananas (approved):

PR #1110, opened 2023-06-12
 Implemented methods put(), get() and size() for StorageImpl

Sergey-Bogatkin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_ARRAY_SIZE = 0;
```
okuzan, line 6: Change the value of START_ARRAY_SIZE to 10, as the maximum array size should be 10.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key == keys[i]) || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (keys.length == size) {
+            increaseArray(size + 1);
```
okuzan, line 20: Remove the increaseArray method and its call, as the array size should be constant.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key) || keys[i] != null && keys[i].equals(key)) {
```
okuzan, line 30: extract this common comparison/iterating logic from get and put

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key == keys[i]) || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (keys.length == size) {
```
okuzan, line 19: You don't need to check if keys.length equals size, as the array size is constant. Remove this if statement.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int START_ARRAY_SIZE = 0;
+    private int size = 0;
```
okuzan, line 7: 0 is default for int, no need for this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private Object[] keys = new Object[START_ARRAY_SIZE];
+    private Object[] values = new Object[START_ARRAY_SIZE];
```
```suggestion
    private K[] keys;
    private V[] values;
```

okuzan: image

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+
+    public StorageImpl() {
+
+    }
+
+    public StorageImpl(Object[] keys, Object[] values, int size) {
+        this.keys = keys;
+        this.values = values;
+        this.size = size;
+    }
```
olekskov, line 20: According to Oleh's comment before:
```suggestion
    private static final int STORAGE_SIZE = 10;
    private K[] keys;
    private V[] values;
    private int size;
    
    public StorageImpl() {
        this.keys = (K[]) new Object[STORAGE_SIZE];
        this.values = (V[]) new Object[STORAGE_SIZE];
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+
+    private boolean hasAKey(K key) {
+        for (int i = 0; i < size; i++) {
+            index = i;
+            if ((key == keys[i]) || key != null && key.equals(keys[i])) {
+                return true;
+            }
+        }
+        return false;
```
olekskov, line 53: Using class variable to hold some state should be used only when really necessary as such decision requires additional monitoring of it's state and may cause unintentional behaviour.
Here we may easily return index instead of keeping it in class variable:
```suggestion
    private int getKeyIndex(K key) {
        for (int i = 0; i < size; i++) {
            if ((key == keys[i]) || key != null && key.equals(keys[i])) {
                return i;
            }
        }
        return -1;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     @Override
     public void put(K key, V value) {
+        if (hasAKey(key)) {
+            values[index] = value;
+            return;
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
```
olekskov, line 31: And then use it here:
```suggestion
    public void put(K key, V value) {
        int index = getKeyIndex(key);
        if (index != -1) {
            values[index] = value;
            return;
        }
        keys[size] = key;
        values[size] = value;
        size++;
    }
```
And similarly in other method

olekskov (changes requested):

kozhukhovsky (approved):

PR #1113, opened 2023-06-14
jv-storage done; implementing method put which checks if the key is n…

danmalins:
…ull and if the storage already contains such element; if storage already contains such element it rewrites data with value if not and storage is less than MAX_SIZE we add new pair and make size++, when the storage is full we output message and throw a stateException, so do with keys. In size() we gain current size of the storage and in get() we gain value by key. IN equals() and HashCode() we compare two objects StorageImpl

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,72 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Map<K, V> storageMap;
+    private int size;
+
+    public StorageImpl() {
+        this.storageMap = new HashMap<>();
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            if (storageMap.containsKey(null)) {
+                storageMap.put(null, value);
+            } else {
+                if (size < MAX_SIZE) {
+                    storageMap.put(null, value);
+                    size++;
+                } else {
+                    throw new IllegalStateException("Storage is full");
+                }
+            }
+        } else {
+            if (storageMap.containsKey(key)) {
+                storageMap.put(key, value);
+            } else {
+                if (size < MAX_SIZE) {
+                    storageMap.put(key, value);
+                    size++;
+                } else {
+                    throw new IllegalStateException("Storage is full");
+                }
+            }
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return storageMap.get(key);
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        StorageImpl<?, ?> other = (StorageImpl<?, ?>) obj;
+        return Objects.equals(storageMap, other.storageMap);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(storageMap);
     }
 }
+


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Map<K, V> storageMap;
```
aaananas, line 10: image

aaananas (changes requested):
**Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected**

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
```
aaananas, line 14:
```suggestion
```
calss fields have default values, for int default value is 0

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
+
+    private int arrayHashCode(Object[] arr) {
+        if (arr == null) {
+            return 0;
+        }
+        int result = 1;
+        for (Object element : arr) {
+            result = 31 * result + (element == null ? 0 : element.hashCode());
+        }
+        return result;
```
aaananas, line 128: you should remove those methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        StorageImpl<?, ?> other = (StorageImpl<?, ?>) obj;
+        return arrayEquals(keys, other.keys)
+                && arrayEquals(values, other.values)
+                && size == other.size;
```
aaananas, line 90: remove this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
+
+    private void putValue(Object key, Object value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null ? key == null : keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
```
aaananas, line 76:
```suggestion
        for (int i = 0; i < size; i++) {
            if (keys[i] == null ? key == null : keys[i].equals(key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
+                putValue(key, value);
+            } else {
+                if (size < MAX_SIZE) {
+                    putValue(key, value);
+                    size++;
+                } else {
+                    throw new IllegalStateException("Storage is full");
+                }
+            }
+        }
```
aaananas, line 41: you dont need this logic, move code from your method putValue here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,128 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean containsKey(Object key) {
```
aaananas, line 59: dont need this method

aaananas (changes requested):

Rommelua (approved):

PR #1116, opened 2023-06-16
Implement Storage interface in StorageImpl. In the implementation, tw…

Olga-Melnichenko:
…o arrays are used to store the keys and values, with a maximum capacity of 10 elements. If a key-value pair with the same key already exists in the storage, the value is replaced.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,46 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
     @Override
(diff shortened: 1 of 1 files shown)


Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            size++;
+        } else {
+            throw new RuntimeException("Vault full");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i]) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                return;
+            }
+        }
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Vault full");
+        }
```
okuzan, line 30:
```suggestion
        if (size >= MAX_SIZE) {
            throw new RuntimeException("Vault full");
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

okuzan (changes requested):

kozhukhovsky (approved):

PR #1118, opened 2023-06-16
added implementation method

andriihomenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+
+    public StorageImpl() {
+        keysArray = (K[]) new Object[SIZE_ARRAY];
+        valuesArray = (V[]) new Object[SIZE_ARRAY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keysArray[i], key)) {
```
aaananas, line 20: you are not allowed to use Objects class here, see common mistakes

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
andriihomenko: it was a tip from IDEA

aaananas (approved):

PR #1119, opened 2023-06-16
Completed solution

AntonU91:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final Object[] keyArr = new Object[MAX_STORAGE_SIZE];
+    private final Object[] valuesArr = new Object[MAX_STORAGE_SIZE];
```
Ivan95kos, line 8: You should set default size in constructor
```suggestion
    private final Object[] keys;
    private final Object[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
...
+    private int currentSize = 0;
```
Ivan95kos, line 9: int has default value 0
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
...
+        if (currentSize < MAX_STORAGE_SIZE) {
+            putIfKeyIsNonNull(key, value);
+        } else {
+            System.out.println("You can't put new key-value pair "
+                    + "because storage size is overfilled");
+        }
     }
 
     @Override
     public V get(K key) {
```
Ivan95kos, line 26: a lot of actions,
You should have only this:
```
    @Override
    public V get(K key) {
                return values[i];
            }
        }

        return null;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final Object[] keyArr = new Object[MAX_STORAGE_SIZE];
+    private final Object[] valuesArr = new Object[MAX_STORAGE_SIZE];
+    private int currentSize = 0;
+
     @Override
     public void put(K key, V value) {
```
Ivan95kos, line 12: The same:
```
    @Override
                values[i] = value;
                return;
            }
        }

        keys[size] = key;
        values[size] = value;
        size++;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,83 @@
...
+            if (keyArr[i] != null && keyArr[i].equals(key)) {
+                return (V) valuesArr[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return currentSize;
```
Ivan95kos, line 44:
```suggestion
        return size;
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
AntonU91: Thanks a lot for such  valuable advices!

Ivan95kos (approved):
Nice

PR #1124, opened 2023-06-20
meyson-hw-jv-storage-solution

Meysonsan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,78 @@
...
+    private static final int MAX_SIZE = 10;
+    private static final int ELEMENT_NOT_FOUND = -1;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 16: 0 is default for int

okuzan (changes requested):
Simplify  your code, you don't need to handle cases with null separately. You should have null save comparison like this `key == null && keys[i] == null || key != null && key.equals(keys[i])`

PR #1130, opened 2023-06-25
jv-storage tests done

OlgaStarikova:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int lastStorageIndex;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_SIZE];
+        values = (V[]) new Object[MAX_STORAGE_SIZE];
+        lastStorageIndex = 0;
```
kozhukhovsky, line 14:
```suggestion
```

0 is int value by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
```
kozhukhovsky, line 18: It can be made a little easier. Loop through the array of keys and check if the key matches. If it matches, we simply write the new value to this cell and do a return. If there are no matches, then simply after the loop we record a new key and a new value in the next cell.

kozhukhovsky (approved):

PR #1133, opened 2023-06-26
Implement StorageImpl class to store key-value data

barbequeue:


barbequeue:
I agree, that this is a VERY complicated solution, but i just wanted to practice my skills with recursion and `this` keyword. I think, i've created some kind of Stack implementation 😅

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,101 @@
...
+                        && Objects.equals(value, storage.value)
+                        && Objects.equals(next, storage.next);
+            }
+
+            return Objects.equals(key, storage.key)
+                    && Objects.equals(value, storage.value);
+        }
+
+        return false;
     }
```
olekskov, line 100: We aren't going to compare storages so these methods are redundant
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,101 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K key;
+    private V value;
+    private StorageImpl<K, V> next;
+
+    public StorageImpl() {
+        next = this;
+    }
```
olekskov, line 13: Task requirement "You should solve this task using one or two arrays" is not followed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,101 @@
...
     public void put(K key, V value) {
+        if (next == this) {
+            this.key = key;
+            this.value = value;
+            next = null;
+            return;
+        }
+
+        // Overriding the value if the keys match
+        if (Objects.equals(this.key, key)) {
```
olekskov, line 25: Usage of Objects class is prohibited for this task
https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-use-class-objects

olekskov (changes requested):
Please rework solution according to task description.
You will be asked to implement own ArrayList, LinkedList and HashMap in further tasks

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final byte MAX_ITEMS_COUNT = 10;
+    private byte size;
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        size = 0;
```
Rommelua, line 12:
```suggestion
```

Rommelua (approved):

PR #1135, opened 2023-06-26
Task complete

anna-poplavska:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_LENGTH];
+        values = (V[]) new Object[MAX_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_LENGTH; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key && values[i] != null) {
```
olekskov, line 19: Here we may use same structure as in get method:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```

olekskov (approved):
Good job!

src/main/java/core/basesyntax/impl/StorageImpl.java
anna-poplavska: If I remove it, two tests will fail.
image

PR #1138, opened 2023-06-28
Added implements methods: put, get, size with use generic

hanshyn:
Added implements methods: put, get, size with use generic.
check mvn.
check test 7 of 7.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,43 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_STORAGE = 10;
+    private final K[] storageKeys = (K[]) new Object[MAX_SIZE_STORAGE];
+    private final V[] storageValues = (V[]) new Object[MAX_SIZE_STORAGE];
+    private int currentSizeStorage = 0;
+    private boolean repeatCheck = false;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currentSizeStorage; i++) {
+            if (key != null && key.equals(storageKeys[i])
+                    || (key == null && storageKeys[i] == null)) {
+                storageValues[i] = value;
+                repeatCheck = true;
+                currentSizeStorage--;
+            }
+        }
+        if (!repeatCheck) {
+            repeatCheck = false;
+            storageKeys[currentSizeStorage] = key;
+            storageValues[currentSizeStorage] = value;
+        }
+        currentSizeStorage++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentSizeStorage; i++) {
+            if (key != null && key.equals(storageKeys[i])
+                    || storageKeys[i] == key) {
+                return storageValues[i];
+            }
+        }
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_STORAGE = 10;
+    private final K[] storageKeys = (K[]) new Object[MAX_SIZE_STORAGE];
```
overpathz, line 7: Minor remark.
Since we have already within storage context it's better to name storageKeys/Values just as keys/values.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+    private boolean repeatCheck = false;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < currentSizeStorage; i++) {
+            if (key != null && key.equals(storageKeys[i])
+                    || (key == null && storageKeys[i] == null)) {
+                storageValues[i] = value;
+                repeatCheck = true;
+                currentSizeStorage--;
```
overpathz, line 19: Can we avoid of using repeatCheck variable?
Maybe we can exit the method after updating the value (return).
In other way, if the key doesn't exist in the storage, add it to the end.

overpathz (changes requested):
Nice)
Little fixes are needed to approve.

src/main/java/core/basesyntax/impl/StorageImpl.java
hanshyn: create new method for check key exist in the storage?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_STORAGE = 10;
+    private final K[] keys = (K[]) new Object[MAX_SIZE_STORAGE];
+    private final V[] values = (V[]) new Object[MAX_SIZE_STORAGE];
```
Ivan95kos, line 8: You should use constructor to initialize this variable
```suggestion
    private final K[] keys;
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private int currentSizeStorage = 0;
```
Ivan95kos, line 9: int has default value 0
do not use name of class in variables
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
+    public boolean checkKeyInStorage(K key, V value) {
```
Ivan95kos, line 11: Methods that use only in class should be private
All private methods should be below public methods

Give more appropriate names to the actions in the method, 
The `checkKeyInStorage`  method saves and ckecks

Ivan95kos (changes requested):

aaananas (approved):

PR #1139, opened 2023-06-29
implemented methods and add new method

JeniaSan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        if (index == -1) {
+            return null;
+        }
+        return values[index];
+    }
+
+    private int contains(K key) {
+        for (int i = 0; i < size; i++) {
+            boolean contain = key == null ? keys[i] == null : key.equals(keys[i]);
+            if (contain) {
```
nacenik, line 45: Why do you need an additional boolean variable? You can do all checks just in `if()` using `&&`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     public V get(K key) {
-        return null;
+        int index = contains(key);
+        if (index == -1) {
+            return null;
+        }
+        return values[index];
+    }
+
+    private int contains(K key) {
```
nacenik, line 42: All private methods should be after all public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
nacenik, line 14: 0 is default integer value
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = contains(key);
+        if (index == -1) {
+            return null;
```
nacenik, line 36: I'd recommend you to throw an exception here with a message to avoid NPE

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #1149, opened 2023-07-06
added solution

satori11111:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAXIMUM_STORAGE_SIZE];
+        values = (V[]) new Object[MAXIMUM_STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
```
Dimagaa, line 21: Don't use class Objects - it is required!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
Dimagaa, line 35: The same - don't use class Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
+                    values[i] = value;
+                }
+            }
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
Dimagaa, line 28: What else size > keys.length and values.length? It is better to throw a new RuntimeException with an informative message!

Dimagaa (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
satori11111: Should we throw IndexOutOfBounds or Runtime would be enough?

Dimagaa (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAXIMUM_STORAGE_SIZE];
+        values = (V[]) new Object[MAXIMUM_STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if ((keys[i] == null && keys[i] == key)
```
Vasul-Pankiv, line 20: Redundant check for null
```suggestion
                if ((keys[i] == key)
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+                        || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                }
+            }
+        } else if (size <= keys.length && size <= values.length) {
```
Vasul-Pankiv, line 25: Its enough to check length only for keys or value
```suggestion
        } else if (size <= keys.length) {
```

Vasul-Pankiv (changes requested):

Vasul-Pankiv (approved):
Good job!

PR #1150, opened 2023-07-06
StorageImplFunctions

ConfusedSatlan:


DmytroV95 (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    private int size;
+
+    public Pair<K, V>[] increaseArray(Pair<K, V>[] oldArray) {
+        int newSize = size * 2;
+        Pair<K, V>[] newArray = new Pair[newSize];
+        for (int i = 0; i < size; ++i) {
+            newArray[i] = oldArray[i];
+        }
+        return newArray;
+    }
```
nacenik, line 18: do you really need to increase capacity?? 
it should be maximum 10 elements 
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (size == storageArray.length) {
+            storageArray = increaseArray(storageArray);
+        }
+        for (int i = 0; i < size; ++i) {
+            if (key != null && storageArray[i].getKey() != null) {
+                if (storageArray[i].getKey().equals(key)) {
```
nacenik, line 27: maybe better to merge and put into separate method these two conditions to reuse them?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
     public V get(K key) {
+        for (int i = 0; i < size; ++i) {
+            if (key != null && storageArray[i].getKey() != null) {
+                if (storageArray[i].getKey().equals(key)) {
+                    return storageArray[i].getValue();
+                }
+            } else if ((key == null && storageArray[i].getKey() == null)
+                    && storageArray[i].getValue() != null) {
+                return storageArray[i].getValue();
+            }
```
nacenik, line 50: you can simplify here using only one if

nacenik (changes requested):
Good solution, let's improve it)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
nacenik, line 6: redundant line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    public static final int DEFAULT_SIZE_PAIR = 10;
```
nacenik, line 7: is it default size or max size?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private Pair<K, V>[] storageArray;
+    private int size;
+
+    public StorageImpl() {
+        storageArray = new Pair[MAX_SIZE_PAIR];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == storageArray.length) {
```
KseniiaMakarova, line 16:
```suggestion
        if (size == MAX_SIZE_PAIR) {
```
just easier to read

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            throw new ArrayIndexOutOfBoundsException(
+                    "Max capacity of array is 10! You cant add more items");
+        }
+        for (int i = 0; i < size; ++i) {
+            if (equals(key, storageArray[i].getKey())) {
+                storageArray[i] = new Pair<>(key, value);
+                return;
+            } else if (key == null && storageArray[i].getKey() == null) {
```
KseniiaMakarova, line 24: we don't need else after return

and also you don't need this check, because if they are both null - your `equals()` method will return true

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean equals(K key, K keySecond) {
```
KseniiaMakarova, line 48:
```suggestion
    private boolean equals(K key, K secondKey) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE_PAIR = 10;
```
```suggestion
    public static final int MAX_SIZE = 10;
```

KseniiaMakarova (changes requested):

PR #1151, opened 2023-07-06
work done

mate-academy-student-2023:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            pair.setValue(value);
+            return;
+        }
+        if (size >= MAX_SIZE) {
+            throw new RuntimeException("This storage can store up to " + MAX_SIZE + " elements");
+        }
+        pairs[size++] = new Pair<>(key, value);
+    }
+
+    private Pair<K, V> find(K key) {
```
```suggestion
    private Pair<K, V> getPairByKey(K key) {
```

okuzan (approved):
Good job

src/main/java/core/basesyntax/impl/StorageImpl.java
okuzan: private method should go after all the public ones

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    // we should be OK here to have it suppressed
+    // since Pair instances are not shared with outside world
+    @SuppressWarnings("unchecked")
+    private final Pair<K, V>[] pairs = new Pair[MAX_SIZE];
```
Ivan95kos, line 11: For the future, it is better to use constructor for such filed
```suggestion
    private final Pair<K, V>[] pairs;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    // we should be OK here to have it suppressed
```
Ivan95kos, line 8: Thanks for the comments, but the code should be clear without them

Ivan95kos (approved):

PR #1152, opened 2023-07-07
implemented generic storageImpl class

Den-moroz:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        }
+        return values[index];
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int contains(K key) {
```
andrii-hoienko, line 46: method "contains" from its name should return true or false, but this method returns position of key in your storage, please rename this method

andrii-hoienko, line 46: What if key will be null? Is null valid value to store in storage?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = contains(key);
+        if (index == -1) {
+            if (size < MAX_SIZE) {
```
andrii-hoienko, line 20: it will be better to not make if inside of if. I would prefere to make 2 not nested if statements, first will be check for max size and if max size reached throw exception. Second one will check for index

andrii-hoienko (changes requested):
Good, but some improvenets required

src/main/java/core/basesyntax/impl/StorageImpl.java
Den-moroz: I tried to add a check for null, but an error occurs in the tests because null can be stored in storage

pom.xml
```diff
@@ -25,6 +25,24 @@
             <version>4.12</version>
```
Vasul-Pankiv, line 25: why you push your pom? revert your commit

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOfKey(key);
+        if (size > MAX_SIZE) {
+            throw new RuntimeException("Storage size is full");
+        }
```
Vasul-Pankiv, line 21: size == MAX_SIZE is invalid too
also, what if storage full, but you need to replace value for some key?
move your check to the rigth place

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        return values[index];
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOfKey(K key) {
```
Vasul-Pankiv, line 45: better rename it to getIndex or getIndexByKey()

Vasul-Pankiv (changes requested):

KseniiaMakarova (approved):

PR #1156, opened 2023-07-08
Implement storage

MurphyEdward:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K[] keyDatabase;
+    private V[] valueDatabase;
```
Ivan95kos, line 8: We don't use the database
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
     public void put(K key, V value) {
+        if (doesKeyExist(key)) {
+            replaceValue(key, value);
+            return;
+        }
+
+        addValue(key, value);
+    }
+
+    private boolean doesKeyExist(K key) {
```
Ivan95kos, line 26: All private methods should be below public methods

Ivan95kos (changes requested):
Good job overall!

olekskov (approved):
Good job

PR #1157, opened 2023-07-08
Implement put(K key, V value), get(K key) methods in StorageImpl class.

Dimagaa:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+        return pair == null ? null : pair.value;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private static class Pair<K, V> {
```
Vasul-Pankiv, line 39: Make it non static

Vasul-Pankiv (changes requested):
Great solution!!!

PR #1159, opened 2023-07-09
implemented the Storage of key-value type

ArtemReizer:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private final boolean[] occupied;
```
Den-moroz, line 9: you don't need these variable to solve this problem

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        occupied = new boolean[MAX_SIZE];
+        size = 0;
```
Den-moroz, line 16: useless lines, size is 0 by default and occupied is useless variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
+            values[index] = value;
+        } else {
+            if (size >= MAX_SIZE) {
+                throw new RuntimeException("Storage is full");
+            }
+            index = findNextAvailableIndex();
+            keys[index] = key;
+            values[index] = value;
+            occupied[index] = true;
+            size++;
```
Den-moroz, line 32: incorrect assignment variables. Will be enough to use only size.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            return values[index];
+        }
         return null;
```
Den-moroz, line 42: it would be better if you were first ckecked on the  (index == -1) and return null;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
+    }
+
+    private int findNextAvailableIndex() {
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (!occupied[i]) {
+                return i;
+            }
+        }
         return -1;
     }
```
Den-moroz, line 66: you can delete it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,65 @@
...
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (occupied[i] && (keys[i] == null ? key == null : keys[i].equals(key))) {
```
Den-moroz, line 52: change your condition taking into account new changes

Den-moroz (changes requested):
Let's improve)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -6,40 +6,35 @@ public class StorageImpl<K, V> implements Storage<K, V> {
...
-        if (index != -1) {
-            values[index] = value;
+        if (index == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
         } else {
             if (size >= MAX_SIZE) {
                 throw new RuntimeException("Storage is full");
             }
```
Den-moroz: First check for MAX_SIZE condition and then initialization the value because RuntimeExcption may occur and why do you use 'else' first and then 'if' if you can use 'else if'

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -49,16 +44,7 @@ public int size() {
...
-        return -1;
-    }
-
-    private int findNextAvailableIndex() {
-        for (int i = 0; i < MAX_SIZE; i++) {
-            if (!occupied[i]) {
+            if ((key != null && key.equals(keys[i])) || (key == keys[i])) {
                 return i;
             }
         }
```
Den-moroz: u forgot to return -1 if you don't find the key, also in the condition ```(key == keys[i])``` reduand brackets

Den-moroz (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ArtemReizer: in the line 51 i return -1. isn't it right?

Den-moroz: Sorry, I didn't see

ArtemReizer: never mind)) Thanks a lot!

Den-moroz (commented):
Looks nice. Good job!

Den-moroz (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (size > MAX_SIZE) {
+            throw new RuntimeException("Storage is full");
+        }
```
Vasul-Pankiv, line 21: what if index will be 3?
Move size check to right place

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || key == keys[i]) {
```
KseniiaMakarova, line 47:
```suggestion
            if (key != null && key.equals(keys[i]) || key == keys[i]) {
```

KseniiaMakarova (approved):

PR #1160, opened 2023-07-10
Implemented StorageImpl<K, V> class with methods

Someboty:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[10];
+        values = (V[]) new Object[10];
```
padu12, line 12: 10 is a magic number. Put it in constants

padu12 (commented):
A good solution! But there is still one small mistake. Hope you make it fast 🚀

padu12 (changes requested):
.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[ARRAY_MAX_SIZE];
+        values = (V[]) new Object[ARRAY_MAX_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || (key == null && keys[i] == null)) {
```
Vasul-Pankiv, line 20: you can simplify it
```suggestion
            if (keys[i] == key || (key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
Vasul-Pankiv, line 33: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || (key == null && keys[i] == null)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
```
Vasul-Pankiv, line 25: what if size > 10? 
Check it and throw some exception

Vasul-Pankiv (changes requested):

PR #1161, opened 2023-07-10
Storage of the key-value type, which is parameterized by two data typ…

OlegStashkiv:
…es, is implemented

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -4,6 +4,7 @@ import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     private static final int MAX_SIZE = 10;
+
     private K[] keys;
     private V[] values;
     private int size = 0;
@@ -15,7 +16,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     @Override
     public void put(K key, V value) {
-        int index = contains(key);
+        int index = getIndex(key);
         if (index != -1) {
             values[index] = value;
             return;
@@ -31,7 +32,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     @Override
     public V get(K key) {
-        int index = contains(key);
+        int index = getIndex(key);
         if (index != -1) {
             return values[index];
         }
@@ -43,7 +44,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
         return size;
     }
 
-    private int contains(K key) {
+    private int getIndex(K key) {
         for (int i = 0; i < size; i++) {
             if ((key != null && key.equals(keys[i])) || (key == keys[i])) {
                 return i;
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,52 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int contains(K key) {
```
BlueVioletTeti, line 46: Function name doesn't represent its activity. Something like getIndex() or getKeyIndex() might be a better choice

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
```
BlueVioletTeti, line 6: there should be an extra line between constants and other field variables

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        return size;
+    }
+
+    private int contains(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || (key == keys[i])) {
+                return i;
+            }
+        }
         return -1;
```
BlueVioletTeti, line 52: Why don't you use here null instead of -1? I believe it would make code above more understandable

BlueVioletTeti (changes requested):
Nice job, it was quite insightful for me :)

src/main/java/core/basesyntax/impl/StorageImpl.java
OlegStashkiv: It cannot be used here null. Since the return type int.

BlueVioletTeti (approved):
Good!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    private K[] keys;
+    private V[] values;
```
Vasul-Pankiv, line 9: make it final

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private int size = 0;
```
Vasul-Pankiv, line 10: it is 0 by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
Vasul-Pankiv: -1 is correct

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || (key == keys[i])) {
```
KseniiaMakarova, line 49:
```suggestion
            if (key != null && key.equals(keys[i]) || key == keys[i]) {
```

KseniiaMakarova (approved):

PR #1165, opened 2023-07-10
added an implementation of the main task

ivan0dyatlyukkk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return positionOfLastPutting;
+    }
+
+    public int containsKey(K key) {
```
keda-vlad, line 44: why its public, do u use it enyvere else? Also rename to 'indexOfKey' or 'positionOfKey'.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_STORAGE_SIZE; i++) {
+            if (containsKey(key) != -1) {
+                int position = containsKey(key);
+                return valueStorage[position];
+            }
+        }
```
keda-vlad, line 35: 'position' should be before 'if' block, and used it in if check condition. This will allow the method 'containsKey' to be called only 1n times.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+
+    public int containsKey(K key) {
+        for (int i = 0; i < MAX_STORAGE_SIZE; i++) {
+            if (key != null && key.equals(keyStorage[i])) {
+                return i;
+            }
+            if (key == keyStorage[i] && valueStorage[i] == null) {
+                positionOfLastPutting++;
+                return i;
+            }
```
keda-vlad, line 52: Do u really need it? For what?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final K[] keyStorage;
+    private final V[] valueStorage;
+    private int positionOfLastPutting = 0;
```
keda-vlad, line 9: 'Common mistakes' - All non-constant fields should be initialized in the constructor.
And also positionOfLastPutting will be initialized 0 by default.

keda-vlad (changes requested):
Not bad, but you need to fix some moments

src/main/java/core/basesyntax/impl/StorageImpl.java
ivan0dyatlyukkk: In my case, I used two arrays, so for getting the correct result of adding and setting a new value.
In case, when a new key equals null and a value doesn't, I should set those data in the arrays correctly.
Perhaps, it's not the best idea (mine)

keda-vlad: In this case u get true when key and keyStorage[i] are null bouth(or have the same referense) and valueStorage[i] is null. So u just find emty index for bouth arrays, why after this u need 'positionOfLastPutting++'?

keda-vlad: In this case u get true when key and keyStorage[i] are null bouth(or have the same referense) and valueStorage[i] is null. So u just find emty index for bouth arrays, why after this u need 'positionOfLastPutting++'?

keda-vlad (approved):

IvanFadieiev (approved):
Looking good! Nice solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return positionOfLastPutting;
+    }
+
+    private int indexOfKey(K key) {
```
Vasul-Pankiv, line 44: better to name this method like getIndexByKey or just getIndex

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private final K[] keyStorage;
+    private final V[] valueStorage;
+    private int positionOfLastPutting;
```
Vasul-Pankiv, line 9: this field represents size of your storage, so better to name it like that)

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
ivan0dyatlyukkk: Okay, thanks for the review! I will fix it after the lesson :)

Vasul-Pankiv (approved):

PR #1166, opened 2023-07-10
done solution

DaniilNepelev:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findKey(key);
+
+        if (index != -1) {
+            values[index] = value;
```
JeniaSan, line 21: You can add empty return , and after that you don't need an "else" block

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        } else {
+            if (size >= MAX_SIZE) {
+                throw new IllegalStateException("Storage is full");
```
JeniaSan, line 24: You can make single method like this 
```suggestion
private boolean isFull() {
            if (size >= MAX_SIZE) {
                throw new IllegalStateException("Storage is full");
                }
                return false;
                }
                
```
and call this method in "if" block. If your storage is full , you throw exception. But if you return false , your code on 27 - 29 lines will work

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        int index = findKey(key);
+
+        if (index != -1) {
+            return values[index];
+        } else {
+            return null;
```
JeniaSan, line 40: Do you really need "else" ? You can write only return

JeniaSan (changes requested):
Nice work, but you can simplify your code

src/main/java/core/basesyntax/impl/StorageImpl.java
DaniilNepelev: Okay, I will improve my code by your recomenedations

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findKey(key);
+        if (index != -1 || isFull()) {
+            values[index] = value;
+            return;
```
JeniaSan, line 21: in this method you need not one "if" block , because you have to cases with different
 action :
 1) You have key, and change value ;
 2) You have a new pair(key, value) and need to add in your storage ;
 Your code should be something like that : 
 
```suggestion
        int index = findKey(key);
        if (index != -1) {
            values[index] = value;
            return;
         }
         if (!isFull()) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+        }
+
```
JeniaSan, line 23: redundant empty line again

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
     }
 
     @Override
     public V get(K key) {
+        int index = findKey(key);
+
+        if (index != -1) {
+            return values[index];
+        }
+
```
JeniaSan, line 36: redundant empty line again

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean isFull() {
+        if (size >= MAX_SIZE) {
+            throw new IllegalStateException("Storage is full");
+        } else {
```
JeniaSan, line 57: you don't need else , just " return false "

JeniaSan (approved):
good work

OleksandrSaven:
Looks good

OleksandrSaven (approved):
Looks good.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKey(K key) {
```
Vasul-Pankiv, line 42: It return key or key index?
Better to rename it to getIndex or getIndexByKey

KseniiaMakarova (approved):

PR #1168, opened 2023-07-10
My solution

Zel1oy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                }
+            }
+            return;
+        }
```
Vasul-Pankiv, line 25: move return iside if block
```suggestion
        if (get(key) != null) {
            for (int i = 0; i < size; i++) {
                if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
                    values[i] = value;
                    return;
                }
            }
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_ARRAY = 10;
+    private int size = 0;
+    private final K[] keys;
+    private final V[] values;
```
Vasul-Pankiv, line 9: First final than non fina, also size is 0 by default
```suggestion
    private static final int MAX_SIZE_ARRAY = 10;
    private final K[] keys;
    private final V[] values;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                }
+            }
+            return;
+        }
+        if (size == MAX_SIZE_ARRAY) {
```
Vasul-Pankiv, line 26: better to use >=
```suggestion
        if (size >= MAX_SIZE_ARRAY) {
```

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: This comment is not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_ARRAY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
andrii-hoienko, line 9: size is 0 by default.

andrii-hoienko (changes requested):
Please be more cerafull to mentor comments

andrii-hoienko (approved):
Good job!

PR #1170, opened 2023-07-10
added solution with generics K and V

AntonZhdanov:


OlegTuryk (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_NUMBER_OF_ELEMENTS = 10;
+    private int size;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_NUMBER_OF_ELEMENTS];
+        values = (V[]) new Object[MAX_NUMBER_OF_ELEMENTS];
+        size = 0;
```
nacenik, line 14: `size` is 0 by default 
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                }
+            }
```
nacenik, line 24: it's a good idea to create the method `indexOf()` which returns the index of the searched element if it exists, or -1 if it's a new element. and you don't need to have 2 arrays in the row (your method calls get, and if it returns not null - you go through every element one more time to find an index)

nacenik (changes requested):
good solution, let's improve it a little bit :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -11,32 +11,27 @@ public class StorageImpl<K, V> implements Storage<K, V> {
...
-        } else if (size < values.length) {
+        int index = indexOfKey(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
             keys[size] = key;
             values[size] = value;
             size++;
-        } else {
-            throw new RuntimeException("Can't put the file into storage" + key + value);
```
Vasul-Pankiv: Why you delete this?
Just say in message that storage is full
Also add size check befor adding new value in your storage

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Can't put the file into storage, that storage is full");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
KseniiaMakarova, line 33: why loop here?

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #1172, opened 2023-07-10
Implement Storage

NikitaKryvenets:


Skarbey (commented):
Seems everything right. Tests passed.

Skarbey (approved):

Vladiclevch (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private int size = 0;
+    private final K[] keys;
+    private final V[] values;
```
Vasul-Pankiv, line 9: final fields should be before non finals
also size will be 0 by default
```suggestion
    private final K[] keys;
    private final V[] values;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (size > MAX_STORAGE_SIZE) {
+            throw new RuntimeException("Storage space is too full!");
+        }
```
Vasul-Pankiv, line 26: better to check it before loop

Vasul-Pankiv (changes requested):

Vasul-Pankiv (approved):

PR #1178, opened 2023-07-10
Implemented storage key-value class

CodeWebWeaver:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (get(key) != null) {
+            for (int i = 0; i < size; i++) {
+                if ((keys[i] == key)
+                        || (keys[i] != null && keys[i].equals(key))) {
+                    values[i] = value;
+                }
+            }
```
irynamekh, line 24: here the logic is repeated, the same as in the get() method. Try to move the repeating logic into a separate private method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        } else {
+            throw new RuntimeException("Storage is already filled. Can not add new elements");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key)
+                    || (keys[i] != null && keys[i].equals(key))) {
```
irynamekh, line 38: everything is correct, but I prefer to use Objects.equals(), then the code looks clearer. This is just an idea, you can not change anything here.

irynamekh (changes requested):
if my fixes break something, please tell me

src/main/java/core/basesyntax/impl/StorageImpl.java
Skarbey: Thanks. I agree but in common misstakes Objects are restricted

irynamekh (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAXIMUM_ARRAY_SIZE = 10;
```
```suggestion
    public static final int MAXIMUM_STORAGE_SIZE = 10;
```

nacenik (changes requested):
Good solution, let's improve it)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getIndexByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == key) || (keys[i] != null && keys[i].equals(key))) {
```
KseniiaMakarova, line 43:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

KseniiaMakarova (approved):

PR #1182, opened 2023-07-11
implemented put, size and get methods for storage

Pafmortim:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        values = (V[]) new Object[STORAGE_LIMIT];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index == -1) {
+            keys[currentStorageSize] = key;
+            values[currentStorageSize] = value;
+            currentStorageSize++;
```
IhorKalaur, line 22: What if currentSize will be equal STORAGE_LIMIT. We can get IndexOutOfBoundsException. Let's let's predict it and throw new Exception.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     }
 
     @Override
     public int size() {
+        return currentStorageSize;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < currentStorageSize; i++) {
+            if (key != null && key.equals(keys[i]) || (key == keys[i])) {
```
IhorKalaur, line 44: I'm not sure, but maybe makes sense to check first (key == keys[i]). It's will be less resource-intensive to compare their addresses in memory, and only then to use equals() method.

IhorKalaur (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_LIMIT = 10;
+    private static final String OUT_OF_LIMIT_EXCEPTION = "The storage is full";
+    private final K[] keys;
+    private final V[] values;
+    private int currentStorageSize = 0;
```
Vasul-Pankiv, line 10: Better to name it size
```suggestion
    private int size = 0;
```

Vasul-Pankiv (approved):

PR #1193, opened 2023-07-29
Implemented a key-value storage type that is parameterized by two dat…

mhorianin:
…a types.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,41 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null || key != null && key.equals(keys[i])) {
```
olekskov, line 19: Condition may be simplified:
```suggestion
            if (keys[i] == key || key != null && key.equals(keys[i])) {
```

olekskov (approved):
Good job, approved.
Please check comment below for further

PR #1195, opened 2023-08-02
solved task "Storage" in module "Java Core/Generics"

Krasnov-Maksim:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+            if (key == null) {
+                if (keys[i] == null) {
+                    return i;
+                }
+            } else {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
```
aaananas, line 58:
```suggestion
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || key != null && key.equals(keys[i])) {
                return i;
            }
        }
```

aaananas (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAXIMUM_NUMBER_OF_ELEMENTS = 10;
+    private static final int NOT_EXISTING_INDEX = -1;
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
Rommelua, line 10:
```suggestion
    private int size;
```

Rommelua (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
Krasnov-Maksim: understand, thanks.

PR #1206, opened 2023-08-10
Jv storage solution

gordgelin:


olekskov (approved):
I'd say that solution with 2 arrays would be more readable due to absence of lots of *2 and /2 but this solution is fully valid. Approved

gordgelin:
> I'd say that solution with 2 arrays would be more readable due to absence of lots of *2 and /2 but this solution is fully valid. Approved

completely agree but since task description suggested both approaches as possible i decided to do via one array to have some fun :)

PR #1210, opened 2023-08-11
implemented StorageImpl class and added Pair class

ystankevych:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
 package core.basesyntax.impl;
 
+import core.basesyntax.Pair;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private final Pair[] pairs;
```
Oleksandr-Tymoshenko, line 8: you can create generic array and return data without converting it to the desired type using (V)
```suggestion
    private final Pair<K, V>[] pairs;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+            } else {
+                pairs[++index] = new Pair<>(key, value);
+            }
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return getKeyIndex(key) >= 0 ? (V) pairs[getKeyIndex(key)].getValue() :
```
Oleksandr-Tymoshenko, line 30: you are calling getKeyIndex twice, it goes throw the whole cycle twice, lets call this method only once

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private final Pair[] pairs;
+    private int index;
+
+    public StorageImpl() {
+        pairs = new Pair[STORAGE_CAPACITY];
+        index = -1;
```
Oleksandr-Tymoshenko, line 13: maybe it is better to initializate it with 0 and it will be easier to work with it, because you are adding 1 in size() mthod, you added equals in cycles.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
 
     @Override
     public int size() {
+        return index + 1;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i <= index; i++) {
+            if ((pairs[i].getKey() != null && pairs[i].getKey().equals(key))
+                    || (pairs[i].getKey() == null && key == null)) {
```
Oleksandr-Tymoshenko, line 42:
```suggestion
            if ((key == pairs[i].getKey()) || (pairs[i].getKey() != null && pairs[i].getKey().equals(key))) {
```
also, you are calling getKey() 3 times, try to fix it

Oleksandr-Tymoshenko (changes requested):
Good job, lets improve your solution!

liliia-ponomarenko (changes requested):
Student reject

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,54 @@
...
     @Override
     public int size() {
+        return index;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size(); i++) {
+            K current = pairs[i].getKey();
+            if ((current != null && current.equals(key))
+                    || (current == null && key == null)) {
```
Oleksandr-Tymoshenko, line 44: it can by simplified like
```suggestion
                    || (current == key)) {
```

Oleksandr-Tymoshenko (approved):
Nice!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,54 @@
...
+        index = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (isFull()) {
+            return;
+        }
+        int keyIndex = getKeyIndex(key);
+        if (keyIndex >= 0) {
```
liliia-ponomarenko, line 22:
```suggestion
        if (keyIndex != -1) {
```

liliia-ponomarenko (approved):
Great 😊

PR #1211, opened 2023-08-13
JV Storage impl

Oleksandr-Tymoshenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_ELEMENTS_COUNT];
+        values = new Object[MAX_ELEMENTS_COUNT];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            System.out.println(keys[i] + " " + key);
```
RedArtelerist, line 19: Why we need this println

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+            }
+        }
+        keys[size] = key;
+        values[size++] = value;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == keys[i]) || (keys[i] != null && keys[i].equals(key))) {
```
RedArtelerist, line 32: duplicated code, you can move the index finding into a separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_COUNT = 10;
+    private int size = 0;
```
RedArtelerist, line 7: It's better to initialize this in the constructor

RedArtelerist (changes requested):
Try to think how you can implement using just one array instead of two Object[]

src/main/java/core/basesyntax/impl/StorageImpl.java
RedArtelerist, line 7: It's better to initialize this in the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
RedArtelerist, line 19: Why we need this println

src/main/java/core/basesyntax/impl/StorageImpl.java
RedArtelerist, line 32: duplicated code, you can move the index finding into a separate method

RedArtelerist (approved):
Looks great

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     public V get(K key) {
-        return null;
+        int i = getIndex(key);
+        return i != -1 ? items[i].getValue() : null;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            K currentKey = items[i].getKey();
+            if ((key == currentKey) || (currentKey != null && currentKey.equals(key))) {
```
liliia-ponomarenko, line 34:
```suggestion
            if (key == currentKey || currentKey != null && currentKey.equals(key)) {
```
Redundant  brackets

src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax.impl;
+
+class Pair<K, V> {
+    private K key;
+    private V value;
+
+    //write your code here
```
liliia-ponomarenko, line 7: Let’s  remove this comment

liliia-ponomarenko (approved):
Good job! Minor mistakes;)

PR #1213, opened 2023-08-15
implementation of storage

sonyasapsan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                currentIndex++;
+            } else {
+                array[index][VALUE_INDEX] = value;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentIndex; i++) {
```
ystankevych, line 30: Look, you have same code on line 30-32 and 45-47. You just can call findKeyIndex here and if index -1 return null otherwise  return value by index =)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int AMOUNT_OF_OBJECTS = 2;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private int currentIndex = 0;
+
+    private Object[][] array = new Object[MAX_SIZE][AMOUNT_OF_OBJECTS];
```
ystankevych, line 12:
```suggestion
    private Object[][] array;
    
    public StorageImpl() {
        array = new Object[MAX_SIZE][AMOUNT_OF_OBJECTS];
    }
```

maybe it would be better to create array[amount_of_objects][max_size]. In this case you create only two array with 10 elements each. And in your case - 10 array with 2 elements. But its up to you and you could or could not change it =)

ystankevych (changes requested):
Realy great solution! I left some minor comments =)

src/main/java/core/basesyntax/impl/StorageImpl.java
sonyasapsan: I'm not sure, but wouldn't that change the logic of the program? Or maybe I understood something wrong (Anyway, I shortened the code via the boolean function.

ystankevych (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int AMOUNT_OF_OBJECTS = 2;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private int currentIndex = 0;
```
liliia-ponomarenko, line 10:
```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+                currentIndex++;
+            } else {
+                array[index][VALUE_INDEX] = value;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentIndex; i++) {
```
liliia-ponomarenko, line 34: agree it duplicates `findKeyIndex` method. Let’s simplify it.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+            if (isSameKey(key, i)) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean isSameKey(K key, int index) {
+        return (key == null && array[index][KEY_INDEX] == null)
+                || (key != null && key.equals(array[index][KEY_INDEX]));
```
liliia-ponomarenko, line 58:
```suggestion
        return (key == array[index][KEY_INDEX]
                || key != null && key.equals(array[index][KEY_INDEX]);
```
can be simplified

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+
+    private Object[][] array;
+
+    public StorageImpl() {
+        array = new Object[MAX_SIZE][AMOUNT_OF_OBJECTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (currentIndex < MAX_SIZE) {
```
liliia-ponomarenko, line 20: Try to avoid such constructions. You can stop the method if the index bigger than the max size  
`if (currentIndex >= MAX_SIZE) {
  return;
}`

liliia-ponomarenko (changes requested):
Interesting solution but let’s improve some moments ;)

liliia-ponomarenko (changes requested):
One more thing. Now the array is not parameterized, so if the code doesn't work correctly keys can be stored as values and values as keys. This is bad, so I advise you to still use separate arrays with parameterized values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_SIZE) {
+            return;
+        }
+        int index = findKeyIndex(key);
+        if (index == -1) {
+            keys[this.size] = key;
```
Andriy-dz, line 23: you can remove the this keyword

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean isSameKey(K key, int index) {
+        return (key == keys[index])
+                || key != null && key.equals(keys[index]);
+    }
```
Andriy-dz, line 54: you can do without this method since you only use it once

Andriy-dz (approved):

PR #1214, opened 2023-08-15
first version

Alex-zzzz:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,27 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private final Map<K,V> states = new HashMap<K, V>();
```
overpathz, line 9: image

Actually, a great solution, but in this task we have to use a different approach :)
Consider using 2 arrays. One for keys and one for values and use some pointer(index)/size to manage adding and removing elements synchronously.

overpathz (changes requested):
Please, use a different approach to implement this task. Thanks!

Ivan95kos (changes requested):
Please, pay attention to common mistakes:
https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
okuzan, line 6: remove this line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        values = (V[]) new Object[MAX_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean isNewKey = true;
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                //size++;
```
okuzan, line 23: usually, there shouldn't be comments in your submitted code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+                isNewKey = false;
+                break;
+
+            } else if (keys[i] == null && keys[i] == (key)) {
+                values[i] = value;
+                isNewKey = false;
+                //size++;
+                break;
+            }
```
okuzan, line 32: you can see that the body of your two clauses is the same, this means you should combine the logical conditions to 
key == keys[i]) || key != null && key.equals(keys[i])

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == (key)) {
+                return values[i];
+            }
+            if (keys[i] != null && !(keys[i].getClass().isPrimitive())) {
+                if (keys[i].equals(key)) {
+                    return values[i];
+                }
+            }
+        }
```
okuzan, line 53: same here

okuzan (changes requested):
You have the same logic in both put and get method - given Key you try to find its index, It would be a good idea to extract it to a private method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+                break;
+
+            }
+        }
+        if (isNewKey) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
+
```
Ivan95kos, line 32:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                values[i] = value;
                return;
            }
        }
        
            keys[size] = key;
            values[size] = value;
            size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == (key) || (keys[i] != null && !(keys[i].getClass().isPrimitive())
+                    && keys[i].equals(key))) {
+                return values[i];
+            }
+        }
```
Ivan95kos, line 42: redundant verification
```suggestion
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
                return values[i];
            }
        }
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ELEMENTS];
+        values = (V[]) new Object[MAX_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean isNewKey = true;
```
Ivan95kos, line 18: useless variable
```suggestion
```

Ivan95kos (approved):

PR #1215, opened 2023-08-16
Jv storage

AndriiMarushchak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Pair<K, V>[] pairs;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        pairs = (Pair<K, V>[]) new Pair[10];
+        size = 0;
```
okuzan, line 12: no need for this, 0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Pair<K, V>[] pairs;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        pairs = (Pair<K, V>[]) new Pair[10];
```
okuzan, line 11: this should be class constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
...
+        return obj1.equals(obj2);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void increaseCapacity() {
+        int newCapacity = pairs.length * 2;
+        Pair<K, V>[] newPairs = (Pair<K, V>[]) new Pair[newCapacity];
+        System.arraycopy(pairs, 0, newPairs, 0, size);
+        pairs = newPairs;
+    }
```
okuzan, line 58: this is out of scope for the task, we don't need to handle growing here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
...
+            if (compareKeys(key, pairs[i].getKey())) {
+                pairs[i].setValue(value);
+                return;
+            }
+        }
+        if (size == pairs.length) {
+            increaseCapacity();
+        }
+        pairs[size] = new Pair<>(key, value);
+        size++;
```
```suggestion
        pairs[size++] = new Pair<>(key, value);
```
Ok as is, only FYI

okuzan (changes requested):
Good job overall!

okuzan (changes requested):
Please go thru all the comments before requesting a review

Ivan95kos (approved):

PR #1216, opened 2023-08-16
was implemented StorageImpl class

Vt0rtikV:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int AMOUNT_OF_ARRAYS = 2;
+    private static final int AMOUNT_OF_ITEMS = 10;
+    private static final int INDEX_OF_KEYS = 0;
+    private static final int INDEX_OF_VALUES = 1;
+    private Object[][] dictionary = new Object[AMOUNT_OF_ARRAYS][AMOUNT_OF_ITEMS];
+    private int size = 0;
```
mmosii, line 11: All non-constant fields should be initialized in the constructor (from common mistakes:)
``
```suggestion
    private Object[][] dictionary;
    private int size;
    
    public StorageImpl() {
         dictionary = new Object[AMOUNT_OF_ARRAYS][AMOUNT_OF_ITEMS];
         size = 0;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int checkKey(K key) {
```
mmosii, line 40: i think it's more informative name of method
```suggestion
    private int getKeyIndex(K key) {
```

mmosii (changes requested):
Good job, let's try to improve it

mmosii (approved):

DenisGorda (approved):
I think you present a very cool solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = getKeyIndex(key);
+        if (index != -1) {
+            return (V) dictionary[INDEX_OF_VALUES][index];
+        }
         return null;
```
Andriy-dz, line 37: you can make it a ternary operator

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (dictionary[INDEX_OF_KEYS][i] == key
+                    || key != null
+                    && key.equals(dictionary[INDEX_OF_KEYS][i])) {
```
Andriy-dz, line 49:
```suggestion
            if (key != null && key.equals(dictionary[INDEX_OF_KEYS][i])) {
```
I think this will be enough

Andriy-dz (approved):
Very good solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
Vt0rtikV: but if i have key null it wouldn't work correctly

PR #1217, opened 2023-08-16
Implemented StorageImpl class

mmosii:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] != null && keys[i].equals(key) || keys[i] == null && keys[i] == key)) {
```
Vt0rtikV, line 45: May be you can do it like this:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
Vt0rtikV, line 25: You can do it like this
```suggestion
            values[size++] = value;
```

Vt0rtikV (approved):
Good job)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1219, opened 2023-08-17
methods implemented

DenisGorda:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 9;
+    private K key;
+    private V value;
+    private int size = 0;
+    private StorageImpl[] storArr = new StorageImpl[MAX_SIZE];
+
+    public StorageImpl() {
+    }
```
Vt0rtikV, line 13: All non-constant fields should be initialized in the constructor (from common mistakes)
```suggestion
    private int size;
    private StorageImpl[] storArr;

    public StorageImpl() {
    size = 0;
    storArr = = new StorageImpl[MAX_SIZE];
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < storArr.length; i++) {
+            if ((storArr[i] != null && storArr[i].key == null) && key == null) {
+                storArr[i].value = value;
+                break;
+            } else if (storArr[i] != null && storArr[i].key != null && storArr[i].key.equals(key)) {
+                storArr[i].value = value;
+                break;
```
Vt0rtikV, line 39: i think it could be in one if, cause it's the same cases, where you are setting new value for a key that you already have.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+    private V value;
+    private int size = 0;
+    private StorageImpl[] storArr = new StorageImpl[MAX_SIZE];
+
+    public StorageImpl() {
+    }
+
+    public K getKey() {
+        return key;
+    }
```
Vt0rtikV, line 17: this method is never used

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+
+    public void setKey(K key) {
+        this.key = key;
+    }
+
+    public V getValue() {
+        return value;
```
Vt0rtikV, line 24: this method is never used

Vt0rtikV (changes requested):
Interesting solution. Let's try to improve it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
     public V get(K key) {
-        return null;
+        V res = null;
+        for (int i = 0; i < storArr.length; i++) {
+            if ((key == null
+                    && storArr[i] != null
+                    && storArr[i].key == null)
+                    || (storArr[i] != null
+                    && storArr[i].key != null
+                    && storArr[i].key.equals(key))) {
```
Vt0rtikV, line 60: you have the same code in your put method. It may be better, if it will one new method checkKey() or findIndex()
also I think your if may be like this: 
` K currentKey = storArr[i].getKey();
  if ((key == currentKey) || (currentKey != null && currentKey.equals(key))`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 9;
+    private K key;
+    private V value;
+    private int size = 0;
+    private StorageImpl[] storArr = new StorageImpl[MAX_SIZE];
```
RedArtelerist, line 10: Interesting solution, but I think it's not a good way to implement storage. Because your program will create many `StorageImpl` objects which are not used in any way. You in `StorageImpl` store an array of the same class which will contain empty arrays. It's not very good on the memory side. It's better to create two arrays of `Object[]` under keys and values.

RedArtelerist (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        V result = null;
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if ((key == null
+                    && values[i] != null
+                    && keys[i] == null)
+                    || (values[i] != null
+                    && keys[i] != null
+                    && keys[i].equals(key))) {
+                result = (V) values[i];
+                break;
```
RedArtelerist, line 44: This cycle is duplicated, because you also use it in `put `method, so you can move to a separate function finding the index by key. And then use it in get and put methods

RedArtelerist (changes requested):
It's better, but try to improve your solution a little

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        size = 0;
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size + 1; i++) {
+            if (((values[i] != null && keys[i] == null) && key == null)
+                    || (keys[i] != null && keys[i].equals(key))) {
```
Vt0rtikV, line 21: it's might be called getIndex. And don't check value, you can check only the key.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
```
Vt0rtikV, line 24: The same story as a comment above

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                keys[i] = key;
+                values[i] = value;
+                size++;
```
Vt0rtikV, line 27: you can take it out from cicle. It would be better

Vt0rtikV (changes requested):
Some more comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        if (index == -1) {
+            return null;
+        }
+        return (V) values[index];
```
Vt0rtikV, line 36: you can make it a ternary operator

Vt0rtikV (approved):
Great job)

src/main/java/core/basesyntax/impl/StorageImpl.java
Vt0rtikV, line 36:
```suggestion
   return index == -1 ? null : (V) values[index];
```

RedArtelerist (approved):
Cool

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 9;
+    private int size;
+    private Object[] keys;
+    private Object[] values;
```
Andriy-dz, line 9:
```suggestion
    private K[] keys;
    private V[] values;
```
if you do this, you won't need to casting

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        int index = getIndex(key);
+        if (index == -1) {
+            keys[size] = key;
+            values[size++] = value;
+        } else {
+            values[index] = value;
+        }
+
```
Andriy-dz, line 26: excta empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        return index == -1 ? null : (V) values[index];
+    }
+
+    private int getIndex(K key) {
```
Andriy-dz, line 35: let's sort the methods by access modifier

Andriy-dz (changes requested):
Good job! Lets improve your solution

TetianaDenchyk (approved):

PR #1225, opened 2023-08-20
Storage implement

don-bigdad:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        if (index == -1) {
+            return null;
+        }
+        return pairs[index].getValue();
```
VolodymyrSkvira, line 33: I would recommend using a ternary operator
```suggestion
        return index == -1 ? null :  pairs[index].getValue();
```

VolodymyrSkvira (changes requested):
Interesting solution, I personally used two arrays so I would prefer leaving this for the mentors eye

VolodymyrSkvira (approved):
LGTM, nice work

Andriy-dz (approved):
Very good solution!

PR #1227, opened 2023-08-21
implement methodes get, put, size

anigilator479:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private Object[] keysArray = new Object[MAX_ARRAY_SIZE];
+    private Object[] valuesArray = new Object[MAX_ARRAY_SIZE];
```
Alexsandr-Yablunovskyi, line 8: You may try to use only one array

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keysArray[i] == null && key == null) {
+                return (V) valuesArray[i];
+            }
+            if (keysArray[i] != null && keysArray[i].equals(key)) {
+                return (V) valuesArray[i];
+            }
```
Alexsandr-Yablunovskyi, line 32: Can you combine these conditions?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    public boolean checkKeysDublicate(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keysArray[i] != null && keysArray[i].equals(key)) {
+                valuesArray[i] = value;
+                return false;
+            }
+            if (keysArray[i] == null && key == null) {
+                valuesArray[i] = value;
+                return false;
+            }
```
Alexsandr-Yablunovskyi, line 51: Can you combine these conditions?

Alexsandr-Yablunovskyi (changes requested):
Nice work! Try to improve it a little bit.

src/main/java/core/basesyntax/impl/StorageImpl.java
anigilator479: Behind the condition of the task can be done through 2 arrays

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean checkKeysDublicate(K key, V value) {
```
Andriy-dz, line 40: uninformative name for the method, because it is clear from the name that it should check whether the given key exists, and here you check and change the value. uninformative name for the method, because it is clear from the name that it should check whether the given key exists, and here you check and change the value

Andriy-dz, line 40: and why public?)

Andriy-dz (changes requested):
Good job! Lets improve your solution)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean updateExistingKey(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keysArray[i] != null && keysArray[i].equals(key)
+                    || keysArray[i] == null && key == null) {
```
Andriy-dz, line 43:
```suggestion
            if (keysArray[i] != null && keysArray[i].equals(key)
                    || keysArray[i] == key) {
```
this is also suitable, but in this case you have a duplicate condition that can be transferred to a separate method

Andriy-dz (approved):

PR #1228, opened 2023-08-21
first commit

Lupen07:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKindex(K key) {
```
okuzan, line 49:
```suggestion
    private int getKeyIndex(K key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+            values[index] = value;
+        } else {
+            if (size >= MAX_SIZE) {
+                // Storage is full, cannot add more elements
+                throw new IllegalStateException("Storage is full");
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
okuzan, line 31: Good job, but handling storage overflow is out of the scope of this task :) Also, remove comments before committing 
```suggestion
        if (index >= 0) {
            values[index] = value;
        } else {
            keys[size] = key;
            values[size++] = value;
        }
```

okuzan (changes requested):

olekskov (approved):
Good job

PR #1230, opened 2023-08-22
Added impl for StorageImpl.java

SalnikovMarkiian:


validatorr (approved):
Great solution! You can try to implement the same with one array, but that's up to you.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        } else {
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndexByKey(key);
+        return (index != exitWithNull)
```
Andriy-dz, line 33:
```suggestion
        return index != exitWithNull
```

Andriy-dz (approved):
Good job

PR #1231, opened 2023-08-22
created StorageImpl

validatorr:


Cayman4010 (approved):
LGTM. May be, you need to make your arrays "keys" and "values" final, but I`m not sure.

validatorr:
> LGTM. May be, you need to make your arrays "keys" and "values" final, but I`m not sure.

I could be wrong, but if make them final, it will be impossible to change it's size. What do you think?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     public void put(K key, V value) {
+        int index = findIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size < MAX_CAPACITY) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            }
```
Andriy-dz, line 27: I would add some message to the user so that he understands that something went wrong when adding 11 elements

Andriy-dz (approved):
Good Job!

PR #1235, opened 2023-08-22
solution added

anastasiia-fi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    public int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
```
KDanil2000, line 41: I'd like this
`````
`suggestion
```
```

**Comment by KDanil2000** [comment]
An excellent solution!

File: src/main/java/core/basesyntax/impl/StorageImpl.java [hunk 2]
```diff
@@ -3,17 +3,50 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return keyExists(key) ? values[findKeyIndex(key)] : null;
```
Andriy-dz, line 31: let's use only the findKeyIndex method and if the result = -1 then return null

Andriy-dz (changes requested):

Andriy-dz (changes requested):
Good Work! lets improve it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return findKeyIndex(key) == -1 ? null : values[findKeyIndex(key)];
```
Andriy-dz, line 31: you call the method twice and pass the same parameter there. Let's better make a variable for the result of the method

Andriy-dz (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    public StorageImpl() {
+        keys = (K[])new Object[MAX_SIZE];
+        values = (V[])new Object[MAX_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (findKeyIndex(key) != -1) {
+            values[findKeyIndex(key)] = value;
```
NazarSmith228, line 21: no need to call this method twice

call it one time, store as a variable and use that variable in further

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key != null && key.equals(keys[i]) || key == keys[i]) {
```
NazarSmith228, line 42: 2 points:

1. think whether your implementation should support null keys. if yes - you can leave `key != null`. otherwise - remove it
2. no need to do reference equality check (`==`). you can get around with just `equals()` method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (findKeyIndex(key) != -1) {
+            values[findKeyIndex(key)] = value;
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
NazarSmith228, line 25: what will happen when `size` will exceed `MAX_SIZE`? correct - an exception

you should handle such scenario

NazarSmith228 (changes requested):
Great job! Now let's fix some remarks

src/main/java/core/basesyntax/impl/StorageImpl.java
anastasiia-fi: with this corrections it doesn't work
everything else I fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -17,12 +17,17 @@ public StorageImpl() {
 
     @Override
     public void put(K key, V value) {
-        if (findKeyIndex(key) != -1) {
-            values[findKeyIndex(key)] = value;
+        if (size == MAX_SIZE) {
+            return;
+        }
```
liliia-ponomarenko, line 22: now you don't have the option to update the value by key, although you should

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):
Good job!

PR #1236, opened 2023-08-22
implement task

romazan07:
implement task

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,47 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            keys[size] = key;
+            values[size] = value;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
```
OlehReex, line 20: You can create constant with MINIMAL_INDEX and change you check from `if (index != -1) {` to `if (keyIndex >= MIN_INDEX) {`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = findKeyIndex(key);
+        return index != -1 ? values[index] : null;
```
OlehReex, line 32: The same here. You cfn use constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
OlehReex, line 25: Try to avoid if-else construcrion. And think about what will happen if you try to put more than 10 elements in your storage?

OlehReex (changes requested):
Good solution. But here it's some things to improvement

Andriy-dz (commented):
Perfect

Andriy-dz (approved):
Perfect

PR #1238, opened 2023-08-22
Hw1

evgenkolner:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ATTEMPTS = 10;
+    private K [] keys;
+    private V [] values;
```
vsych321, line 8: you can make them also final

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ATTEMPTS];
+        values = (V[]) new Object[MAX_ATTEMPTS];
+    }
+
```
vsych321, line 15: you forgot to assign 0 to size value

vsych321 (approved):

vsych321 (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ATTEMPTS = 10;
```
```suggestion
    private static final int MAX_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ATTEMPTS];
+        values = (V[]) new Object[MAX_ATTEMPTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == MAX_ATTEMPTS) {
+            return;
+        }
```
liliia-ponomarenko, line 20: you should  move this check to the else block because  now you can’t  update the value if the array is full

liliia-ponomarenko (changes requested):
Everything fine! Two minor mistakes  ;)

liliia-ponomarenko (approved):
Good job!

PR #1240, opened 2023-08-22
done

RomanYuzviak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    private final Object[][] elements;
+
+    public StorageImpl() {
+        topElement = 0;
+        elements = new Object[MAX_SIZE][2];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < topElement; i++) {
```
liliia-ponomarenko, line 18: Let’s  move the logic of finding the index of the element to a separate private method and use it in `get` and `put `

liliia-ponomarenko (changes requested):
good job! Let’s improve your solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
 
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < topElement; i++) {
+            if (key == null) {
+                if (elements[i][0] == null) {
+                    return i;
+                }
+            } else if (key.equals(elements[i][0])) {
+                return i;
+            }
```
liliia-ponomarenko, line 45:
```suggestion
            if (key == elements[i][0] ||  key != null && key.equals(elements[i][0])) {
                return i;
            }
```
Let’s  combine these conditionals and simplify  your solution

liliia-ponomarenko (changes requested):
Great! But  let’s  fix one moment ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1243, opened 2023-08-22
Implemented a `Storage' of type `key-value', which is parameterized b…

pastcwdt:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+import core.basesyntax.Pair;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_LENGTH = 10;
+    private final Pair<K, V>[] pairs;
+    private int index;
+
+    public StorageImpl() {
+
```
liliia-ponomarenko, line 12: Redundant  empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+        pairs = new Pair[STORAGE_LENGTH];
+        index = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (isFull()) {
+            return;
+        }
```
liliia-ponomarenko, line 21: you should move this check to the else block because now you can’t update the value if the array is full

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_LENGTH = 10;
+    private final Pair<K, V>[] pairs;
+    private int index;
+
+    public StorageImpl() {
+
+        pairs = new Pair[STORAGE_LENGTH];
+        index = 0;
```
liliia-ponomarenko, line 14:
```suggestion
```
0 - default value for int

liliia-ponomarenko (changes requested):
Everything fine! Several minor mistakes ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
pastcwdt: didn't know that! thx

liliia-ponomarenko (approved):
Great!

PR #1244, opened 2023-08-23
imlemented storage project

yevhenkukharchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
-}
+package core.basesyntax.impl;
+
+import core.basesyntax.Storage;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_LENGTH = 10;
+    private K [] keys;
+    private V [] values;
+    private int size = 0;
```
liliia-ponomarenko, line 9:
```suggestion
    private int size;
```
0 - default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object [MAX_LENGTH];
+        values = (V[]) new Object [MAX_LENGTH];;
+    }
+
+    public int getKeyIndex(K key) {
```
liliia-ponomarenko, line 16: this method should  be private and move it after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+            }
+        }
+        return -1;
+    }
+
+    @Override
+    public void put(K key, V value) {
+        if (size == MAX_LENGTH) {
+            return;
+        }
```
liliia-ponomarenko, line 29: you should move this check to `if` block because now you can’t update the value if the array is full

liliia-ponomarenko (changes requested):
Good job! Let’s fix a few minor mistakes  ;)

liliia-ponomarenko (approved):
Great! 😊

PR #1245, opened 2023-08-23
Completed homework (I honestly thought I've made a push yesterday)

torbinofficial:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAXIMUM_SIZE = 10;
```
liliia-ponomarenko, line 6: Why public?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+    private Pair<K, V>[] storage;
+    private int size = 0;
```
liliia-ponomarenko, line 8:
```suggestion
    private int size;
```
0 - default  value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+
+    @SuppressWarnings("unchecked")
```
liliia-ponomarenko, line 10: Let’s remove this annotation ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+    public StorageImpl() {
+        storage = new Pair[MAXIMUM_SIZE];
+    }
+
+    private static class Pair<K, V> {
```
liliia-ponomarenko, line 15: Let’s  move private  class after  public and private  methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
     public void put(K key, V value) {
+        if (size < MAXIMUM_SIZE) {
+            int index = getIndex(key);
+            if (index == -1) {
+                storage[size] = new Pair<>(key, value);
+                size++;
+            } else {
+                storage[index].setValue(value);
+            }
+        }
```
liliia-ponomarenko, line 47: avoid high code nesting. Also, you should check the size only when you add a new element. Now you can't update the value if the array is full

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+    }
+
+    public int getIndex(K key) {
```
liliia-ponomarenko, line 50: this method should be private and move after public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+            }
+        }
+        return -1;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        return (index != -1) ? storage[index].getValue() : null;
```
liliia-ponomarenko, line 63:
```suggestion
        return index != -1 ? storage[index].getValue() : null;
```
Redundant  brackets

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,68 @@
...
+
+        public K getKey() {
+            return key;
+        }
+
+        public V getValue() {
+            return value;
+        }
+    }
+
```
liliia-ponomarenko, line 69: Redundant empty line

liliia-ponomarenko (approved):
LGTM 🔥

PR #1249, opened 2023-08-30
solution

IhorTrokhymchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 14: it's redundant, ints are 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null ? key == null : keys[i].equals(key)) {
```
okuzan, line 20: extract this common logic (you have it in put and get) to a private utility method

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K[] keysArray, K keyToFind) {
```
```suggestion
    private int findKeyIndex(K[] keys, K keyToFind) {
```
don't use data type in naming, java is a strongly typed language, keys can only be array

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            values[index] = value;
+            return;
+        }
+
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Size is over");
```
okuzan, line 29:
```suggestion
            throw new RuntimeException("Size is exceeded");
```

okuzan (approved):
Good job!

PR #1252, opened 2023-08-31
implementation of Storage  which parameterized with two types of data

nastasiiaKulinich:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size = 0;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_ITEMS_NUMBER];
+        this.values = new Object[MAX_ITEMS_NUMBER];
+    }
+
```
```suggestion
    private K[] keys;
    private V[] values;
    private int size;

    public StorageImpl() {
        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
        values = (V[]) new Object[MAX_ITEMS_NUMBER];
    }

```
we should use type parameters here, 0 is the default for int, `this` are redundant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
okuzan, line 18: same logic in two methods, extract to a private utility method
image

okuzan (changes requested):

kozhukhovsky (approved):

PR #1254, opened 2023-09-01
add solution

NazarOliinyk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+    private static final int MAX_CAPACITY = 10;
+
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
kozhukhovsky, line 15:
```suggestion
```

0 is default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+            }
+            return null;
+        }
+
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && key.equals(keys[i])) {
+                return values[i];
+            }
+        }
         return null;
```
kozhukhovsky, line 65:
```suggestion
    for (int i = 0; i < size; i++) {
            if (key == keyArray[i] || key != null && key.equals(keyArray[i])) {
                return valueArray[i];
            }
        }
        return null;
```

make it simple

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+            }
+        }
+
+        if (index == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[index] = value;
+        }
```
kozhukhovsky, line 46:
```suggestion
for (int i = 0; i < size; i++) {
            if (key == keyArray[i] || key != null && key.equals(keyArray[i])) {
                valueArray[i] = value;
                return;
            }
        }
        keyArray[size] = key;
        valueArray[size] = value;
        size++;
```

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        values = (V[]) new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_CAPACITY) {
+            throw new RuntimeException("Storage is full");
+        }
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
olekskov, line 23:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
olekskov, line 37:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

olekskov (approved):
Good job, approved.
Condition may be simplified further as already advised by Anton before

PR #1261, opened 2023-09-12
add StorageImpl parametrized class that implements Storage and got...

RhymeWriter:
… realization of its methods

### DIFF
--- a/src/main/java/Main.java
+++ b/src/main/java/Main.java
@@ -1,0 +1,21 @@
+import core.basesyntax.Storage;
+import core.basesyntax.impl.StorageImpl;
+
+public class Main {
+    public static void main(String[] args) {
+        String string = "okey";
+        String stringTwo = "bingo";
+        String stringThree = "cool";
+
+        Storage<Integer, String> storage = new StorageImpl<>();
+        storage.put(1, string);
+        storage.put(2, stringTwo);
+        storage.put(3, stringThree);
+
+        String value = storage.get(3);
+        System.out.println(value);
+        int size = storage.size();
+        System.out.println(size);
+
+    }
+}
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,55 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private final K[] keysStorage = (K[]) new Object[STORAGE_SIZE];
+    private final V[] valuesStorage = (V[]) new Object[STORAGE_SIZE];
+
     @Override
     public void put(K key, V value) {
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private final K[] keysStorage = (K[]) new Object[STORAGE_SIZE];
+    private final V[] valuesStorage = (V[]) new Object[STORAGE_SIZE];
+
     @Override
     public void put(K key, V value) {
```
Ivan95kos, line 12: Try to use one operator if
example:

```
    @Override
            if (compare(keys[i], key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
    }

    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
```
Ivan95kos, line 7:
```suggestion
    private static final int STORAGE_SIZE = 10;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (int i = 0; i < valuesStorage.length; i++) {
+            if (valuesStorage[i] != null) {
+                count++;
+            }
+        }
+        return count;
```
Ivan95kos, line 55:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+
+    private final K[] keysStorage = (K[]) new Object[STORAGE_SIZE];
+    private final V[] valuesStorage = (V[]) new Object[STORAGE_SIZE];
```
Ivan95kos, line 9: use the constructor to initialize these variables

src/main/java/Main.java
```diff
@@ -0,0 +1,21 @@
+import core.basesyntax.Storage;
+import core.basesyntax.impl.StorageImpl;
+
+public class Main {
```
Ivan95kos, line 4: Please, remove the `Main`
You have the `StorageImplTest` class

Ivan95kos (changes requested):

kozhukhovsky (approved):

PR #1262, opened 2023-09-14
Realized StorageImpl with 2 arrays and ability to save null keys

Trayvor:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    public static final int ARRAY_SIZE = 10;
+
+    private final K[] keyArr = (K[])new Object[ARRAY_SIZE];
+    private final V[] valArr = (V[])new Object[ARRAY_SIZE];
+    private int pos = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < pos; i++) {
+            if (key != null && keyArr[i] != null) {
```
kozhukhovsky, line 15: You can combine the check so that there is only one if block
key == keys[i] || key != null && key.equals(keys[i])

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int ARRAY_SIZE = 10;
+
+    private final K[] keyArr = (K[])new Object[ARRAY_SIZE];
```
kozhukhovsky, line 8:
```suggestion
    private final K[] keys = (K[])new Object[ARRAY_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private final V[] valArr = (V[])new Object[ARRAY_SIZE];
```
```suggestion
    private final V[] values = (V[])new Object[ARRAY_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private int pos = 0;
```
kozhukhovsky, line 10:
```suggestion
    private int size;
```

0 if default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        }
+        keyArr[pos] = key;
+        valArr[pos] = value;
+        pos++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < pos; i++) {
+            if (keyArr[i] != null && key != null) {
```
kozhukhovsky, line 35: same, combine check

kozhukhovsky (changes requested):

olekskov (changes requested):

olekskov (approved):
Good job

PR #1264, opened 2023-09-14
add Storage implementation

VoitseshukV:
Automatic storage increase is also implemented

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,66 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
+
+    public StorageImpl() {
+    }
+
     @Override
     public void put(K key, V value) {
+        if (keys != null) {
+            // Find & update value
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
+        if (keys == null || values == null || keys.length <= size) {
+            // Increase max storage size
+            increaseSize(key.getClass(), value.getClass());
+        }
+        // Add new pair
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (keys != null) {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
+                    return values[i];
+                }
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private void increaseSize(Class keyClass, Class valueClass) {
+        int newSize = keys == null || values == null ? 10 : keys.length + 10;
+        K[] newKeys = (K[]) Array.newInstance(keyClass, newSize);
+        V[] newValues = (V[]) Array.newInstance(valueClass, newSize);
+        if (keys != null) {
+            System.arraycopy(keys, 0, newKeys, 0, keys.length);
+        }
+        keys = newKeys;
+        if (values != null) {
+            System.arraycopy(values, 0, newValues, 0, values.length);
+        }
+        values = newValues;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size = 0;
```
kozhukhovsky, line 11:
```suggestion
    private int size;
```

0 is default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
+
+    public StorageImpl() {
+    }
```
kozhukhovsky, line 14: init keys and values in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private void increaseSize(Class keyClass, Class valueClass) {
```
kozhukhovsky, line 54: no need new size in this task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (keys != null) {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
```
kozhukhovsky, line 41: see checklist
Don't use Objects class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
+
+    public StorageImpl() {
+    }
+
     @Override
     public void put(K key, V value) {
+        if (keys != null) {
+            // Find & update value
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(keys[i], key)) {
```
kozhukhovsky, line 21: don't use Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,68 @@
...
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
+        if (keys == null || values == null || keys.length <= size) {
+            // Increase max storage size
+            increaseSize(key.getClass(), value.getClass());
+        }
```
kozhukhovsky, line 30:
```suggestion

```

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
VoitseshukV: @kozhukhovsky By the way, is this just a limitation in the learning phase to check understanding of how it works? Because IDE itself suggests replacing the condition with use Objects.
[screenshot](https://i.imgur.com/khACocq.png)

okuzan: yes, it's just so you can practice it's impl, in real life we would use objects.equals

okuzan (changes requested):
Extract common logic from put and get - in both methods you iterate from 0 to the end and check equality. As an option, you could implement a private method, which would do this and return the index if found, and for example -1 if not found

src/main/java/core/basesyntax/impl/StorageImpl.java
VoitseshukV: Thanks

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        return values[index];
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findByValue(K key) {
```
olekskov, line 44:
```suggestion
    private int findIndexByKey(K key) {
```

olekskov (approved):
Good job, approved.
Please check method naming proposal below. 
1) Method names should be as descriptive as possible
2) We have variable value so having this word in method name when it's not used there is confusing

PR #1265, opened 2023-09-15
found some problems

KhrystynaDidykh:
I can't figure out how should it work

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,12 +3,24 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_SIZE];
     @Override
     public void put(K key, V value) {
+      for (int i = 0; i < valueArray.length; i++) {
+          keyArray[i] = key;
+          valueArray[i] = value;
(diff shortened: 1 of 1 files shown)


okuzan (changes requested):
Please always make sure the build passes locally before creating a PR and requesting a mentor's review. Here, you should make a new line before the first method at least, Also, revisit checklist of some common mistakes

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_SIZE = 10;
+    private K[] keyArray;
+    private V[] valueArray;
+    private int size;
+
+    public StorageImpl() {
+        keyArray = (K[]) new Object[MAX_SIZE];
+        valueArray = (V[]) new Object[MAX_SIZE];
+        //size = 0;
```
okuzan, line 14: please remove comments before committing

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size <= MAX_SIZE) {
+            for (int i = 0; i < size; i++) {
+                if (keyArray[i] == key
+                        || keyArray[i] != null
+                        && keyArray[i].equals(key)) {
```
okuzan, line 23: this is common logic for two methods, extract it to a private utility method and call it in both places

okuzan (changes requested):
Good job so far!

olekskov (approved):
Good job

PR #1266, opened 2023-09-16
Implementation of the task "JV-storage"

OlhaHirniak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STORAGE_SIZE = 10;
+    private int size = 0;
+    private Object[] keys;
+    private Object[] values;
```
Ivan95kos, line 9: Use parameterized type
```suggestion
    private K [] keys;
    private V [] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        values = new Object[STORAGE_SIZE];
+    }
+
+    public Object[] getKeys() {
+        return keys;
+    }
+
+    public Object[] getValues() {
+        return values;
+    }
```
Ivan95kos, line 22: useless methods
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STORAGE_SIZE = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```
0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == null && key == null) {
```
okuzan, line 19:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```
extract the logic of iteratin and comparing - it's common for both methods, put and get. Or at least extract the comparison logic

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == null && keys[i] == key) {
```
okuzan, line 32:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == keys[i]) {
```

okuzan (changes requested):

olekskov (approved):
Good job

PR #1268, opened 2023-09-20
added the solution

flekros:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_CAPACITY) {
+            throw new RuntimeException("Storage is full.");
```
bhdnchui, line 20: I think you should specify that exception.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_CAPACITY) {
```
bhdnchui, line 19: You can simplify that with `==` operation.
You will never make a `size` more than MAX_CAPACITY if you make such a check at the start of the `put` method.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
bhdnchui, line 14: It is redundant. Default int value is 0.
See the Default Values article [here](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html).

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,74 @@
...
+
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
```
bhdnchui, line 50: Also think about [Objects#equals](https://www.baeldung.com/java-comparing-objects#equals-static) method here.

bhdnchui (changes requested):
Good job, but please, take a look at my comments.

dima-semeniuk (approved):
Great job!

bhdnchui (approved):
Looks good to me.

PR #1269, opened 2023-09-21
Solution for jv-storage

ivankusur:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
olekskov, line 9: 0 is default value of int, no need to explicitly set it
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            return;
+        }
+
+        if (size < MAX_ITEMS_NUMBER) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            keys[MAX_ITEMS_NUMBER - 1] = key;
+            values[MAX_ITEMS_NUMBER - 1] = value;
```
olekskov, line 30: replacing last element without any warning may cause confusion to user
better to throw some exception in such cases to give a clue that something went wrong (storage is full in this case)

olekskov (approved):
Good job overall, approved.
Comments left for further consideration

PR #1272, opened 2023-09-22
Implemented storage with its methods

dima-semeniuk:


hehetenya (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private Element<K, V>[] storage = null;
+    private int position = 0;
+
+    private static class Element<K, V> {
```
bhdnchui, line 10: Please, move that class to a separate file.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private Element<K, V>[] storage = null;
```
bhdnchui, line 7: Please, follow [that rule](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor).

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+    private int position = 0;
```
bhdnchui, line 8: It is redundant. Default int value is 0.
See the Default Values article [here](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html).

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        public Element(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size() == 0) {
+            storage = new Element[STORAGE_SIZE];
```
bhdnchui, line 23: It would be better to initialize storage in a constructor.
https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#all-non-constant-fields-should-be-initialized-in-the-constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+
+        public Element(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size() == 0) {
```
bhdnchui, line 22: Why you cannot use `position` field here?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        }
+        if (position < STORAGE_SIZE) {
+            Element<K, V> element = new Element<>(key, value);
+            storage[position++] = element;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (size() > 0) {
```
bhdnchui, line 39: You dont need that check if you will initialize `storage` in a constructor.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return position;
+    }
+
+    private boolean compareKeyAndStorageKey(K key, K storageKey) {
+        return ((key == null && storageKey == null)
+                || (storageKey != null && storageKey.equals(key)));
```
bhdnchui, line 56: Please, think about using [Objects#equals](https://www.baeldung.com/java-comparing-objects#equals-static) method here.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return position;
+    }
+
+    private boolean compareKeyAndStorageKey(K key, K storageKey) {
```
bhdnchui, line 54: Please, think about simplifying that name.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        return ((key == null && storageKey == null)
+                || (storageKey != null && storageKey.equals(key)));
     }
 }
+
+
```
bhdnchui, line 60: Remove those redundant lines.

bhdnchui (changes requested):
Good job, but please, look at my comments.

src/main/java/core/basesyntax/impl/StorageImpl.java
dima-semeniuk: @bhdnchui but why is it necessary to move the class  Element to a separate file, if I need it only inside this class? If I move it I will need to create getters to have access to K and V

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return position;
+    }
+
+    private boolean keysComparing(K key, K storageKey) {
+        return Objects.equals(key, storageKey);
```
bhdnchui, line 46: Please, don't use `Object` class. 
It is [common mistake](https://mate-academy.github.io/jv-program-common-mistakes/java-core/generics/storage.html#dont-use-class-objects).

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
+import core.basesyntax.Element;
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private Element<K, V>[] storage;
+    private int position;
```
bhdnchui, line 10: Can we rename it to `size`?
If we can void the term `position` then we will get a little simpler code. 
We do not use `position` separately from the term `size` anywhere.

bhdnchui (changes requested):
Very good job!
But just a few comments need to be fixed.

kshuryhin (approved):
Great job!

PR #1280, opened 2023-09-29
Implemented a key-value storage that is parameterized by two types of…

CapybaraFrog:
… data. We have the ability to add a new key-value pair and get the value by key.

If a key-value pair is specified and the storage already contains this key, we replace the value.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,50 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean getStatus(K key1, K key2) {
```
kozhukhovsky, line 49:
```suggestion
    private boolean isEqual(K key1, K key2) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            size++;
+        } else {
+            throw new IllegalStateException("Storage is full");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key != null && key.equals(keys[i])) || (keys[i] == key)) {
```
kozhukhovsky, line 37: use method isEqual

kozhukhovsky (approved):

PR #1283, opened 2023-10-02
Overrided methods in storage class

HovorukhaBohdan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+        return size;
+    }
+
+    private int findByKey(K key) {
+        for (int i = 0; i < this.key.length; i++) {
+            if (this.key[i] == key || this.key[i] != null && this.key[i].equals(key)) {
+                return i;
+            }
+        }
+
```
proxychang, line 55: You don`t need empty line here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
     @Override
     public int size() {
+        int size = 0;
+
+        for (V element : value) {
+            if (element != null) {
+                size++;
+            }
+        }
+
```
proxychang, line 45: No need for an empty line here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] key;
+    private V[] value;
+    private int index = 0;
+
+    public StorageImpl() {
+        this.key = (K[]) new Object[MAX_SIZE];
+        this.value = (V[]) new Object[MAX_SIZE];
```
proxychang, line 13: It seems that you can initialize K[] and V[] on the class level

proxychang (approved):
Good job! I noticed couple of points and it will help to make your code better

src/main/java/core/basesyntax/impl/StorageImpl.java
proxychang: > It seems that you can initialize K[] and V[] on the class level

My mistake, you did right by declaring it on a class level and initializating in constructor

proxychang (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
...
+        V[] valueCopy = (V[]) new Object[size << 1];
+        System.arraycopy(value, 0, valueCopy, 0, value.length);
+        value = valueCopy;
+    }
+
+    private int countNotNullValues() {
+        int size = 0;
+
+        for (V element : value) {
+            if (element != null) {
```
kshuryhin, line 72: Correct me if I am wrong, but we can put our _key_ and **value=null**. So it will increase our `size`, but `size()` method will not consider those entry. What do you think about that?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,77 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] key;
+    private V[] value;
```
kshuryhin, line 8: Minor, but it is better to use plural here, like `keys`, `values`

kshuryhin (changes requested):
Cool, but let's discuss some moments

kshuryhin (approved):
Nice!

PR #1286, opened 2023-10-05
-Implement the Storage of key-value type

andriimikhaylovskiy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        if (size <= MAX_SIZE) {
+            int index = findIndex(key);
+            if (findIndex(key) != -1) {
+                valueArray[index] = value;
+                size--;
+            }
+            keyArray[size] = key;
+            valueArray[size] = value;
+            size++;
+        }
```
olekskov, line 39: it's strange that after replacement of existing key you still adding additional key/value pair in the end
```suggestion
        if (size <= MAX_SIZE) {
            int index = findIndex(key);
            if (findIndex(key) != -1) {
                valueArray[index] = value;
                return;
            }
            keyArray[size] = key;
            valueArray[size] = value;
            size++;
        }
```

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    static final int MAX_SIZE = 10;
+    private K[] keyArray;
+    private V[] valueArray;
```
okuzan, line 8:
```suggestion
    private K[] keys;
    private V[] values;
```
we don't usually include types in the names of a variable - Java is a strongly typed language

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keyArray = (K[]) new Object[MAX_SIZE];
+        valueArray = (V[]) new Object[MAX_SIZE];
+    }
+
+    private <K> int findIndex(K key) {
+        int index;
```
okuzan, line 17: private methods should go after public ones

Ivan95kos (approved):
Nice

PR #1287, opened 2023-10-06
implemented the put get size method and added a constructor

DevIgork:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        values[usedSpace] = value;
+        usedSpace++;
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            key = (K) KEY_TYPE_NULL;
+        }
+        for (int i = 0; i < keys.length; i++) {
```
ponomvrenko, line 41: We can replace keys.length with usedSpace field.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final String KEY_TYPE_NULL = "null";
```
ponomvrenko, line 6: Sorry can't understand properly, why do we need it?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        keys = (K[])new Object[MAX_ITEMS_NUMBER];
+        values = (V[])new Object[MAX_ITEMS_NUMBER];
+        usedSpace = INITIAL_USED_SPACE;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            key = (K) KEY_TYPE_NULL;
+        }
```
ponomvrenko, line 24: Maybe it will be easier to add this logic in the for-loop condition.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+        for (int i = 0; i < usedSpace; i++) {
+            if (keys[i].equals(key)) {
```
ponomvrenko, line 26: I recommend you to take a look inside Objects.equals method. It can help to remove the redundant code below

src/main/java/core/basesyntax/impl/StorageImpl.java
ponomvrenko, line 6: Sorry can't understand properly, why do we need it?

src/main/java/core/basesyntax/impl/StorageImpl.java
ponomvrenko, line 24: Maybe it will be easier to add this logic in the for-loop condition.

src/main/java/core/basesyntax/impl/StorageImpl.java
ponomvrenko, line 26: I recommend you to take a look inside Objects.equals method. It can help to remove the redundant code below

src/main/java/core/basesyntax/impl/StorageImpl.java
ponomvrenko, line 41: We can replace keys.length with usedSpace field.

ponomvrenko (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
vadymhrnk, line 24: I think this is redundant, you need to change if statement for null check

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
     @Override
     public V get(K key) {
+        if (key == null) {
+            key = (K) KEY_TYPE_NULL;
+        }
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] != null && keys[i].equals(key))) {
+                return values[i];
+            }
+        }
```
vadymhrnk, line 45: This for loop and other for loop in method `public void put(K key, V value)` might be transferred to a separate method

vadymhrnk (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        return usedSpace;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < usedSpace; i++) {
+            if (key == keys[i] || (keys[i] != null && keys[i].equals(key))) {
+                return i;
+            }
+        }
         return -1;
```
kshuryhin, line 46: As far as we use `-1` in several places, what do you think about making it a Java constant?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+            return;
+        }
+        keys[usedSpace] = key;
+        values[usedSpace] = value;
+        usedSpace++;
```
kshuryhin, line 25: Probably we should provide some mechanism to deal with situations, when our internal arrays are full

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = findKeyIndex(key);
+        V value = index == -1 ? null : values[index];
```
kshuryhin, line 31:
```suggestion
        return index == -1 ? null : values[index];
```

kshuryhin (changes requested):
Good job, but there are some points to consider

kshuryhin (approved):
Good job!

PR #1299, opened 2023-10-10
Storage functional

TymurPanchenk0:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_SIZE = 10;
```
EdenParadise, line 6: You should consider changing the access modifier of this field to private

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
+    }
```
EdenParadise, line 15: This piece of code is a relic. You can directly initialize the fields within the class body.

EdenParadise (changes requested):
Really decent approach. Those if() blocks almost made me crazy!

src/main/java/core/basesyntax/impl/StorageImpl.java
TymurPanchenk0: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
TymurPanchenk0: ok

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null || keys[i] != null && keys[i].equals(key)) {
```
mnyshenko, line 14: A few points here:
1) If you have a long condition that contains multiple && and || it's better to divide such conditions by using parentheses, so it improves readability 
2) You refer to keys[i] multiple times here, so it's better to extract it to a separate variable
3) It seems that line `keys[i] == null && key == null || keys[i] != null && keys[i].equals(key)` used twice in your code, thus it's code duplication. You should extract this expression to a separate method and reuse

mnyshenko (changes requested):
A few point to consider

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            K element = (K) keys[i];
+            if ((element == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
d1sam, line 15: Can you use here your `checkEquality(K k, K key)` method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                values[i] = value;
+                return;
+            }
+        }
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full");
```
d1sam, line 25: Make this message as a constant.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public boolean checkEquality(K k, K key) {
```
d1sam, line 45:
```suggestion
    private boolean checkEquality(K k, K key) {
```
Also please think about better names for input parameters. `K k, K key` - it's not a good idea.

d1sam (changes requested):
Good job, but please look at my comments and edit mistakes.

d1sam (approved):
good job)

PR #1300, opened 2023-10-10
jv-storage solution

BolsunovaNataliia:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     public void put(K key, V value) {
+        if (findKeyIndex(key) != -1) {
+            values[findKeyIndex(key)] = value;
+        } else {
+            if (numberOfElements < MAXIMUM_CAPACITY) {
+                keys[numberOfElements] = key;
+                values[numberOfElements] = value;
+                numberOfElements++;
+            } else {
+                throw new IndexOutOfBoundsException();
```
BohdanTrue, line 27: Try to provide some detailed error message

BohdanTrue (changes requested):
Great job!
Minor updates are required.

BohdanTrue (approved):

BohdanTrue (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return numberOfElements;
     }
```
mnyshenko, line 54: It's always better to put public methods before private because when you open some class, you want to see its general logic, not some specific private methods that serve public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        }
+        return -1;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < numberOfElements; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
+                return (V) values[i];
+            }
```
mnyshenko, line 46: `keys[i]` is used twice here, let's extract it to a separate variable

mnyshenko, line 46: `findKeyIndex` and `get` seem to have almost identical logic, let's try to avoid code duplication and refactor this
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    public StorageImpl() {
+        this.keys = new Object[MAXIMUM_CAPACITY];
+        this.values = new Object[MAXIMUM_CAPACITY];
+        this.numberOfElements = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (findKeyIndex(key) != -1) {
+            values[findKeyIndex(key)] = value;
```
mnyshenko, line 20: when you enter this if you call findKeyIndex(key) `again` which is bad, let's store it outside of the if and reuse the result

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        } else {
+            if (numberOfElements < MAXIMUM_CAPACITY) {
+                keys[numberOfElements] = key;
+                values[numberOfElements] = value;
+                numberOfElements++;
+            } else {
+                throw new IndexOutOfBoundsException("Storage is full");
+            }
```
mnyshenko, line 28: Let's try to reduce the number of indentations here by removing redundant if else constructions here. Hint: we can use `return` if void methods

mnyshenko (changes requested):
A few points to consider

mnyshenko (approved):
Great

PR #1305, opened 2023-10-10
Homework branch. Made everything clear

EdenParadise:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    return (V) values[i];
+                }
+            } else if (key.equals(keys[i])) {
+                return (V) values[i];
+            }
```
d1sam, line 46: `(keys[i] == null && key == null) || (key != null && key.equals(keys[i]))` also put this condition inside one `if` statement and write your method without else.

d1sam, line 46:
```suggestion
            if ((keys[i] == null && key == null) || (key != null && key.equals(keys[i]))) {
                return (V) values[i];
            }
```
There is example of simplifying with if statement, which you should do inside put method.

d1sam, line 46: `(keys[i] == null && key == null) || (key != null && key.equals(keys[i]))` also put this condition inside one `if` statement and write your method without else.

d1sam (changes requested):
Good job, but some mistakes should be resolved.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (key != null && key.equals(keys[i]))) {
+                return (V) values[i];
+            }
+        }
         return null;
```
mnyshenko, line 33: `get` and `put` have almost identical parts, let's refactor this part so there is no code duplication
image

mnyshenko (changes requested):
Critical point to consider

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        }
+    }
+
+    private Object[] copyArray(Object[] array, int newCapacity) {
+        Object[] newArray = new Object[newCapacity];
+        System.arraycopy(array, 0, newArray, 0, size);
+        return newArray;
+    }
+
+    private boolean ifKeyIsValid(K key, int arrayIndex) {
```
d1sam, line 55:
```suggestion
    private boolean isKeyEquals(K key, int arrayIndex) {
```

d1sam (approved):
Good job, but there are some points which are need in your attention.

PR #1316, opened 2023-10-20
Solution to the task

druhov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,67 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY = 10;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 9: Use your type parameters here - K,V

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,67 @@
...
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[INITIAL_CAPACITY];
+        this.values = new Object[INITIAL_CAPACITY];
+        this.size = 0;
```
okuzan, line 15: this. are redundant, as is the last line - 0 is the default for int, it's useless

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,67 @@
...
+        return storedKey == null ? newKey == null : storedKey.equals(newKey);
+    }
+
+    private void ensureCapacity() {
+        if (size >= keys.length) {
+            int newCapacity = keys.length * 2;
+            keys = Arrays.copyOf(keys, newCapacity);
+            values = Arrays.copyOf(values, newCapacity);
+        }
+    }
```
okuzan, line 66: out of scope for this task, you don't need to care about it :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,67 @@
...
+            if (keyEquals(keys[i], key)) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean keyEquals(Object storedKey, K newKey) {
+        return storedKey == null ? newKey == null : storedKey.equals(newKey);
+    }
```
okuzan, line 58: good job on your private methods!

okuzan (changes requested):
Good job so far! 💪

olekskov (approved):
Good job

PR #1322, opened 2023-10-25
storage_my_solution

OlegShvets1:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+
+    private Object[] keys = new Object[MAX_CAPACITY];
+    private Object[] values = new Object[MAX_CAPACITY];
```
Ivan95kos, line 10: use the parameterized types and constructor to initialize them
```suggestion
    private K [] keys;
    private V [] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
...
+    private int size = 0;
```
Ivan95kos, line 11: int has default value `0`
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (Objects.equals(key, keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
```
Ivan95kos, line 33: use one loop like:
```
            if (keys[i] == null ? key == null : keys[i].equals(key)) {
                values[i] = value;
                return;
            }
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,61 @@
...
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
```
Ivan95kos, line 40: the same use one `loop`

Ivan95kos (changes requested):

kozhukhovsky (approved):

PR #1327, opened 2023-10-30
did my homework

Nazar090:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
```
sokimaaa, line 10: Use generic types. 
```suggestion
    private K[] keys;
    private V[] values;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+        size = 0;
```
sokimaaa, line 15: size=0 is redundant, java will instantiate the same value manually. 
So to be removed
```suggestion
```

sokimaaa (changes requested):

Sanderfuck (approved):

PR #1329, opened 2023-10-30
Storage Test Solution 31/10

VadimKadasiev:


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
okuzan, line 4: undo this change

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INDEX_KEY = 0;
+    private static final int INDEX_VALUE = 1;
+    private int arraySize = 0;
+    private int arrayIndex = 0;
+    private final Object[][] tempValuesArray = new Object[10][2];
+    private Object[][] resultValuesArray = new Object[0][0];
```
okuzan, line 11: image

okuzan (changes requested):
Interesting approach! Although you don't use K,V type parameters
Other things to consider:
- 0 is the default for int
- Extract common logic to a private method (iterating and comparing keys)
- Do you really need a temp array as a field?
- simplify equals logic -  return (key1 == key2) || (key1 != null && key2 != null && key1.equals(key2));

VadimKadasiev:
Hi! I was looking for a way to rid of the temporary array. Not sure about array size solution.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return arraySize;
     }
 }
+
+
```
kozhukhovsky, line 51:
```suggestion
```

kozhukhovsky (approved):

PR #1330, opened 2023-10-31
A generic class named StorageImpl is implemented, which implements th…

rafik2211:
Storage

A generic class named StorageImpl is implemented, which implements the Storage interface for key-value storage.
It defines a maximum size (MAX_SIZE) for the storage, which is set to 10. It uses two arrays, 'keys' and 'values', to store key-value pairs, both initially created with a size of MAX_SIZE. The 'size' variable keeps track of the number of elements currently in the storage. The put(K key, V value) method allows adding key-value pairs to the storage. It first checks if the key is null and, if so, searches for a null key in the storage. If the key is not null, it looks for an existing key in the storage. If the key is found, the method updates the associated value. If the key is not found, and the storage has room (size < MAX_SIZE), a new key-value pair is added to the arrays. The get(K key) method retrieves a value from the storage based on the provided key. It checks if the key is null and, if so, searches for a null key. If the key is not null, it searches for the key in the storage and returns the associated value. If the key is not found, it returns null. The size() method returns the current size of the storage, indicating how many key-value pairs are stored.

### DIFF
--- a/src/main/java/core/basesyntax/Storage.java
+++ b/src/main/java/core/basesyntax/Storage.java
@@ -1,7 +1,6 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
-
     void put(K key, V value);
 
     V get(K key);
--- a/src/main/java/core/basesyntax/Storage.java
+++ b/src/main/java/core/basesyntax/Storage.java
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
     void put(K key, V value);
 
     V get(K key);
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -4,33 +4,38 @@ import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     private static final int MAX_SIZE = 10;
-    private Object[] keys = new Object[MAX_SIZE];
-    private Object[] values = new Object[MAX_SIZE];
+    private K[] keys;
+    private V[] values;
     private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
 
     @Override
     public void put(K key, V value) {
+        if (size >= MAX_SIZE) {
+            throw new RuntimeException("Maximum size exceeded");
+
+        }
         for (int i = 0; i < size; i++) {
-            if (key == keys[i] || key != null && key.equals(keys[i])) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
                 values[i] = value;
                 return;
             }
         }
 
-        if (size < MAX_SIZE) {
-            keys[size] = key;
-            values[size] = value;
-            size++;
-        } else {
-            throw new RuntimeException("Maximum size exceeded");
-        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
         for (int i = 0; i < size; i++) {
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
kozhukhovsky, line 4:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
+    private int size = 0;
```
```suggestion
    private int size;
```

0 is default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
```
kozhukhovsky, line 15: You can combine these two checks into one

kozhukhovsky: key == keys[i] || key != null && key.equals(keys[i])

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
+
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
```
kozhukhovsky, line 32: This is better to do anyway. But it turns out that the method worked without errors, but the result did not change.
It is necessary that if it is already greater than the maximum value, throw some kind of error.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
```
kozhukhovsky, line 40: same, combine

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Storage.java
Ivan95kos: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
```
Ivan95kos, line 8: Use the parameterized types
All non-constant fields should be initialized in the constructor
```suggestion
    private K [] keys;
    private V [] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
Ivan95kos: not fixed

Ivan95kos (changes requested):
please, fix the previous comments

kozhukhovsky (approved):

PR #1337, opened 2023-11-09
successfully passed all tests

bohdan2704:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,42 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.ArrayList;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final ArrayList<K> kArray;
```
sokimaaa, line 7: [Don't use any kind of List, Set or Map](https://github.com/mate-academy/jv-storage/blob/master/checklist.md#dont-use-any-kind-of-list-set-or-map)

sokimaaa, line 7: You should solve this task using one or two arrays (this depends on your implementation). If you use arrays for your solution, we assume that the maximum number of elements in our Storage is 10.

sokimaaa (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,76 @@
...
+            index++;
+        }
+        return -1; // Element not found in the array
+    }
+
+    public <T> T[] addElem(T[] arr, T element) {
+        T[] newArr = Arrays.copyOf(arr, arr.length + 1);
+        newArr[arr.length] = element;
+        return newArr;
     }
```
kerrrusha, line 75: you should add element just by index of `size` value, and then increment `size` value. also add check for Storage overflow (when trying to add element when Storage is already full - `size == MAX_STORAGE_SIZE`)

kerrrusha (changes requested):

sokimaaa (approved):

PR #1341, opened 2023-11-14
implemented StorageImpl

letmerelaxpls:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (size < MAX_SIZE) {
```
kerrrusha, line 24: would be nice to inform to user the cause of rejecting to put element to the storage (for example, throwing exception that the storage is full)

kerrrusha (approved):

PR #1344, opened 2023-11-17
implemented the Storage of the key-value type

YaroslavRadevychVynnytskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,77 @@
...
+    private static final int INITIAL_SIZE = 0;
+    private int currentSize;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[INITIAL_SIZE];
+        this.values = (V[]) new Object[INITIAL_SIZE];
+        this.currentSize = INITIAL_SIZE;
+    }
```
sokimaaa, line 16: image

Make initial size = 10, to optimize our code (do not copy array for increasing size)

by default java will set `int currentSize to 0`, so we don't need `15 code line`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,77 @@
...
-        return -1;
+        return currentSize;
+    }
+
+    public void expandCapacity() {
+        if (currentSize == keys.length) {
+            int newCapacity = keys.length + 1;
+            keys = Arrays.copyOf(keys, newCapacity);
+            values = Arrays.copyOf(values, newCapacity);
+        }
```
sokimaaa, line 75: 1 method should be private
2 we don't need such logic by the task description
```suggestion
    public void expandCapacity() {
        if (currentSize == keys.length) {
            int newCapacity = keys.length + 1;
            keys = Arrays.copyOf(keys, newCapacity);
            values = Arrays.copyOf(values, newCapacity);
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,77 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < currentSize; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    repeatKey = true;
+                    break;
+                }
+            }
```
sokimaaa, line 36: you have pretty similar for-loops. 
think about using only one.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,77 @@
...
+                    return values[i];
+                }
+            }
+        } else {
+            for (int i = 0; i < currentSize; i++) {
+                if (key.equals(keys[i])) {
+                    return values[i];
+                }
+            }
+        }
```
sokimaaa, line 61: you have two similar loops. 
you dont have to check null key separately.

sokimaaa (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    public StorageImpl() {
+        this.keys = (K[]) new Object[INITIAL_SIZE];
+        this.values = (V[]) new Object[INITIAL_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean repeatKey = false;
+        for (int i = 0; i < currentSize; i++) {
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
Rommelua, line 20:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

Rommelua (approved):

PR #1346, opened 2023-11-18
Added solution to jv-storage

MilianCode:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
sokimaaa, line 8: Use generic types here
```suggestion
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private int numberOfElements = 0;
```
sokimaaa, line 9:
```suggestion
    private int numberOfElements;
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #1348, opened 2023-11-23
Completed task, implement override method put(), get(), size, and cre…

AriohDK:
…ate new method isEquals().

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,45 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (isEquals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
```
sokimaaa, line 23: repeating code should be extracted to the private method

```suggestion
        for (int i = 0; i < size; i++) {
            if (isEquals(keys[i], key)) {
                values[i] = value;
                return;
            }
        }
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #1350, opened 2023-11-23
solution

DojiDoe:


Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/core/basesyntax/StorageObject.java
```diff
@@ -0,0 +1,53 @@
...
+    public int hashCode() {
+        int result = 17;
+        result = 31 * result + (key == null ? 0 : key.hashCode());
+        result = 31 * result + (value == null ? 0 : value.hashCode());
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "{" + "key="
```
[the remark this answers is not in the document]
sokimaaa: make it nested class

it uses only inside `Storage, so we don't need it public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,59 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private static final int PRIMARY_SIZE = 0;
+    private final StorageObject[] storage;
+    private int size;
+
+    public StorageImpl() {
+        this.size = PRIMARY_SIZE;
```
sokimaaa, line 14: redundant, java set 0 as default for primitive
```suggestion
    private final StorageObject[] storage;
    private int size;

    public StorageImpl() {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,59 @@
...
     public V get(K key) {
+        for (StorageObject object : storage) {
+            if (object == null) {
+                break;
+            } else if (object.getKey() == null && key == null) {
+                return (V) object.getValue();
+            } else if (object.getKey() != null
+                    && object.getKey().equals(key)) {
+                return (V) object.getValue();
+            }
```
sokimaaa, line 50: if statement could be simplified. 

you have similar logic in each clause

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,59 @@
...
+                    break;
+                } else if (storage[i].getKey() == null
+                        && storageToAdd.getKey() == null) {
+                    storage[i] = storageToAdd;
+                    break;
+                } else if (storage[i].getKey() != null
+                        && storage[i].getKey().equals(storageToAdd.getKey())) {
+                    storage[i] = storageToAdd;
+                    break;
+                }
```
sokimaaa, line 35: simplify if-else

sokimaaa (changes requested):

sokimaaa (approved):

PR #1352, opened 2023-11-29
implemented realize f method put() and get()

petro2061:
In this task implemented the Storage of the key-value type that is parameterized with two types of data. 
For add datas to Storage uses method put() and method get() use for get value by key.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,51 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] values;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    }
+
+    public int keyIndexFounder(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
+        }
+        return -1;
+    }
```
olekskov, line 24: 1) Do not expose internal logic
And also put private methods after public ones

2) Methods names should be verbs for better code readability
```suggestion
    private int findKeyIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                return i;
            }
        }
        return -1;
    }
```

olekskov (changes requested):

PR #1358, opened 2023-12-10
task completed

YaroslavBerezhniak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }
+
```
okuzan, line 66: please make sure each file ends with exactly one empty line

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+            size++;
+        }
+
     }
 
     @Override
     public V get(K key) {
+        int i;
+        for (i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null) || (key != null
```
okuzan, line 53: redundant parentheses and first part can be simplified
```suggestion
            if (key == keys[i) || key != null
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+            }
+
+            if (size == MAX_SIZE) {
+                throw new IllegalStateException("Storage is full");
+            }
+
+            keys[size] = null;
+            values[size] = value;
+            size++;
+        } else {
```
okuzan, line 30: remove external if-else, instead amend the logical condition - combine them (line 32, line 17). We should always avoid unjustified code duplication. Simplify this method. Iterating and comparing keys - you have this logic in get too. Extract i to a private method and reuse in both public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
```
okuzan, line 9: That's where we should use type parameters.  Also, initialize in constructor
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+    private int size = 0;
```
okuzan, line 10: 0 is default for int

okuzan (changes requested):
Before you submit your work for review, please go through the [checklist ](https://github.com/mate-academy/jv-storage/blob/master/checklist.md) under the task and make any necessary corrections. 📝✅ Your mentor will be happy to help once those points are sorted out.

okuzan (changes requested):
_comment for sync with platform__

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    private Object[] keys;
+    private Object[] values;
```
Sanderfuck, line 9: You have generics for this types

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private boolean areKeysEqual(K key1, Object key2) {
+        return (key1 == null && key2 == null)
+                || (key1 != null && key1.equals(key2));
     }
```
Sanderfuck, line 44: Private methods place after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        int i = findIndex(key);
+        return (i != -1) ? (V) values[i] : null;
```
Sanderfuck, line 49: You don`t need cast here

Sanderfuck (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            if (areKeysEqual(key, keys[i])) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean areKeysEqual(K key1, Object key2) {
+        return (key1 == null && key2 == null)
+                || (key1 != null && key1.equals(key2));
```
Sanderfuck, line 53:
```suggestion
        return key1 == key2 || key1 != null && key1.equals(key2);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        int i = findIndex(key);
+        if (i != -1) {
+            values[i] = value;
+        } else if (size == MAX_SIZE) {
+            throw new IllegalStateException("Storage is full");
+        } else {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
Sanderfuck, line 27:
```suggestion
        if (i != -1) {
            values[i] = value;
            return;
        if (size == MAX_SIZE) {
            throw new IllegalStateException("Storage is full");
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

Sanderfuck: try this way

Sanderfuck (changes requested):

olekskov (approved):
Good job

PR #1359, opened 2023-12-11
implemented Storage key-value

MaksymMazlov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
```
sokimaaa, line 9: Use generic type here, e.g. K[], V[]
```suggestion
    private final Object[] keys = new Object[MAX_SIZE];
    private final Object[] values = new Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private int size = 0;
```
sokimaaa, line 10:
```suggestion
    private int size;
```

sokimaaa (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+
+    private final K[] keys = (K[]) new Object[MAX_SIZE];
+    private final V[] values = (V[]) new Object[MAX_SIZE];
```
kerrrusha, line 9: init this in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size == MAX_SIZE) {
+                throw new RuntimeException("Storage is full");
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
kerrrusha, line 25: dont complicate if-else logic
```suggestion
            return;
        } 
        if (size == MAX_SIZE) {
            throw new RuntimeException("Storage is full");
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

kerrrusha (changes requested):
can be improved a little bit

sokimaaa (approved):

PR #1362, opened 2023-12-14
jv-storage

rus1980707:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX];
+        values = new Object[MAX];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
olekskov, line 20: Task requires you to avoid usage of Objects class

Please write your own null-safe comparison code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
+            keys[size] = key;
+            values[size] = (V) value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
olekskov, line 35: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,46 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX];
+        values = new Object[MAX];
+    }
```
olekskov, line 15: We need to use class parametrization to avoid type mismatch
```suggestion
    private static final int MAX = 10;
    private final K[] keys;
    private final V[] values;
    private int size;

    public StorageImpl() {
        keys = (K[]) new Object[MAX];
        values = (V[]) new Object[MAX];
    }
```

olekskov (changes requested):

olekskov (changes requested):
`

kozhukhovsky (approved):

PR #1364, opened 2023-12-15
hm for review

vdbond:


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
kozhukhovsky, line 4:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,68 @@
 
 import core.basesyntax.Storage;
 
+// import javax.swing.*;
+
```
kozhukhovsky, line 6:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,68 @@
...
+        V result = null;
+        for (int i = 0; i < keys.length; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    result = values[i];
+                }
+            } else {
+                if (key.equals(keys[i])) {
+                    result = values[i];
+                }
```
kozhukhovsky, line 59: u can combine this

key == keys[i] || key != null && key.equals(keys[i])

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,68 @@
...
+    private K[] keys;
+    private V[] values;
+    private int storageSize;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+        values = (V[]) new Object[MAX_ARRAY_LENGTH];
+    }
+
+    public void getRewrittenKeysArray() {
```
kozhukhovsky, line 19: why do we need this method?

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+        keys = (K[]) new Object[MAX_ARRAY_LENGTH];
+        values = (V[]) new Object[MAX_ARRAY_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        keys[storageSize] = key;
+        values[storageSize] = value;
+        storageSize++;
     }
```
olekskov, line 21: You should add logic that will check if key is already present in storage and override value if so.
Storage cannot contain key duplicates. Otherwise get() method will process only first key

olekskov (changes requested):

olekskov (approved):
Good job

PR #1366, opened 2023-12-18
Implemented Storage.

davydov-software:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,90 @@
...
+import java.lang.reflect.Array;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LEN = 10;
+    @SuppressWarnings("unchecked")
+    private final Entry[] entries = (Entry[]) Array.newInstance(Entry.class, ARRAY_LEN);
+    private int size = 0;
+
+    public StorageImpl() {
+    }
```
olekskov, line 13: No-args constructor is available by default if no other constructors are present
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,90 @@
...
+
+    private void checkArraySize() {
+        if (size == ARRAY_LEN) {
+            throw new RuntimeException("Storage is full. Maximum capacity of " + ARRAY_LEN
+                    + " reached.");
+        }
+    }
+
+    private boolean areKeysEqual(K one, K other) {
+        return one == null && other == null
```
olekskov, line 65: May be a bit simplified
```suggestion
        return one == other
```

olekskov (approved):
Good job, approved

PR #1367, opened 2023-12-22
Implemented the Storage of the key-value type that is parameterized w…

Khrestina:
…ith two types of data.


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+        return null;
+    }
+
+    private V getNullKey() {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                return (V) values[i];
+            }
+        }
         return null;
```
kerrrusha, line 67: you can refactor your code and avoid duplicating by creating `getIndex` method

kerrrusha (changes requested):

sokimaaa (approved):

PR #1368, opened 2023-12-23
added fields, constructor, Override methods put, get, size(added priv…

RCherniavskiy:
…ate method findIndex)

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,50 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_SIZE];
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            values[index] = value;
+        } else {
+            if (size < MAX_SIZE) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new IllegalStateException("Storage is full");
+            }
+        }
```
olekskov, line 30: Generally it's considered that less tabulation levels we have the more readable becomes code.
Below you may see same logic but with slightly redesigned `if` blocks
```suggestion
        int index = findIndex(key);
        if (index != -1) {
            values[index] = value;
            return;
        }
        if (size > MAX_SIZE) {
            throw new IllegalStateException("Storage is full");
        } 
        keys[size] = key;
        values[size] = value;
        size++;            
```

olekskov (approved):
Good job, please pay attention to comment for further

src/main/java/core/basesyntax/impl/StorageImpl.java
RCherniavskiy: Your logic is clearer than mine. I completely agree with this. I will take your comments into account in the future. Thank you, your comment is invaluable to me.

PR #1371, opened 2024-01-01
Solution

suprun-anna:
created arrays for storing keys and values, methods for putting and getting values, getting size of storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,62 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[]{};
+        values = (V[]) new Object[]{};
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOf(key);
+        if (index == -1) {
+            if (keys.length + 1 > MAX_ITEMS_NUMBER) {
+                throw new RuntimeException("Cannot add more elements!"
+                        + " The size of storage is " + MAX_ITEMS_NUMBER);
+            }
+            keys = Arrays.copyOf(keys, keys.length + 1);
+            values = Arrays.copyOf(values, values.length + 1);
+            int size = keys.length;
+            keys[size - 1] = key;
+            values[size - 1] = value;
+        } else {
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (compareKeys(key, keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return keys.length;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (compareKeys(key, keys[i])) {
+                return i;
+            }
+        }
         return -1;
+    }
+
+    private boolean compareKeys(K key1, K key2) {
+        return key1 == key2 || key1 != null && key1.equals(key2);
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[]{};
+        values = (V[]) new Object[]{};
```
Rommelua, line 14:
```suggestion
        keys = (K[]) new Object[MAX_ITEMS_NUMBER ]{};
        values = (V[]) new Object[MAX_ITEMS_NUMBER ]{};
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+
+    private K[] keys;
+    private V[] values;
```
Rommelua, line 10:
```suggestion
    private K[] keys;
    private V[] values;
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
...
+            if (compareKeys(key, keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
+        return keys.length;
```
Rommelua, line 47:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
...
+        if (index == -1) {
+            if (keys.length + 1 > MAX_ITEMS_NUMBER) {
+                throw new RuntimeException("Cannot add more elements!"
+                        + " The size of storage is " + MAX_ITEMS_NUMBER);
+            }
+            keys = Arrays.copyOf(keys, keys.length + 1);
+            values = Arrays.copyOf(values, values.length + 1);
+            int size = keys.length;
+            keys[size - 1] = key;
+            values[size - 1] = value;
```
Rommelua, line 29: do not use Arrays
use size field

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+            if (size + 1 > MAX_ITEMS_NUMBER) {
+                throw new RuntimeException("Cannot add more elements!"
+                        + " The size of storage is " + MAX_ITEMS_NUMBER);
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[index] = value;
+        }
```
olekskov, line 31: 1) In general when code has less tabulations it looks simpler and cleaner.
Below same code but a bit reordered
2) Please avoid +1 or -1 in conditions when case may be handled with <= or >=
```suggestion
        int index = indexOf(key);
        if (index != -1) {
            values[index] = value;
            return;
        }
        if (size >= MAX_ITEMS_NUMBER) {
            throw new RuntimeException("Cannot add more elements!"
                    + " The size of storage is " + MAX_ITEMS_NUMBER);
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+    private static final int MAX_ITEMS_NUMBER = 10;
+
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+        size = 0;
```
olekskov, line 15: 0 is default int value, no need to explicitly set it
```suggestion
```

olekskov (changes requested):

olekskov (approved):
good job

PR #1373, opened 2024-01-02
implementedStorage

shumYevh:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+            }
+        }
+    }
+
+    private void putFirst(K key, V value) {
+        keys = (K[]) new Object[this.size];
+        keys[0] = key;
+        values = (V[]) new Object[this.size];
+        values[0] = value;
+    }
```
olekskov, line 30: Please initialize internal arrays with some capacity in no-args constructor instead of waiting for first input.
It will make code much simpler

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+        return this.size;
+    }
+
+    public void setKeys(K[] keys) {
+        this.keys = keys;
+    }
+
+    public void setValues(V[] values) {
+        this.values = values;
     }
```
olekskov, line 94: and setters will not be required

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return this.size;
```
olekskov, line 85: 'this' keyword is required to differentiate method argument and class variable with same name. Here we don't have any other size except class variable
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+    }
+
+    private boolean isKey(K wantedKey) {
+        for (K key : this.keys) {
+            if (key != null && wantedKey != null && key.equals(wantedKey)) {
+                return true;
+            }
+            if (key == wantedKey) {
+                return true;
+            }
```
olekskov, line 65: may be simplified
```suggestion
            if (key = wantedKey || key != null && key.equals(wantedKey)) {
                return true;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+            if (key != null && keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            }
+            if (key == keys[i]) {
+                values[i] = value;
+                break;
+            }
+        }
+    }
```
olekskov, line 43: You go through array to check if key exist and then do it again to replace value.
Let method that checking key existence return index in order to iterate only once

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,93 @@
...
+        newKeys[newKeys.length - 1] = key;
+        setKeys(newKeys);
+
+        V[] newValues = (V[]) new Object[this.size];
+        System.arraycopy(this.values, 0, newValues, 0, this.values.length);
+        newValues[newValues.length - 1] = value;
+        setValues(newValues);
+    }
+
+    private boolean isKey(K wantedKey) {
```
olekskov, line 58: All private method should be placed after all public methods

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
okuzan, line 6:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+    private K[] keys = (K[]) new Object[this.size];
+    private V[] values = (V[]) new Object[this.size];
+    private int size = 0;
```
```suggestion
    private int size;
```
0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] != null && key != null && keys[i].equals(key)) {
+                return values[i];
+            }
+            if (keys[i] == key) {
+                return values[i];
+            }
```
okuzan, line 29: Use your private method here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        newKeys[newKeys.length - 1] = key;
+        this.keys = newKeys;
+
+        V[] newValues = (V[]) new Object[this.size];
+        System.arraycopy(this.values, 0, newValues, 0, this.values.length);
+        newValues[newValues.length - 1] = value;
+        this.values = newValues;
+    }
+
+    private int checkKey(K wantedKey) {
```
```suggestion
    private int getKeyIndex(K wantedKey) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private K[] keys = (K[]) new Object[this.size];
+    private V[] values = (V[]) new Object[this.size];
```
okuzan, line 8: init in constructor

okuzan, line 8: don't use ` this` unless necessary

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        this.size++;
+        K[] newKeys = (K[]) new Object[this.size];
+        System.arraycopy(this.keys, 0, newKeys, 0, this.keys.length);
+        newKeys[newKeys.length - 1] = key;
+        this.keys = newKeys;
+
+        V[] newValues = (V[]) new Object[this.size];
+        System.arraycopy(this.values, 0, newValues, 0, this.values.length);
+        newValues[newValues.length - 1] = value;
+        this.values = newValues;
```
okuzan, line 44: Try to do this task without methods like System.arraycopy, it can easily be done with indexes only

okuzan (changes requested):

okuzan (changes requested):
remove redundant private methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
     public V get(K key) {
-        return null;
+        int searchedIndex = checkKey(key);
+        return (searchedIndex == -1) ? null : values[searchedIndex];
+    }
+
+    private void addKeyValue(K key, V value) {
+        this.size++;
+        addKey(key);
+        addValue(value);
```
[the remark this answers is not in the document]
okuzan: not fixed, you don't need to create another array, copy elements to it. just access existing array by correct indexes

okuzan (changes requested):
addKeyValue is overengineered, not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_SIZE = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_SIZE];
+        values = (V[]) new Object[DEFAULT_SIZE];
+        size = 0;
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
okuzan: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
     @Override
     public V get(K key) {
-        return null;
+        int searchedIndex = checkKey(key);
+        return (searchedIndex == -1) ? null : values[searchedIndex];
+    }
+
+    private int checkKey(K wantedKey) {
+        for (int i = 0; i < this.size; i++) {
+            if (Objects.equals(keys[i], wantedKey)) {
```
okuzan, line 38: сорі, я в себе використовував Objects.equals, бо так швидше, але в цій тасці студентам не [можна ](https://github.com/mate-academy/jv-storage/blob/master/checklist.md)😅

okuzan (changes requested):

olekskov (approved):

PR #1374, opened 2024-01-04
Implemented key-value type storage using 2 arrays.

Promaksua:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size;
+    private Object[] keys;
+    private Object[] values;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
```
olekskov, line 14: 0 is default int value, no need to explicitly aset it
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    }
+
+    private int getIndex(K key) {
```
olekskov, line 17: private methods should be placed after all public ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null) {
+                return i;
+            } else if (keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
```
olekskov, line 23: when you have same body in if and else then it for sure may be combined
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return i;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+
+        if (index != -1) {
+            values[index] = value;
+        } else if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
```
olekskov, line 40: Better to make majority of code without additional tabulations and minimize else blocks
With such approach code are simpler for reading and understanding
```suggestion
        if (index != -1) {
            values[index] = value;
            return;
        } 
        if (size >= MAX_SIZE) {
            throw new RuntimeException("Storage is full");
        } 
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private int size;
+    private Object[] keys;
+    private Object[] values;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
```
olekskov, line 13: It's better to make proper types in class variables to avoid type casting in methods code
```suggestion
    private K[] keys;
    private V[] values;

    public StorageImpl() {
        this.keys = (K[]) new Object[MAX_SIZE];
        this.values = (V[]) new Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = getIndex(key);
+        if (index != -1) {
+            return (V) values[index];
```
olekskov, line 47:
```suggestion
            return values[index];
```

olekskov (changes requested):
Good job overall, just some style comments

Sanderfuck (approved):

PR #1376, opened 2024-01-06
implemented StorageImpl

orest40th:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_CAPACITY];
+        this.values = (V[]) new Object[MAX_CAPACITY];
+        this.size = 0;
```
kerrrusha, line 14: int variables are zero by default, remove that
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+            values[index] = value;
+        } else {
+            if (size < MAX_CAPACITY) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new RuntimeException("Storage is full");
+            }
+        }
```
kerrrusha, line 30: avoid identations as much as possible
```suggestion
        if (index != -1) {
            values[index] = value;
            return;
        }
        if (size >= MAX_CAPACITY) {
            throw new RuntimeException("Storage is full");
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

kerrrusha (changes requested):

kerrrusha (approved):

PR #1377, opened 2024-01-07
Implement Storage

oKukHtina:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY = 10;
+    private int size;
+    private final Object[] arrayOfKeys;
+    private final Object[] arrayOfValues;
```
okuzan, line 9: use type parameters, K V
rename to just keys, values, we usually don't include type in variables' names

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+
+    public StorageImpl() {
+        arrayOfKeys = new Object[INITIAL_CAPACITY];
+        arrayOfValues = new Object[INITIAL_CAPACITY];
+        size = 0;
```
okuzan, line 14: this doesn't achieve anything, 0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        if (key == null) {
+            key = (K) "null";
+        }
+
+        for (int i = 0; i < size; i++) {
+            if (arrayOfKeys[i].equals(key)) {
+                arrayOfValues[i] = value;
+                return;
+            }
+        }
```
okuzan, line 28: extract common logic from put and get to a private method
remove first `if` and change condition in the second - look up code for Objects.equals, do the same, since you can't use it in this task

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+        keys = (K[]) new Object[INITIAL_CAPACITY];
+        values = (V[]) new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            key = (K) "null";
+        }
+
```
okuzan, line 22:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i],key)) {
```
okuzan, line 24: you can't use Objects.equals in this task, write your own equivalent code
format this file - Ctrl Alt L for windows, we need space after comma

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+            }
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        K searchKey = key != null ? key : (K) "null";
```
```suggestion
```

okuzan (changes requested):
Extract common code from put and get
If you have any problems, don't hesitate to ask in the chat!

kozhukhovsky (approved):

PR #1379, opened 2024-01-10
method was implemented

Artem-Kushniruk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_ARRAY_LENGTH = 10;
+    private final K[] keyStorage = (K[]) new Object[STORAGE_ARRAY_LENGTH];
+    private final V[] valueStorage = (V[]) new Object[STORAGE_ARRAY_LENGTH];
```
Sanderfuck, line 9: better init this in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
...
     public int size() {
-        return -1;
+        int count = 0;
+        for (K k: keyStorage) {
+            if (k == null && valueStorage[count] == null) {
+                break;
+            }
+            count++;
+        }
+        return count;
```
Sanderfuck, line 33:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
 
 import core.basesyntax.Storage;
 
+@SuppressWarnings("unchecked")
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_ARRAY_LENGTH = 10;
+    private final K[] keyStorage = (K[]) new Object[STORAGE_ARRAY_LENGTH];
+    private final V[] valueStorage = (V[]) new Object[STORAGE_ARRAY_LENGTH];
+
```
Sanderfuck, line 10:
```suggestion
private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
...
+                break;
+            }
+            count++;
+        }
+        return count;
+    }
+
+    public int getIndex(K key) {
+        int index = 0;
+        for (K k : keyStorage) {
```
Sanderfuck, line 38: k - it`s bad naming, write more informative

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,47 @@
...
+            if (k == null && valueStorage[index] == null
+                    || (key == k || (key != null && key.equals(k)))) {
```
Sanderfuck, line 40:
```suggestion
            if (k == key ||  k != null && k.equals(key)))) {
```

Sanderfuck (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Artem-Kushniruk: Ok, but why?

src/main/java/core/basesyntax/impl/StorageImpl.java
Artem-Kushniruk: I do not understand what do you mean? Because if delete lines 39-40 code won't work.

src/main/java/core/basesyntax/impl/StorageImpl.java
Sanderfuck: Because it`s another value as array length and you will have less code using this field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,54 @@
...
     public int size() {
-        return -1;
+        int count = 0;
+        for (K keyInStorage: keyStorage) {
+            if (keyInStorage == null && valueStorage[count] == null) {
+                break;
+            }
+            count++;
+        }
+        this.size = count;
```
Sanderfuck, line 39:
```suggestion
```

Sanderfuck: You don`t need this code, here just return the value of size! But you need to calculate size in a method where you add a new element

src/main/java/core/basesyntax/impl/StorageImpl.java
Sanderfuck: image
Red background need to delete and green need to add

Sanderfuck (changes requested):
Fix it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,52 @@
...
+                size++;
+                break;
+            }
+            if (keyInStorage == key || keyInStorage != null && keyInStorage.equals(key)) {
+                keyStorage[index] = key;
+                valueStorage[index] = value;
+                break;
+            }
+            index++;
+        }
```
kerrrusha, line 33: lets make code more simple
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keyArray[i] || key != null && key.equals(keyArray[i])) {
                valueArray[i] = value;
                return;
            }
        }
        [write code to check if size >= MAX_STORAGE_SIZE then throw exception]
        [write code to append new key and value]
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -2,18 +2,52 @@
...
 
     @Override
     public V get(K key) {
+        int index = 0;
+        for (K keyInStorage: keyStorage) {
+            if (keyInStorage == key || keyInStorage != null && keyInStorage.equals(key)) {
+                return valueStorage[index];
+            }
+            index++;
+        }
```
kerrrusha, line 44: let's make code more simple
```suggestion
        for (int i = 0; i < size; i++) {
            if (keyArray[i] == key || keyArray[i] != null && keyArray[i].equals(key)) {
                return valueStorage[i];
            }
        }
```

kerrrusha (changes requested):

sokimaaa (approved):

PR #1380, opened 2024-01-11
hw-jv-storage

VovaMaryniak:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,24 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final Map<K, V> storageMap = new HashMap<>();
```
okuzan, line 8: image
yes, this task can easily be solved via map, but using a ready solution wouldn't teach us anything here :(

okuzan (changes requested):
Make sure to read the checklist before submitting your solution 🙏

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[DEFAULT_CAPACITY];
+        this.values = new Object[DEFAULT_CAPACITY];
+        this.size = 0;
```
okuzan, line 14: we need to use `this` to distinguish between fields and arguments  which share the same name, that's not the case here, we can safely omit them
0 is default for int, this line doesn't change anything

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 8: use your type parameters - K, V

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+        int newCapacity = keys.length * 2;
+        keys = resizeArray(keys, newCapacity);
+        values = resizeArray(values, newCapacity);
+    }
+
+    private Object[] resizeArray(Object[] array, int newCapacity) {
+        Object[] newArray = new Object[newCapacity];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
+    }
```
okuzan, line 66: good job, but resizing is out of scope for this task :) we assume the capacity is always 10 and don't handle overflow

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+
+    private boolean areKeysEqual(K key1, Object key2) {
+        if (key1 == null) {
+            return key2 == null;
+        }
+        return key1.equals(key2);
```
okuzan, line 72: not a mistake, but we can write this in one line
```suggestion
        return key1 == key2 && (key1 != null && key1.equals(key2));
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+        values = resizeArray(values, newCapacity);
+    }
+
+    private Object[] resizeArray(Object[] array, int newCapacity) {
+        Object[] newArray = new Object[newCapacity];
+        System.arraycopy(array, 0, newArray, 0, array.length);
+        return newArray;
+    }
+
+    private boolean areKeysEqual(K key1, Object key2) {
```
okuzan, line 68:
```suggestion
    private boolean areKeysEqual(K key1, K key2) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+        int index = indexOf(key);
+        return (index != -1) ? (V) values[index] : null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
```
okuzan, line 41: good job on extracting common logic to a private method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,72 @@
...
+        for (int i = 0; i < size; i++) {
+            if (areKeysEqual(key, keys[i])) {
+                return i;
+            }
+        }
         return -1;
```
okuzan, line 47: it can be perceived as a magic number, so let's create a NOT_FOUND_INDEX and use it

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,98 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[DEFAULT_CAPACITY];
+        values = (V[]) new Object[DEFAULT_CAPACITY];
+        size = 0;
```
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,98 @@
...
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
+        }
+        return -1;
```
olekskov, line 54: Please combine condition to work with both null key and nonNull key
With this approach you will not need to duplicate and handle null and nonNull cases separately
```suggestion
    private int indexOf(K key) {
        for (int i = 0; i < size; i++) {
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
                return i;
            }
        }
        return -1;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,98 @@
...
+        int nullKeyIndex = indexOfNullKey();
+        return (nullKeyIndex != -1) ? values[nullKeyIndex] : null;
+    }
+
+    private int indexOfNullKey() {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                return i;
+            }
+        }
```
olekskov, line 97: Null key should be handled in same way as any other key
```suggestion
```

olekskov (changes requested):
Please simplify code by handling null key same as any other one

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
...
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = indexOf(key);
+        return (index != -1) ? (V) values[index] : null;
```
Ivan95kos, line 34:
```suggestion
        return (index != -1) ? values[index] : null;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,62 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
Ivan95kos, line 44: Don't use class Objects

Ivan95kos (approved):
For the future, read carefully the common mistakes

PR #1381, opened 2024-01-12
added solution to the task

yanna-stepanova:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,85 @@
...
     public int size() {
-        return -1;
+        int count = 0;
+        for (Pair pair : pairs) {
+            if (pair != null) {
+                count++;
+            }
+        }
+        return count;
+    }
```
Ivan95kos, line 64: add the field `size` to StorageImpl
```suggestion
        return size;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,85 @@
...
     @Override
     public V get(K key) {
+        int length = size();
+        for (int i = 0; i < length; i++) {
+            if ((pairs[i].myKey == null && key == null)
+                    || ((pairs[i].myKey != null && key != null)
+                           && pairs[i].myKey.equals(key))) {
+                return (V) pairs[i].myValue;
+            }
+        }
```
Ivan95kos, line 51: try to simplify this method
```suggestion
        for (int i = 0; i < size; i++) {
            if (compare(keys[i], key)) {
                return values[i];
            }
        }
```
and add the next method:
```
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,85 @@
...
+                for (int i = 0; i < filledSize; i++) {
+                    if ((pairs[i].myKey == null && key == null)
+                            || ((pairs[i].myKey != null && key != null)
+                                    && pairs[i].myKey.equals(key))) {
+                        pairs[i].myValue = value;
+                    }
+                }
+            }
+        }
+
```
Ivan95kos, line 39: This method is overloaded, try to simplify it
like:
```
    @Override
            if (compare(keys[i], key)) {
                values[i] = value;
                return;
            }
        }
        keys[size] = key;
        values[size] = value;
        size++;
    }
```

Ivan95kos (changes requested):
left a few comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER = 10;
+    private static final int ZERO = 0;
+    private int size;
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        this.size = ZERO;
```
Ivan95kos, line 13: `int` has the default value `0`
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        this.keys = (K[]) new Object[MAX_NUMBER];
+        this.values = (V[]) new Object[MAX_NUMBER];
+    }
+
+    private boolean compare(K keyLeft, K keyRight) {
```
Ivan95kos, line 18: All private methods should be below public methods

Ivan95kos (approved):
Great job with minor comments

PR #1382, opened 2024-01-13
Implemented the `Storage` of the `key-value` type that is parameteriz…

Ovikt0r:
…ed with two types of data.


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[INITIAL_CAPACITY];
+        values = (V[]) new Object[INITIAL_CAPACITY];
+        size = 0;
```
Ivan95kos, line 18: `int` has the default value `0`
```suggestion
```

Ivan95kos (approved):
Great job

src/main/java/core/basesyntax/impl/StorageImpl.java
Ovikt0r: Ok, got it. I'll remember it next time

PR #1385, opened 2024-01-19
fixed

LizaBalshv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private Object[] keys;
+    private Object[] values;
```
```suggestion
    private K[] keys;
    private V[] values;
```
that's where we should use type parameters

okuzan (changes requested):
Good job

okuzan (changes requested):
for sync, ignore this comment

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (size < MAX_ITEMS_NUMBER) {
+            for (int i = 0; i < size; i++) {
+                if (key == null && keys[i] == null || key != null && key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
```
okuzan, line 24: image
repeated logic, please extract to a private method
you can return an index of an equal key, -1 if not found

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            int index = findKeyIndex(key);
+            if (index != -1) {
+                values[index] = value;
+            } else {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            }
+        } else {
+            System.out.println("The storage is full, it is impossible to make a new entry");
```
Ivan95kos, line 28: For the future, it is not good practice to log data using such a way,
It is better to throw an Exception or use the logger
```suggestion
```

Ivan95kos (approved):
Great job with a minor comment

PR #1386, opened 2024-01-19
release StorageImpl

u3IOm4uk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE_STORAGE = 10;
+    private int sizeStorage;
```
Ivan95kos, line 7: It is not good practice to give the name of the Class
```suggestion
    private int size;
```

Ivan95kos (approved):
Great job with a minor commnet

PR #1392, opened 2024-01-27
Implemented the Storage

ynik404:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((storageForKeys[i] == null && storageForKeys[i] == key)
+                    || (storageForKeys[i] != null && storageForKeys[i].equals(key))) {
```
okuzan, line 42: extract repeated logic to a separate private utility method
```suggestion
            if ((storageForKeys[i] == key)
                    || (storageForKeys[i] != null && storageForKeys[i].equals(key))) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+                    && storageForValue[i] != value)) {
+                storageForValue[i] = value;
+                break;
+            }
+            if (storageForKeys[i] == null && storageForKeys[i] == key
+                    && storageForValue != value) {
+                storageForValue[i] = value;
+                break;
+            }
+        }
```
okuzan, line 35: you have the same logic in `if`s body, simplify/combine them, you only need one

okuzan (changes requested):

sokimaaa (approved):

PR #1394, opened 2024-01-29
implemented custom map storage

ghost:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int CAPACITY = 10;
+
+    private Object[] keys;
+    private Object[] values;
```
```suggestion
    private K[] keys;
    private V[] values;
```

That's where we should use class type parameters

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    public StorageImpl() {
+        keys = new Object[CAPACITY];
+        values = new Object[CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexOfKey = getKeyIndex(key);
+        if (indexOfKey == -1) {
```
okuzan, line 21: let's extract it to a NOT_FOUND_INDEX

okuzan (changes requested):
good job

kozhukhovsky (approved):

PR #1395, opened 2024-01-29
implemented the StorageImpl class according to the task requirements

DanilKi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private Pair<K, V>[] pairs;
+
+    private static class Pair<K, V> {
```
olekskov, line 9: Please move private inner class after public methods to have main functionality first

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+        pairs = new Pair[STORAGE_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < pairs.length; i++) {
+            if (pairs[i] != null) {
+                if (pairs[i].getKey() == null ? key == null : pairs[i].getKey().equals(key)) {
+                    pairs[i].setValue(value);
+                    break;
```
olekskov, line 41: result is same but return instead of break will emphasize that we're done with logic
```suggestion
                    return;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+                }
+            } else {
+                pairs[i] = new Pair<>(key, value);
+                break;
```
olekskov, line 45: same
```suggestion
                return;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
     @Override
     public int size() {
-        return -1;
+        int result = 0;
+        for (Pair<K, V> pair : pairs) {
+            if (pair != null) {
+                result++;
+            }
+        }
+        return result;
```
olekskov, line 70: Please introduce class variable and increment it after adding new element in order not to calculate size each time by iterating
```suggestion
        return size;
```

olekskov (changes requested):
Nice overall but there are some comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private Pair<K, V>[] pairs;
+    private int size;
+
+    public StorageImpl() {
+        pairs = new Pair[STORAGE_CAPACITY];
+        size = 0;
```
olekskov, line 12: 0 is default int value, no need to explicitly set it
```suggestion
```

olekskov (approved):
Good job

PR #1401, opened 2024-02-04
Hm jv storage

vladshender:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUE = 10;
+    private static final int VALUE_NOT_FOUND = -1;
+    private K[] keyList;
+    private V[] valueList;
+    private int count = 0;
```
okuzan, line 10: 0 is default for numbers

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUE = 10;
+    private static final int VALUE_NOT_FOUND = -1;
+    private K[] keyList;
+    private V[] valueList;
```
```suggestion
    private K[] keys;
    private V[] values;
```
moreover it's not even a list, but array

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
     @Override
     public int size() {
-        return -1;
+        return count;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size(); i++) {
+            if ((keyList[i] == null && key == null)
+                    || (keyList[i] != null && keyList[i].equals(key))) {
```
```suggestion
            if (keyList[i] == key
                    || keyList[i] != null && keyList[i].equals(key)) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getIndex(key);
+        if (index != VALUE_NOT_FOUND) {
+            return (V) valueList[index];
+        } else {
+            return null;
+        }
```
```suggestion
        } 
        return null;
```
we don't need else after return

okuzan (changes requested):
Good job! minor comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_VALUE = 10;
+    private static final int VALUE_NOT_FOUND = -1;
+    private K[] keys;
+    private V[] values;
+    private int count;
```
okuzan, line 10:
```suggestion
    private int size;
```

okuzan (approved):
Good job

PR #1403, opened 2024-02-04
jv-storage solution

c-kalenyk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+
+    public StorageImpl() {
+        pairs = new Pair[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (Pair pair : pairs) {
+            if (pair != null) {
+                if (key == null && pair.getKey() == null && pair.getValue() != null
```
Rommelua, line 18:
```suggestion
                if (key == pair.getKey() 
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+    private int currentSize;
+
+    public StorageImpl() {
+        pairs = new Pair[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (Pair pair : pairs) {
+            if (pair != null) {
```
Rommelua, line 17: Can we have loop from 0 up to currentSize and remove this  `if (pair != null)`?

Rommelua (changes requested):

PR #1404, opened 2024-02-04
implemented put/get/size methods and add some new methods

Bohadan-Markatov:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        public StorageObject(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getIndexByKey(key);
+        if (index != -1) {
```
Elena-Bruyako, line 27:
```suggestion
        if (index != NO_INDEX) {
```
let's create constant for magic number

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        int index = getIndexByKey(key);
+        return index != -1 ? storage[index].value : null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private boolean keyCheck(K key, K storageObjectKey) {
```
Elena-Bruyako, line 50: To make the code readable it's better to replace this method after getIndexByKey()

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_LENGTH = 10;
+    private int size;
```
Elena-Bruyako, line 7: It's better to initialise the variable in the same place where you declare it or in constructor

Elena-Bruyako (changes requested):
Good job! Let's improve your solution!

andrii-hoienko (approved):
Very nice!

PR #1405, opened 2024-02-04
done

lskErixon:


src/main/java/core/basesyntax/impl/Box.java
```diff
@@ -0,0 +1,55 @@
+package core.basesyntax.impl;
+
+import java.util.Objects;
+
+public class Box {
```
Rommelua, line 5: delete this class please

src/main/java/core/basesyntax/impl/StorageFullException.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax.impl;
+
+public class StorageFullException extends Exception {
```
Rommelua, line 3:
```suggestion
public class StorageFullException extends RuntimeException {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,119 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_SIZE = 10;
```
Rommelua, line 9: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,119 @@
...
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+        size = 0;
+    }
```
Rommelua, line 18: 0 is default value
```suggestion
    public StorageImpl() {
        keys = new Object[MAX_SIZE];
        values = new Object[MAX_SIZE];
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,119 @@
...
+            if (size < MAX_SIZE) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new StorageFullException("Storage is full. Cannot add more items.");
+            }
+        } catch (StorageFullException e) {
+            System.out.println(e.getMessage());
+        }
```
Rommelua, line 44:
```suggestion
            if (size == MAX_SIZE) {
                throw new StorageFullException("Storage is full. Cannot add more items.");
            }
            keys[size] = key;
            values[size] = value;
            size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,119 @@
...
+            return false;
+        }
+        return size == storage.size && Arrays.equals(keys, storage.keys)
+                && Arrays.equals(values, storage.values);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(size);
+        result = 31 * result + Arrays.hashCode(keys);
```
Rommelua, line 115: delete these methods

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size = 0;
```
kerrrusha, line 9: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,88 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
kerrrusha, line 8:
```suggestion
    private K[] keys;
    private V[] values;
```

kerrrusha (changes requested):

PR #1406, opened 2024-02-05
added Storage Impl.

vfilimo:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys = (K[]) new Object[MAX_SIZE];
+    private V[] values = (V[]) new Object[MAX_SIZE];
+    private int size = 0;
```
kerrrusha, line 9: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

kerrrusha (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+    private static final int MAX_SIZE = 10;
+    private K[] keys = (K[]) new Object[MAX_SIZE];
+    private V[] values = (V[]) new Object[MAX_SIZE];
+    private int size;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((key == keys[i]) || ((key != null) && key.equals(keys[i]))) {
+                values[i] = value;
```
okuzan, line 15: this logic is common to both methods, extract it to a private method (DRY)

okuzan (changes requested):

olekskov (approved):
Good job

PR #1409, opened 2024-02-09
version 1

ChabVlad:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    public static final int MAX_SIZE = 10;
+    public static final int KEY_VALUE = 2;
```
okuzan, line 8: no empty lines after class/method definition is needed
why public?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+    private Object[][] storage = new Object[MAX_SIZE][KEY_VALUE];
```
okuzan, line 9: i'm not sure 2D array is a better choice than just two separate arrays for keys and values, it's hard to read, you would need to introduce constants like VALUE_INDEX, KEY_INDEX to use instead of literals
also, use your type parameters - K, V

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
     @Override
     public int size() {
-        return -1;
+        int count = 0;
+        for (Object[] o : storage) {
+            if (o[1] != null) {
+                count++;
+            }
+        }
+        return count;
```
okuzan, line 55: you should store size as a separate field

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
     }
 
     @Override
     public V get(K key) {
+
+        if (size() > MAX_SIZE) {
+            return null;
+        }
+        for (int j = 0; j < MAX_SIZE; j++) {
+            if (storage[j][0] == null && storage[j][1] != null && key == null) {
```
okuzan, line 37: extract equals condition to a separate method

okuzan (changes requested):

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                if (values[i] != null) {
+                    size--;
+                }
+                values[i] = value;
+                break;
+            }
```
olekskov, line 25: If we use return instead of break we don't have to decrement size that will be incremented further
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                values[i] = value;
                return;
            }
```

olekskov (approved):
Good job overall, approved

PR #1410, opened 2024-02-10
solve the storage task

esidarapp:


okuzan (changes requested):
make sure to go through this first:  https://github.com/mate-academy/jv-storage/blob/master/checklist.md

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int KEY_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private static final int VALUE_AND_INDEX = 2;
```
Vasul-Pankiv, line 9:
```suggestion
    private static final int PAIR_SIZE = 2;
```
But in my opinion not the best solutiion, try to create an Pair<K,V> nested class and use it  (its on you)

Vasul-Pankiv (approved):

PR #1411, opened 2024-02-12
Resolved jv-storage Task

DmitryLazyuka:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[10];
+        this.values = (V[]) new Object[10];
```
olekskov, line 13: Please move magic number to constant with proper naming

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+                return;
+            } else if (!Objects.isNull(keys[i]) && Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            } else if (Objects.isNull(key) && Objects.isNull(keys[i])
+                    && !Objects.isNull(values[i])) {
+                values[i] = value;
+                return;
+            }
+        }
```
olekskov, line 32: Usage of Objects class methods is prohibited in this task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (!Objects.isNull(key) && Objects.equals(keys[i], key)) {
+                return values[i];
+            } else if (Objects.isNull(key) && Objects.isNull(keys[i])
+                    && !Objects.isNull(values[i])) {
+                return values[i];
+            }
+        }
```
olekskov, line 44: same here

olekskov, line 44: And we should deal with null key same as any other. For us it doesn't matter
```suggestion
        for (int i = 0; i < keys.length; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return values[i];
            }
        }    
```
Please use similar approach in put method

olekskov (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[INITIAL_CAPACITY];
+        this.values = (V[]) new Object[INITIAL_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (values[i] == null) {
```
Vasul-Pankiv, line 19: what if you want to put couple diff keys with null value?

Vasul-Pankiv (changes requested):

kozhukhovsky (approved):

PR #1412, opened 2024-02-12
Solved jv-storage task:

Shimeriyu:
created implementation of all methods in StorageImpl class. All tests passed.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,59 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 11;
+    //1 reserved for null, so limit set as 11 in case if none of the Keys is null
+    private static final int NULL_DEFAULT_INDEX = 0;
+    private int size = 0;
+    private int[] indexes;
+    private Object[] values;
+
+    public StorageImpl() {
+        values = new Object[MAX_ITEMS_NUMBER];
+        indexes = new int[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            if (values[NULL_DEFAULT_INDEX] == null) {
+                size++;
+            }
+            values[0] = value;
+            return;
+        }
+
+        int index = key.hashCode();
+        for (int i = 1; i < 9; i++) {
+            if (indexes[i] == 0) {
+                indexes[i] = index;
+                values[i] = value;
+                size++;
+                break;
+            }
+            if (indexes[i] == index) {
+                values[i] = value;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            return (V)values[NULL_DEFAULT_INDEX];
+        }
+        int index = key.hashCode();
+        for (int i = 1; i < 9; i++) {
+            if (indexes[i] == index) {
+                return (V)values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 11;
+    //1 reserved for null, so limit set as 11 in case if none of the Keys is null
+    private static final int NULL_DEFAULT_INDEX = 0;
+    private int size = 0;
+    private int[] indexes;
+    private Object[] values;
```
Rommelua, line 11: Let's use these fields
```suggestion
    private static final int MAX_STORAGE_SIZE = 10;
    private int size;
    private K[] keys;
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+
+    public StorageImpl() {
+        values = new Object[MAX_ITEMS_NUMBER];
+        indexes = new int[MAX_ITEMS_NUMBER];
```
Rommelua, line 15:
```suggestion
        keys = (K[]) new Object[MAX_STORAGE_SIZE];
        values = (V[]) new Object[MAX_STORAGE_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
     public void put(K key, V value) {
+        if (key == null) {
+            if (values[NULL_DEFAULT_INDEX] == null) {
+                size++;
+            }
+            values[0] = value;
+            return;
+        }
+
+        int index = key.hashCode();
```
Rommelua, line 28: Using hashCode may cause issues in case of collisions.

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            } else if (keys[i].equals(key)) {
+                values[i] = value;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < 10; i++) {
```
nowickimj, line 35: same here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private V[] values;
+
+    public StorageImpl() {
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < 10; i++) {
```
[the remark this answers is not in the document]
nowickimj: Additionally, why do you go through the max size here instead of the actual one?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            if (key == null || keys[i] == null) {
+                if (keys[i] == null && (key == null || values[i] == null)) {
```
nowickimj, line 20: this if statements seem over-complicated. Please provide more optimal solution e.g. don't check `keys[i] == null` twice.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+                    keys[i] = key;
+                    size += values[i] == null ? 1 : 0;
+                    values[i] = value;
+                    break;
```
nowickimj, line 24: Since the method type is `void`, you can easily use `return` once the operation is completed, additionally you don't need "else if" statement afterwards.

nowickimj (changes requested):
Please think of how you can simplify the insert operation, for example:
1. Go through all elements
2. Once you find one to replace, do it and exit
3. If you don't, check if new object can be inserted (hasn't size reached maximum), possibly come up with handling for such event (for example throw a RuntimeException).
4. Insert object

src/main/java/core/basesyntax/impl/StorageImpl.java
Shimeriyu: It is that way becouse i didn't think of using size variable in for loop so i had more cases to check. Thanks to that it is now a lot simpler

src/main/java/core/basesyntax/impl/StorageImpl.java
Shimeriyu: Else if is protecting me from null pointer exception when only 1 of key and keys[i] is null. I could write code for every possible case {null, null} {null, not null} {not null, null} {not null, not null} and get rid of "else" statement but i think its better that way because its almost halving my code. Or im missing something?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private int size = 0;
```
Rommelua, line 7: 0 is a default value
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null || key == null) {
+                if (keys[i] == key) {
+                    values[i] = value;
+                    return;
+                }
+            } else if (keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
```
Rommelua, line 27:
```suggestion
            if (keys[i] == key 
                || keys[i] != null && keys[i].equals(key)) {
                values[i] = value;
                return;
            }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                if (key == null) {
+                    return values[i];
+                }
+            } else if (keys[i].equals(key)) {
```
Rommelua, line 41: can be simplified as in `put(K key, V value)` method

Rommelua (changes requested):
See how we can improve this solution

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key
+                    || keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        keys[size] = key;
```
nowickimj, line 25: Like I've mentioned in #3 on previous review - you must check the boundaries. Otherwise, if you add new elements, how do you know if you haven't exceeded the maximum size of storage?

nowickimj (approved):

PR #1414, opened 2024-02-13
added Implement the Storage of the key-value, methods put, get, size.

MargoKateDi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 14: when you declare an integer field, it's size is 0 by default, this line doesn't change anything

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
okuzan, line 8: that's why we parametrized our class with K and V, so we don't have to use Object :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        }
```
okuzan, line 33: merge this into one block - for with if-else inside, Use the same condition as in Objects.equals
a == b OR a != null AND a.equals(b)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null) {
+                return (V) values[i];
+            }
+            if (key != null && key.equals(keys[i])) {
+                return (V) values[i];
+            }
```
okuzan, line 52: same here, merge into one
then, you will have code duplication in `put` and `get`  - extract it to a private utility method

okuzan (changes requested):
good job

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+        }
+
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new IllegalStateException("Storage is full");
+        }
```
kerrrusha, line 31: better to have less identation as possible
```suggestion
        if (size >= MAX_SIZE) {
            throw new IllegalStateException("Storage is full");
        } 
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null || key != null && key.equals(keys[i])) {
```
kerrrusha, line 37:
```suggestion
            if (key == null && keys[i] == null || key.equals(keys[i])) {
```

kerrrusha (changes requested):

PR #1416, opened 2024-02-16
Storage interface was implemented

slizko1:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_STORAGE_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_STORAGE_SIZE];
```
SkillfulDev, line 8: initialization is best done through the constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if ((keyArray[i] == null && key == null)) {
+                return valueArray[i];
+            }
+            if (keyArray[i] != null && keyArray[i].equals(key)) {
+                return (V) valueArray[i];
```
SkillfulDev, line 50: we do not need to explicitly lead to type V because the array is already initialized by this type  private V[] valueArray = (V[]) new Object[MAX_STORAGE_SIZE];

SkillfulDev, line 50: These two "if" can be combine because they return the same value
```suggestion
             if (keyArray[i] == null && key == null
                    || keyArray[i] != null && keyArray[i].equals(key)) {
                return valueArray[i];
```

SkillfulDev (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
slizko1: I don’t know if there is a need to create arrays through a constructor in this case

SkillfulDev (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+            }
+
+            if (i >= storageSize && keyArray[i] == null) {
+                this.setKeyAndValue(key, value, i, true);
+                return;
+            }
+        }
+    }
+
+    private void setKeyAndValue(K key, V value, int index, boolean incrementStorageSize) {
```
Elena-Bruyako, line 31: private methods should be placed after all public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_STORAGE_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_STORAGE_SIZE];
+    private int storageSize = 0;
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
```
Elena-Bruyako, line 13: FYI
Method put could look more elegant 

```
if (size == MAX_SIZE) {
            return;
        }
        int index = indexOf(key);
        if (index != NO_INDEX) {
            values[index] = value;
        } else if (index == NO_INDEX) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```


```
private int indexOf(K key) {
                return i;
            }
        }
        return NO_INDEX;
    }
```

private static final int NO_INDEX = -1;

Elena-Bruyako (approved):
Good job! Write comment just FYI

PR #1418, opened 2024-02-19
implemented methods from Storage interface

HAIOVYI:


slizko1 (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENT = 10;
+    private K[] keyArray = (K[]) new Object[MAX_ELEMENT];
+    private V[] valueArray = (V[]) new Object[MAX_ELEMENT];
+    private int size;
```
slizko1, line 9: It will be better if you Initialize  keyArray and valueArray throw the constructor. But in this keys it is not really matter, i discus that at today's second Q&A.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (key == null && keyArray[i] == null) {
+                return valueArray[i];
+            }
+            if (key != null && keyArray[i] != null && keyArray[i].equals(key)) {
+                return valueArray[i];
```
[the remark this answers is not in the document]
HAIOVYI: thx, fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
HAIOVYI: thx

slizko1 (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENT = 10;
+    private K[] keyArray = (K[]) new Object[MAX_ELEMENT];
```
Elena-Bruyako, line 7: [All non-constant fields should be initialized in the constructor](https://github.com/mate-academy/jv-storage/blob/master/checklist.md#all-non-constant-fields-should-be-initialized-in-the-constructor)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+                valueArray[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (key != null && keyArray[i] != null ? key.equals(keyArray[i]) : key == keyArray[i]) {
```
Elena-Bruyako, line 32: It's really bad idea to use ternary operation in if statement. 
I suggest you to create separate method `private int indexOf(K key)` and use it here and in put method to avoid duplication and using for loop twice in your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,41 @@
...
+                return valueArray[i];
```
Elena-Bruyako, line 33: After you create indexOf method let's use the result of it
```suggestion
               return index == NO_INDEX ? null : values[index];
```

where private static final int NO_INDEX = -1;

Elena-Bruyako (changes requested):
Good job! Let's improve your solution!

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: Not resolved

andrii-hoienko (changes requested):
Please resolve all previous comments

src/main/java/core/basesyntax/impl/StorageImpl.java
HAIOVYI: fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
HAIOVYI: fixed

Elena-Bruyako (approved):

PR #1427, opened 2024-02-23
feat: implement Storage

oksana-divkit:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,80 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int DEFAULT_CAPACITY = 10;
+    private static final int INCREASE_FACTOR = 2;
```
VlasenkoYevhenii, line 7: You don't need to increase the storage size, 'cause it's not the default size, it's the max one.
I won't mention it any further, just check the masterpiece you created. But you got it =)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,80 @@
...
+    private Object[] keys;
+    private Object[] values;
+    private int currentCapacity;
+    private int currentSize;
+
+    public StorageImpl() {
+        keys = new Object[DEFAULT_CAPACITY];
+        values = new Object[DEFAULT_CAPACITY];
+        currentCapacity = DEFAULT_CAPACITY;
+        currentSize = 0;
```
VlasenkoYevhenii, line 18: int field will be 0 without initialization ( default int value)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,80 @@
...
     public V get(K key) {
-        return null;
+        V value = null;
+        int index = getIndexOfKey(key);
+
+        if (index >= 0) {
+            value = (V) values[index];
+        }
+        return value;
     }
```
VlasenkoYevhenii, line 40: Can we refactor this method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,80 @@
...
+
     @Override
     public void put(K key, V value) {
+        int index = getIndexOfKey(key);
+        if (index < 0) {
+            addNew(key, value);
+        } else {
+            values[index] = value;
+        }
     }
```
VlasenkoYevhenii, line 29: Can we refactor this method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,80 @@
...
+        currentCapacity *= INCREASE_FACTOR;
+
+        keys = new Object[currentCapacity];
+        values = new Object[currentCapacity];
+
+        for (int i = 0; i < currentSize; i++) {
+            keys[i] = keysTemp[i];
+            values[i] = valuesTemp[i];
+        }
     }
```
VlasenkoYevhenii, line 81: Excessive logic

VlasenkoYevhenii (changes requested):
Good job! 
There are some comments. Check them pls

src/main/java/core/basesyntax/impl/StorageImpl.java
oksana-divkit: yes, it's possible. I just don't like several returns in the methods

src/main/java/core/basesyntax/impl/StorageImpl.java
oksana-divkit: How? :)

andrii-hoienko (changes requested):
Good job, well done for making the logic for expanding the storage, left a couple of small comments

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: I think it is about the `else`  keyword, would be nice to structure our logic in a way to avoid `else`s by using the `return` or other things, because it usually brings complexity in readability
```suggestion
    @Override
    public void put(K key, V value) {
        int index = getIndexOfKey(key);
        if (index < 0) {
            addNew(key, value);
            return;
        }
        values[index] = value;
    }
```

karma-o (approved):
Very nice! Left a comment about your question

PR #1428, opened 2024-02-23
Implemented storage class

ghost:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
-    public void put(K key, V value) {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_ARRAY_SIZE];
+        values = new Object[MAX_ARRAY_SIZE];
     }
```
fedaktaras, line 14: As there is no logic in constructor, arrays might be initialised instantly

Object[] keys = new Object[MAX_ARRAY_SIZE];
Object[] values = new Object[MAX_ARRAY_SIZE];

fedaktaras (changes requested):
Really good and short code.
Can be a little shorter)

src/main/java/core/basesyntax/impl/StorageImpl.java
VlasenkoYevhenii: that's just the way I'm used to. But in this particular case, you're right

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
-    }
+    private static final int MAX_ARRAY_SIZE = 10;
+    private final Object[] keys = new Object[MAX_ARRAY_SIZE];
+    private final Object[] values = new Object[MAX_ARRAY_SIZE];
```
karma-o, line 8: It is better to have fields typed properly right away, rather than casting the return values :wink: 
```suggestion
    private final K[] keys = (K[]) new Object[MAX_ARRAY_SIZE];
    private final V[] values = (V[]) new Object[MAX_ARRAY_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+        for (int i = 0; i < size; i++) {
+            if (compareProvidedKey(key, i)) {
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
+    @Override
+    public void put(K key, V value) {
```
karma-o, line 22: we can also add a size validation here if our storage already contains 10 elements - throw an exception

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+    private final Object[] values = new Object[MAX_ARRAY_SIZE];
+    private int size;
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (compareProvidedKey(key, i)) {
+                return (V) values[i];
+            }
+        }
```
karma-o, line 17: there is a common logic in `put()` and `get()` methods, where you are looking for an element by the key, it can be extracted to a separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,40 @@
...
+        size++;
+    }
+
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean compareProvidedKey(K key, int index) {
```
karma-o, line 39: I think you can actually use something like `Objects.equals(key, keys[index])` instead of this method,

karma-o (changes requested):
Very good! left some minor comments

src/main/java/core/basesyntax/impl/StorageImpl.java
VlasenkoYevhenii: It's restricted by the common mistakes list.

src/main/java/core/basesyntax/impl/StorageImpl.java
VlasenkoYevhenii: Nope =)
There are just some similar lines

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: My bad, didn't see it :sweat_smile:

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: I do not agree, both methods have a for loop and a check for keys, the body is different and this is where we should do a bit of refactoring.
What we need there is to return the index of that key, then we can use that index in both methods to get the value by that index in `get()` and set a new value for that index in `put()`. We also need to handle the case when we do not have such a key yet, but I think I am gonna leave this part for you to implement :wink:

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            }
+        }
         return null;
     }
 
+    @Override
+    public void put(K key, V value) {
+        if (size == MAX_ARRAY_SIZE && !containsKey(key)) {
+            throw new IllegalStateException(FULL_ARRAY_EXCEPTION_MESSAGE);
+        }
```
karma-o, line 26: Move it after the for loop, then we won't need that `containsKey()` method at all

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
-    }
+    private static final int MAX_ARRAY_SIZE = 10;
+    private static final String FULL_ARRAY_EXCEPTION_MESSAGE = "The storage is full";
```
karma-o, line 7: add max size to the message

karma-o (changes requested):
Please check the comment about the common logic in get and put methods. Ask in the chat if you need any help

karma-o (approved):
Nice! Left minor comments

PR #1429, opened 2024-02-23
Implement parameterized key-value storage using arrays

VolandevlodD:
This commit implements a Storage interface for key-value pairs, where keys are of type Integer and values are of type Box. The StorageImpl class utilizes one or two arrays, depending on the implementation. When adding a new key-value pair using the put method, if the key already exists, the value is replaced. The get method retrieves the value associated with a given key. Additionally, the size method returns the current size of the storage.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,56 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size = 0;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_ITEMS_NUMBER];
+        this.values = new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        int keyIndex = getKeyIndex(key);
+        if (keyIndex != -1) {
+            values[keyIndex] = value;
+            return;
+        }
+
+        if (size < MAX_ITEMS_NUMBER) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full!");
+        }
+    }
+
+    private int getKeyIndex(K keyToCompare) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], keyToCompare)) {
+                return i;
+            }
+        }
+        return -1;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getKeyIndex(key);
+        if (index == -1) {
+            return null;
+        }
+        return (V) values[index];
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


makaroshyna (changes requested):
Looks good, but in some places confusing to read.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full!");
+        }
+    }
+
+    private int getKeyIndex(K keyToCompare) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], keyToCompare)) {
```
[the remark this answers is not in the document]
VolandevlodD: Thanks for reviewing, it looks like IDEA's autoformatting formatted the manual comparison line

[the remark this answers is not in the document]
VolandevlodD: Resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
```
karma-o, line 6: just tweak the name a bit and add a space between static and non-static fields
```suggestion
    private static final int MAX_SIZE = 10;
    
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private final Object[] keys;
+    private final Object[] values;
+    private int size = 0;
```
karma-o, line 9: primitive class fields have default value, for int it is 0, so we can remove that
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
karma-o, line 8: It is better to type the fields properly, check next comment for fixing the casts
```suggestion
    private final K[] keys;
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private int size = 0;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_ITEMS_NUMBER];
+        this.values = new Object[MAX_ITEMS_NUMBER];
```
karma-o, line 13: this way you can initialize it properly and not cast the fields on returning them
```suggestion
        this.keys = (K[]) new Object[MAX_ITEMS_NUMBER];
        this.values = (V[]) new Object[MAX_ITEMS_NUMBER];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        if (size < MAX_ITEMS_NUMBER) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full!");
+        }
+    }
+
+    private int getKeyIndex(K keyToCompare) {
```
karma-o, line 33: private methods go after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == keyToCompare) || (keys[i] != null && keys[i].equals(keyToCompare))) {
+                return i;
+            }
+        }
+        return -1;
```
karma-o, line 39: magic number, check its usage in other places too

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            values[keyIndex] = value;
+            return;
+        }
+
+        if (size < MAX_ITEMS_NUMBER) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full!");
```
karma-o, line 29: let's create custom exception or use something like `IllegalStateException` instead + add max size to the message to that a user of that class would have more information on what went wrong

karma-o (changes requested):
GJ! Left some style comments

karma-o:
BTW, commit messages are usually not that long, 2 sentences max. If you need to give more information about the changes that were made, do that in the PR description or split the changes into several commits

andrii-hoienko (approved):
Very nice

PR #1430, opened 2024-02-23
Task solved

fedaktaras:
Task solved with using inner class StorageElement<K, V>

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -4,31 +4,31 @@ import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     private static final int STORAGE_MAX_CAPACITY = 10;
-    private int size = 0;
+    private int size;
     private final StorageElement<K, V>[] storageElements = new StorageElement[STORAGE_MAX_CAPACITY];
 
     @Override
     public void put(K key, V value) {
-        StorageElement<K, V> storageElement = new StorageElement<>(key, value);
+
         for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
             if (storageElements[i] == null) {
                 size++;
-                storageElements[i] = storageElement;
+                storageElements[i] = new StorageElement<>(key, value);
                 return;
-            } else if (storageElements[i].isEmpty()
-                    || storageElements[i].hasKey(storageElement.getKey())) {
-                storageElements[i] = storageElement;
+            } else if (storageElements[i].hasKey(key)) {
+                storageElements[i].value = value;
                 return;
             }
         }
-        throw new RuntimeException("StorageImpl has no more space.");
+        throw new IllegalStateException("StorageImpl has no more space. Max space is"
+                + STORAGE_MAX_CAPACITY);
     }
 
     @Override
     public V get(K key) {
         for (StorageElement<K, V> storageElement : storageElements) {
             if (storageElement != null && !storageElement.isEmpty() && storageElement.hasKey(key)) {
-                return (V) storageElement.value;
+                return storageElement.value;
             }
         }
         return null;
@@ -36,7 +36,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     @Override
     public int size() {
-        return this.size;
+        return size;
     }
 
     private static class StorageElement<K, V> {
@@ -52,18 +52,8 @@ public class StorageImpl<K, V> implements Storage<K, V> {
             return key == null && value == null;
         }
 
-        public K getKey() {
-            return key;
-        }
-
         public boolean hasKey(K key) {
-            if (key == this.key) {
-                return true;
-            }
-            if (key == null || this.key == null) {
-                return false;
-            }
-            return this.key.equals(key);
+            return key == this.key || key != null && key.equals(this.key);
         }
     }
 }
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -4,16 +4,19 @@ import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
     private static final int STORAGE_MAX_CAPACITY = 10;
+    private int size = 0;
     private final StorageElement<K, V>[] storageElements = new StorageElement[STORAGE_MAX_CAPACITY];
 
     @Override
     public void put(K key, V value) {
         StorageElement<K, V> storageElement = new StorageElement<>(key, value);
         for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
-            if (storageElements[i] == null || storageElements[i].isEmpty()) {
+            if (storageElements[i] == null) {
+                size++;
                 storageElements[i] = storageElement;
                 return;
-            } else if (storageElements[i].hasKey(storageElement.getKey())) {
+            } else if (storageElements[i].isEmpty()
+                    || storageElements[i].hasKey(storageElement.getKey())) {
                 storageElements[i] = storageElement;
                 return;
             }
@@ -25,7 +28,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
     public V get(K key) {
         for (StorageElement<K, V> storageElement : storageElements) {
             if (storageElement != null && !storageElement.isEmpty() && storageElement.hasKey(key)) {
-                return (V) storageElement.getValue();
+                return (V) storageElement.value;
             }
         }
         return null;
@@ -33,12 +36,7 @@ public class StorageImpl<K, V> implements Storage<K, V> {
 
     @Override
     public int size() {
-        for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
-            if (storageElements[i] == null) {
-                return i;
-            }
-        }
-        return STORAGE_MAX_CAPACITY;
+        return this.size;
     }
 
     private static class StorageElement<K, V> {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
     @Override
     public void put(K key, V value) {
+        StorageElement<K, V> storageElement = new StorageElement<>(key, value);
+        for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
+            if (storageElements[i] == null || storageElements[i].isEmpty()) {
+                storageElements[i] = storageElement;
+                return;
+            } else if (storageElements[i].hasKey(storageElement.getKey())) {
+                storageElements[i] = storageElement;
+                return;
```
ValentynSharshon, line 18: You have the same logic on lines 14 and 15. Perhaps you should change the condition in the if statement on line 13?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
 
     @Override
     public int size() {
-        return -1;
+        for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
+            if (storageElements[i] == null) {
+                return i;
+            }
+        }
+        return STORAGE_MAX_CAPACITY;
```
ValentynSharshon, line 41: Maybe we should declare the size variable as a class field and just increment it when adding a new object? Then you can use it in your loops so that you don't have to go through the whole StorageElement array. This is just my opinion, I think it is not critical.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+                return true;
+            }
+            if (key == null || this.key == null) {
+                return false;
+            }
+            return this.key.equals(key);
+        }
+
+        public V getValue() {
+            return this.value;
```
ValentynSharshon, line 72: You can remove this, it`s redundant.

ValentynSharshon (changes requested):
Good job, a few things could be improved in my opinion.

src/main/java/core/basesyntax/impl/StorageImpl.java
fedaktaras: I would change logic a bit. But i've seen this approach in other students solution and find it good.

src/main/java/core/basesyntax/impl/StorageImpl.java
fedaktaras: with adding and modifying size variable I have to keep two if statements.  But in this case I agree it can be done in one if statement.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,73 @@
...
+            }
+        }
+        throw new RuntimeException("StorageImpl has no more space.");
     }
 
     @Override
     public V get(K key) {
+        for (StorageElement<K, V> storageElement : storageElements) {
+            if (storageElement != null && !storageElement.isEmpty() && storageElement.hasKey(key)) {
+                return (V) storageElement.getValue();
```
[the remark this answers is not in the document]
fedaktaras: > Are you sure, that you need this cast?

Yes. discussed on Q&A

fedaktaras:
> Good job, a few things could be improved in my opinion.

And really good review. Thanks.

ValentynSharshon (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+            this.value = value;
+        }
+
+        public boolean isEmpty() {
+            return key == null && value == null;
+        }
+
+        public K getKey() {
+            return key;
+        }
```
karma-o, line 57: we can remove that getter as well, because we can reference the key/value fields of that class directly like `element.key` without using them inside `Storage` class, which is a preferred approach in case of nested classes

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+
+        public boolean hasKey(K key) {
+            if (key == this.key) {
+                return true;
+            }
+            if (key == null || this.key == null) {
+                return false;
+            }
+            return this.key.equals(key);
+        }
```
karma-o, line 67: looks too complicated I am sure we can simplify it to 1 `if`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+                return (V) storageElement.value;
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return this.size;
```
karma-o, line 39: this is redundant here, there are no other `size` named variables in scope of that method
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+            if (storageElements[i] == null) {
+                size++;
+                storageElements[i] = storageElement;
+                return;
+            } else if (storageElements[i].isEmpty()
+                    || storageElements[i].hasKey(storageElement.getKey())) {
+                storageElements[i] = storageElement;
+                return;
+            }
+        }
```
karma-o, line 23: you see, you are creating a new Storage element without considering that an element with such a key can already exist and thus it is not very effective memory-wise, what I recommend is:
* first try to find an element by the key and return it if it exists, can be a separate method
* if it does - assign a new value to the existing element object
* if it does not exist(returned element is null?) - create a new one, put it at the `size` index, and increment it

I think that would be much more readable too

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
...
+        throw new RuntimeException("StorageImpl has no more space.");
```
karma-o, line 24: would be nice to create custom exception or use something like `IllegalStateException` instead, + add max size to the message, to make it more informative

karma-o (changes requested):
Left some comments

src/main/java/core/basesyntax/impl/StorageImpl.java
fedaktaras: > you see, you are creating a new Storage element without considering that an element with such a key can already exist and thus it is not very effective memory-wise, what I recommend is:
> 
> * first try to find an element by the key and return it if it exists, can be a separate method
> * if it does - assign a new value to the existing element object
> * if it does not exist(returned element is null?) - create a new one, put it at the `size` index, and increment it
> 
> I think that would be much more readable too

Did it a bit different way. If key exists I just update value of that storage element.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+        public StorageElement(K key, V value) {
+            this.key = key;
+            this.value = value;
+        }
+
+        public boolean isEmpty() {
+            return key == null && value == null;
+        }
+
+        public boolean hasKey(K key) {
```
andrii-hoienko, line 55: hasKey? I don`t think it is a good name. Has often means to check if there is some element in some collection

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
...
+    private int size;
+    private final StorageElement<K, V>[] storageElements = new StorageElement[STORAGE_MAX_CAPACITY];
+
     @Override
     public void put(K key, V value) {
+
+        for (int i = 0; i < STORAGE_MAX_CAPACITY; i++) {
+            if (storageElements[i] == null) {
+                size++;
+                storageElements[i] = new StorageElement<>(key, value);
```
andrii-hoienko, line 16:
```suggestion
                storageElements[i] = new StorageElement<>(key, value);
                size++;
```

andrii-hoienko (changes requested):
Nice, minor comments left

src/main/java/core/basesyntax/impl/StorageImpl.java
fedaktaras: > hasKey? I don`t think it is a good name. Has often means to check if there is some element in some collection

As I understand it is more suitable for entire collection. 
Believe isEqualToKey() is better

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+                size++;
+                return;
+            } else if (storageElements[i].isEqualToKey(key)) {
+                storageElements[i].value = value;
+                return;
+            }
+        }
+        throw new IllegalStateException("StorageImpl has no more space. Max space is"
+                + STORAGE_MAX_CAPACITY);
```
karma-o, line 24: would be more handy to have a size check as a separate if, after the value replacement logic but before the new key logic, so that the exception that you throw would be close to the reason why it is thrown thus improving readability
```suggestion
        for (int i = 0; i < size; i++) {
             if (storageElements[i].isEqualToKey(key)) {
                storageElements[i].value = value;
                return;
            }
        }
        if (size >= STORAGE_MAX_CAPACITY) {
            throw new IllegalStateException("StorageImpl has no more space. Max space is: "
                + STORAGE_MAX_CAPACITY);
        }
        storageElements[size++] = new StorageElement<>(key, value);
```

karma-o (approved):
Left some style comments, GJ!

PR #1431, opened 2024-02-23
Implemented methods in class StorageImpl.

makaroshyna:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (isKeysEquals(keyArray[i], key)) {
+                return valueArray[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
```
Recdt, line 38: Is it would be easier to make size var in class fields, and return it there?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
-        return -1;
+        for (int i = MAX_SIZE - 1; i > 0; i--) {
+            if (keyArray[i] != null && valueArray[i] != null) {
+                return i + 1;
```
Recdt, line 41: MAX_SIZE - 1 and i + 1
Isn't that using magic numbers?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if ((isKeysEquals(key, keyArray[i]))
+                    || (keyArray[i] == null && valueArray[i] == null)) {
+                keyArray[i] = key;
+                valueArray[i] = value;
+                break;
```
Recdt, line 22: Maybe return in void functions would be better?
In this case maybe it`s ok, but in general?

Recdt (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,67 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int INDEX_NOT_FOUND = -1;
+
+    private final K[] keyArray;
+    private final V[] valueArray;
```
karma-o, line 10:
```suggestion
    private final K[] keysArray;
    private final V[] valuesArray;
```
or just
```suggestion
    private final K[] keys;
    private final V[] values;
```

karma-o (approved):
Looks very good! Just left some minor comments

PR #1432, opened 2024-02-23
Created storage implementation

Recdt:


nklimovych (approved):
Looks good, as for me.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int NOT_EXISTS = -1;
```
karma-o, line 6: this constant could be removed in favor of replacing it with the `< 0` check to not add another field to the class, but that is up to you :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int currentSize;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        currentSize = 0;
```
karma-o, line 15: primitive class fields have default value, for int it is 0, so we can remove that
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int isAvailablePosition = availableKey(key);
```
karma-o, line 20: `int` typed variable should not be named as a boolean and begin with `is...` 
this variable essentially represents an index of the key that we received as a parameter, so just `index` would be fine

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return currentSize;
+    }
+
+    private int availableKey(K key) {
```
karma-o, line 46: again, method name implies that it returns boolean, like it is named `isKeyAvailable()` but it in fact returns index of the key, so let's name it `findIndexForKey()`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        int isAvailablePosition = availableKey(key);
+        if (isAvailablePosition != NOT_EXISTS) {
+            values[isAvailablePosition] = value;
+            return;
+        }
+        if (currentSize < MAX_SIZE) {
```
karma-o, line 25: it seems that we would just ignore the input if we have reached the max size, let's throw an exception in that case, to let the user of our class know that it did

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            keys[currentSize] = key;
+            values[currentSize] = value;
+            currentSize++;
+        }
     }
 
     @Override
     public V get(K key) {
+        int availablePosition = availableKey(key);
```
karma-o, line 34: same here
```suggestion
        int index = availableKey(key);
```

karma-o (changes requested):
Left some naming comments

src/main/java/core/basesyntax/impl/StorageImpl.java
Recdt: Is zero wouldn't be a magic number?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -12,28 +12,29 @@ public class StorageImpl<K, V> implements Storage<K, V> {
...
+        if (index != NOT_EXISTS) {
+            values[index] = value;
             return;
         }
         if (currentSize < MAX_SIZE) {
             keys[currentSize] = key;
             values[currentSize] = value;
             currentSize++;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("The storage is full!!!");
```
Recdt, line 29: Is it would be a suitable exception? Or I need to make custom exception in thas case?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int NOT_EXISTS = -1;
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int currentSize;
```
andrii-hoienko, line 10:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: Maybe just IndexOutOfBoundsException will be ok

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: I think we can leave it as is, approach you use is also Ok

andrii-hoienko (changes requested):
GJ< left few minor comments

andrii-hoienko (approved):
GJ!

PR #1433, opened 2024-02-24
Implement StorageImpl

TonyH277:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     @Override
     public int size() {
-        return -1;
+        int counter = 0;
+        for (V element : valueArray) {
+            if (element != null) {
+                counter++;
+            }
+        }
+        return counter;
```
Nikname2303, line 45: you always return -1?

Nikname2303 (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (keyArray[i] == null ? key == null : keyArray[i].equals(key)) {
+                return valueArray[i];
+            }
+        }
```
karma-o, line 33: extract a common logic from put and get methods where you are looking for the index of the key to a separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+                return;
+            }
+        }
+        for (int j = 0; j < keyArray.length; j++) {
+            if (keyArray[j] == null && valueArray[j] == null) {
+                keyArray[j] = key;
+                valueArray[j] = value;
+                return;
+            }
+        }
```
karma-o, line 24: use this new method here, and leave just 2 cases:
* key is already present in the storage - replace value
* keay is not present add value and key and increment size,

hint: it can be done without additional loops and a single if in this method :wink:

karma-o (changes requested):
Commented some things to fix

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[STORAGE_SIZE];
+    private V[] valueArray = (V[]) new Object[STORAGE_SIZE];
+    private int size = 0;
```
andrii-hoienko, line 9:
```suggestion
    private int size; //0 is set by default
```

src/main/java/core/basesyntax/impl/StorageImpl.java
andrii-hoienko: Not resolved, you should create a method something like 'findIndexByKey' and reuse it in get and put methods

andrii-hoienko (changes requested):
Some comments is not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+
     @Override
     public void put(K key, V value) {
+        if (findIndexByKey(key) != -1) {
+            valueArray[findIndexByKey(key)] = value;
```
karma-o, line 14: find index by key is an expensive operation so we want to avoid it being called multiple times

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
+        }
+        keyArray[size] = key;
+        valueArray[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (findIndexByKey(key) != -1) {
+            return valueArray[findIndexByKey(key)];
```
karma-o, line 25: same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,42 @@
...
     }
 
+    private int findIndexByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keyArray[i] == null && key == null)
+                    || (keyArray[i] != null && keyArray[i].equals(key))) {
+                return i;
+            }
+        }
+        return -1;
```
karma-o, line 37: -1 should be a constant

karma-o (changes requested):

karma-o (approved):
Nice!

PR #1435, opened 2024-02-24
All assumptions met, code passed tests

TylTejlor:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
```
nowickimj, line 19: You don't need a separate loop for `null` case - instead, use single loop for all cases with proper null check with `if` condition.

nowickimj (changes requested):
Please refactor the duplicated code.

src/main/java/core/basesyntax/impl/StorageImpl.java
kerrrusha: not fixed, if you need help - feel free to ask in the chat

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+            }
+            if (size < MAX_SIZE) {
+                keys[size] = null;
+                values[size] = value;
+                size++;
+            } else {
+                throw new IllegalStateException("Storage is full");
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
```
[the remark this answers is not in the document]
kerrrusha: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,70 @@
...
+                size++;
+            } else {
+                throw new IllegalStateException("Storage is full");
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
```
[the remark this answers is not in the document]
kerrrusha: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[MAX_SIZE];
+        values = new Object[MAX_SIZE];
```
kerrrusha, line 13: and do related changes in other places
```suggestion
    private K[] keys;
    private V[] values;
    private int size;

    public StorageImpl() {
        keys = (K[]) new Object[MAX_SIZE];
        values = (V[]) new Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+        size = 0;
```
kerrrusha, line 14: In Java, declaring the initial value of a variable when it's first declared is optional. When you write private int size = 0;, you're explicitly setting the initial value of size to 0. However, in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway. So, writing private int size = 0; is redundant in this case because size would already be initialized to 0 by default. You could simply write private int size; to achieve the same result.

kerrrusha (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
nowickimj: More information regarding the default values for primitives: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html
obraz

nowickimj (approved):
All fixed, well done!

PR #1437, opened 2024-02-25
solved problem of class StorageImpl

Yaromyros:
I added constant for better code style. Next, I created a constructor to initialize our variables.

After that, I created 3 methods, first put(), which checks if the key exists and refreshes the value variable. If the key is not found then it checks if there is space in storage and creates a new key with the new value.

The second method get(), gets us the key and variable of that key.

And last one size return size of storage.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,51 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
     }
 
+    @SuppressWarnings("unchecked")
+    @Override
+    public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
+    }
+
+    @SuppressWarnings("unchecked")
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
```
okuzan, line 15: [where ](https://stackoverflow.com/a/2411283) do we need this?
in Java, if you don't initialize an instance variable, it gets a default value based on its type. For int, the default initial value is 0 anyway

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 9: That's what we have type parameters K,V for

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+                return;
+            }
+        }
+        if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
```
okuzan, line 33: Returning early is a [better approach](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement) in SWE - we won't need `else` after throwing, so we can unnest our code, and unnested code is easier to read

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
+    }
+
+    @SuppressWarnings("unchecked")
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 22: [checklist issue](https://github.com/mate-academy/jv-storage/blob/master/checklist.md#dont-use-class-objects) - all points must be fixed before the review

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 16: this line doesn't affect the flow of the program at all, it's redundant. All primitives have some default value. For Objects it's null. For `int` - 0
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size == keys.length) {
+            throw new RuntimeException("Storage is full");
+        }
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
okuzan, line 25: You need to write your own function for this
(a == b) OR ( a IS NOT NULL AND a.equals(b) )
parentheses aren't needed, added for clarity
extract common logic from `get` and `put` to a private method

okuzan (changes requested):

olekskov (approved):

PR #1439, opened 2024-02-25
First solution

Nikname2303:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,58 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
TonyH277, line 52: During education we should avoid using "Objects.equals()", it's better to hardcode this logic

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,58 @@
...
+            }
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = indexOf(key);
+        if (index != -1) {
+            V value = (V) values[index];
```
TonyH277, line 38: Consider parameterizing your arrays to maintain type safety and avoid casting values to "V".
 Using parameterized arrays ensures type consistency and reduces the risk of runtime errors.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,58 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
TonyH277, line 9: maybe its better to create "private K[] keys" &  "private V[] values"? Your way to create those arrays could potentially lead to a ClassCastException at runtime

TonyH277 (changes requested):
Great work on the implementation of those methods! Here are a few ideas that might improve your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (Objects.equals(keyArray[i], key)) {
+                return valueArray[i];
+            }
+        }
```
karma-o, line 34: extract the common logic from `get()` and `put()` methods of finding an index of the key into a separate method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+                return;
+            }
+        }
+        for (int j = 0; j < keyArray.length; j++) {
+            if (keyArray[j] == null && valueArray[j] == null) {
+                keyArray[j] = key;
+                valueArray[j] = value;
+                return;
+            }
+        }
```
karma-o, line 25: there is no need in 2 loops, first one would be in the new method that I mentioned, and second one is redundant once you have an index

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
     @Override
     public int size() {
-        return -1;
+        int counter = 0;
+        for (V element : valueArray) {
+            if (element != null) {
+                counter++;
+            }
+        }
+        return counter;
```
karma-o, line 46: add a `size` filed to `Storage` and update it dynamically in `put()` method, instead of this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_SIZE];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (Objects.equals(keyArray[i], key)) {
```
karma-o, line 14: `Objects` class is not allowed for use in this task, check common mistakes

karma-o (changes requested):

src/main/java/core/basesyntax/impl/Box.java
```diff
@@ -0,0 +1,4 @@
+package core.basesyntax.impl;
+
+public class Box {
```
karma-o, line 3: unused, remove this

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,47 @@
...
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INVALID_INDEX = -1;
+    private static final int MAX_SIZE = 10;
+    private K[] keyArray = (K[]) new Object[MAX_SIZE];
+    private V[] valueArray = (V[]) new Object[MAX_SIZE];
+    private int size;
+
+    public int getIndex(K key) {
```
karma-o, line 13:
```suggestion
    private int getIndexOf(K key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: not resolved

karma-o (changes requested):
1 comment is still not resolved from previous review

PR #1442, opened 2024-02-25
storage

MolchanovAlexander:


src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,50 @@
...
+        return value;
+    }
+
+    public K getKey() {
+        return key;
+    }
+
+    public int hashCode() {
+        return (key == null ? 0 : key.hashCode()) + (value == null ? 0 : value.hashCode());
+    }
```
azzilian, line 22: Not sure but I would use Prime numbers with hashCode checks

azzilian, line 22: Not sure but I would use Prime numbers with hashCode checks

src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,50 @@
...
+                || (pair.getValue() != null && pair.getValue().equals(value)));
+    }
+
+    @Override
+    public String toString() {
+        return "Pair{"
+                + "key=" + key
+                + ", value=" + value
+                + '}';
+    }
```
azzilian, line 49: Are we realy need this in final version of code? we dont use it in tests, so I think this overriding should be removed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+    private final Pair<K, V>[] storageMap = new Pair[10];
+
     public void put(K key, V value) {
+        Pair<K, V> tempPair = findValue(key);
+        if (tempPair == null) {
+            storageMap[size++] = new Pair<>(key, value);
+        } else {
+            tempPair.setValue(value);
+        }
+    }
```
azzilian, line 16: What if our size will be more than 10, yes I know that task says that it will be no more than 10, but maybe better make come checks in case if we will call this method to re-write our keys multipletimes(IMHO).

azzilian (commented):
As resume:
I spotted some unnecessary lines of code - Override to String
I think hashCode should be improved
I think there is vulnerability in increasing size without a limits(or I am bad at reading code)

azzilian (changes requested):
As resume:
I spotted some unnecessary lines of code - Override to String - this 100% sure
I think hashCode should be improved
I think there is vulnerability in increasing size without a limits(or I am bad at reading code)

azzilian (changes requested):
It asked me to review again but no changes  i guess its because of timeout?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,39 @@
...
+                return storageMap[i];
+            }
+        }
+        return null;
     }
 
     @Override
     public V get(K key) {
-        return null;
+        return findValue(key) != null ? findValue(key).getValue() : null;
```
karma-o, line 34: do not call `findValue()` 2 times, make the result of it a variable

karma-o (changes requested):

src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,43 @@
...
+class Pair<K, V> {
+    private K key;
+    private V value;
+
+    public Pair(K key, V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    public V getValue() {
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,79 @@
...
+        if (tempPair == null) {
+            pairs[size++] = new Pair<>(key, value);
+            return;
+        }
+        tempPair.setValue(value);
+    }
+
+    private Pair<K, V> findPairByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if (pairs[i].getKey() != null && pairs[i].getKey().equals(key)
```
andrii-hoienko, line 20: It would be great if you would extract this boolean expression to method, also it could be reused in Pair class

andrii-hoienko (changes requested):
Very nice, left minor comments

andrii-hoienko (approved):
Nice

PR #1443, opened 2024-02-26
Added contructor StrorageImp

azzilian:
Added to values array
put - search for key and add or rewrite existing key get -added check on null - if its null return value anyway Added equals and hashcode without Object

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,81 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+    private int result;
+    private boolean keyFound;
+
+    public StorageImpl() {
+        this.keys = new Object[ARRAY_MAX_SIZE];
+        this.values = new Object[ARRAY_MAX_SIZE];
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size < ARRAY_MAX_SIZE) {
+            keyFound = false;
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null ? key == null : keys[i].equals(key)) {
+                    values[i] = value;
+                    keyFound = true;
+                    break;
+                }
+            }
+            if (!keyFound) {
+                keys[size] = key;
+                values[size] = value;
+                size();
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
+                    return (V) values[i];
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return (V) values[i];
+                }
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size++;
+    }
+
+    public boolean equals(Object object) {
+        if (this == object) {
+            return true;
+        }
+        if (object == null || this.getClass() != object.getClass()) {
+            return false;
+        }
+        StorageImpl<K, V> storageImpl = (StorageImpl<K, V>) object;
+        return (keys == null
+                ? storageImpl.keys == null
+                : keys.equals(storageImpl.keys))
+                && (values == null
+                ? storageImpl.values == null
+                : values.equals(storageImpl.values));
+    }
+
+    public int hashCode() {
+        result = 17;
+        result = 31 * result + (keys != null ? keys.hashCode() : 0);
+        result = 31 * result + (values != null ? values.hashCode() : 0);
+        return result;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
vikkushnir, line 8: Better to use K[], V[] types instead of Object

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+                if (keys[i] == null ? key == null : keys[i].equals(key)) {
+                    values[i] = value;
+                    keyFound = true;
+                    break;
+                }
+            }
+            if (!keyFound) {
+                keys[size] = key;
+                values[size] = value;
+                size();
```
vikkushnir, line 33: I'm not sure, but better to use size++ instead of method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+                && (values == null
+                ? storageImpl.values == null
+                : values.equals(storageImpl.values));
+    }
+
+    public int hashCode() {
+        result = 17;
+        result = 31 * result + (keys != null ? keys.hashCode() : 0);
+        result = 31 * result + (values != null ? values.hashCode() : 0);
+        return result;
```
vikkushnir, line 81: this equals and hashCode methods are extra, code will be working without them

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+                }
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size++;
```
vikkushnir, line 58: use size without ++

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+        this.values = new Object[ARRAY_MAX_SIZE];
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size < ARRAY_MAX_SIZE) {
+            keyFound = false;
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null ? key == null : keys[i].equals(key)) {
```
vikkushnir, line 24: you can create additional methods some indexOfKey and use it for put and get it will simplify your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] == null) {
+                    return (V) values[i];
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return (V) values[i];
```
vikkushnir, line 49: you can use the methods above "indexOfKey" or you can simplify this, by combining key == null and key.equals(keys[i]) as in the end you return one value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,81 @@
...
+    private static final int ARRAY_MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+    private int result;
+    private boolean keyFound;
+
+    public StorageImpl() {
+        this.keys = new Object[ARRAY_MAX_SIZE];
+        this.values = new Object[ARRAY_MAX_SIZE];
```
vikkushnir, line 15: customize them to (K[]), (V[])

vikkushnir (changes requested):
I have proposed some changes to simplify your code, your code is good but a little difficult

src/main/java/core/basesyntax/impl/StorageImpl.java
azzilian: not sure - the thing is I got array this.keys = new Object[ARRAY_MAX_SIZE]; and I cannot do this.keys = new K[ARRAY_MAX_SIZE];

src/main/java/core/basesyntax/impl/StorageImpl.java
azzilian: Well I thought about that too - will change

src/main/java/core/basesyntax/impl/StorageImpl.java
azzilian: Yes you are right, works perfectly well.I dont remember why I need this thought... maybe old code

src/main/java/core/basesyntax/impl/StorageImpl.java
vikkushnir: I mean that you will use K[] keys, and V[] values, and then in the constructor you will parameterize this.keys = (K[]) new Object[ARRAY_MAX_SIZE];

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[ARRAY_MAX_SIZE];
+        this.values = (V[]) new Object[ARRAY_MAX_SIZE];
+        this.size = 0;
```
karma-o, line 14: primitive class fields have a default value, for int it is 0, so we can remove that
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
     @Override
     public void put(K key, V value) {
+        int index = findIndexKey(key);
+        if (index != -1) {
+            values[index] = value;
+        } else if (size < ARRAY_MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
karma-o, line 35: a small tweak for readability, regarding the max size check - would be perfect to check it and throw an exception if it is full, or you can ignore it for now, since it is not in the scope of this task
```suggestion
            return;
        }
        keys[size] = key;
        values[size] = value;
        size++;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[ARRAY_MAX_SIZE];
+        this.values = (V[]) new Object[ARRAY_MAX_SIZE];
```
karma-o, line 13: no need in `this` keyword since there are no other variables with similar names in scope of this method
```suggestion
        keys = (K[]) new Object[ARRAY_MAX_SIZE];
        values = (V[]) new Object[ARRAY_MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = findIndexKey(key);
+        return index != -1 ? (V) values[index] : null;
```
karma-o, line 41: I'd reverse the -1 check to not make it `!=` but make it `==`, which would be easier to read, also make `-1` a constant
```suggestion
        return index == -1 ? null : (V) values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
```
karma-o, line 6: would be nice to split static fields from non static with an empty line
```suggestion
    private static final int ARRAY_MAX_SIZE = 10;
    
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private K[] keys;
+    private V[] values;
```
karma-o, line 8: can be final, but not necessary, so I guess it is up to you

karma-o (changes requested):
GJ! Left some comments

src/main/java/core/basesyntax/impl/StorageImpl.java
azzilian: something like  try catch (ArrayIndexOutOfBoundsException e) ?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
+
+    private final K[] keys;
+    private final V[] values;
```
andrii-hoienko, line 9:
```suggestion
    private final V[] values;
    
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_MAX_SIZE = 10;
```
andrii-hoienko, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```

andrii-hoienko (changes requested):
Nice, minor comments left

src/main/java/core/basesyntax/impl/StorageImpl.java
karma-o: no, we do not need to catch anything, our requirements say that ma size of storage is 10, so to ensure that we won't have more than 10 elements and to not get the raw `ArrayIndexOutOfBoundsException` we should avoid the situation when we get it by checking if size is 10 and we attempt to put another key/value pair - then we throw our own 
 custom `RuntimeException` with a message about max storage size

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    private int size;
+    private int index;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
+    private int findIndexKey(K key) {
```
karma-o, line 20: private methods should be after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+            }
+        }
+        return KEY_NOT_FOUND;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (size > MAX_SIZE) {
+            throw new IllegalStateException("Storage is full");
+        }
```
karma-o, line 33: this should be after the check of the existing key because it will not increase a size, thus is a valid operation.

JFYI: in the previous attempt with try/catch, I have written another comment above on what we should do.
Also wanted to answer the question of why: this is because array index out of bounds is an unchecked/runtime exception which we should prevent, and not catch, unlike checked ones

karma-o (approved):
GJ! just some minor things left

PR #1444, opened 2024-02-26
jv-storage solution

vikkushnir:


kaiiiseeel (changes requested):
LGFM! Left a single comment

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final K[] keyArr;
+    private final V[] valueArr;
+    private int size;
+
+    public StorageImpl() {
+        keyArr = (K[]) new Object[MAX_SIZE];
+        valueArr = (V[]) new Object[MAX_SIZE];
+        size = size;
```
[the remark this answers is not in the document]
vikkushnir: Yes, you are right, I don't need this, I missed that we can use size just in class

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+
+    public StorageImpl() {
+        keyArr = (K[]) new Object[MAX_SIZE];
+        valueArr = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOfKey(key);
+        if (index != -1) {
```
andrii-hoienko, line 19: Make -1 constant

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+            valueArr[index] = value;
+        } else if (size < MAX_SIZE) {
+            keyArr[size] = key;
+            valueArr[size] = value;
+            size++;
```
andrii-hoienko, line 24:
```suggestion
        }
        if (size < MAX_SIZE) {
            keyArr[size] = key;
            valueArr[size] = value;
            size++;
            } else {
            throw new SomeException('Storage is full");
            }
```

andrii-hoienko (changes requested):
Nice, minor comments left

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        keyArr = (K[]) new Object[MAX_SIZE];
+        valueArr = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOfKey(key);
+        if (index != NOT_FOUND_INDEX) {
+            valueArr[index] = value;
+            size--;
```
karma-o, line 22: why do you decrement size?

UPD: I see now, what Andrii meant is using the `return`s instead of complications in the `if` `else`s
```suggestion
            return;
```
next time feel free to ask questions if you think that some decision or comment is strange or if you think you do not completely understand it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        }
+        if (size < MAX_SIZE) {
+            keyArr[size] = key;
+            valueArr[size] = value;
+            size++;
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
```
karma-o, line 30: once again we throw and do not use redundant `else`s
```suggestion
        if (size >= MAX_SIZE) {
            throw new RuntimeException("Storage is full");
        }
        keyArr[size] = key;
        valueArr[size] = value;
        size++;
```

karma-o (approved):
GJ, Please also check the comments that I have left

src/main/java/core/basesyntax/impl/StorageImpl.java
vikkushnir: Thank you, understand it, wanted you ask about this on QnA, but you already explain it

PR #1445, opened 2024-02-26
HomeWork

Yam0r:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keys[i])) {
+                    return i;
+                }
+            }
+        }
+        return -1; // Key not found
```
okuzan, line 66: remove if-else
merge two for loops into one
use the same condition as in Objects.equals
create a NOT_FOUND_INDEX for -1
remove comments everywhere

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = findIndex(key);
+        if (index != -1) {
+            // Key found, return the corresponding value
+            return (V) values[index];
+        } else {
```
okuzan, line 41: use ternary operator

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_SIZE];
+        this.values = new Object[MAX_SIZE];
+        this.size = 0;
```
okuzan, line 14: we don't need `this` here
0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+            values[index] = value;
+        } else {
+            // Key doesn't exist, add a new entry
+            if (size < MAX_SIZE) {
+                keys[size] = key;
+                values[size] = value;
+                size++;
+            } else {
+                throw new RuntimeException("Storage is full");
+            }
```
okuzan, line 31: throw exception early, unwrap the rest of your code - you won't need else after return/throw

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 8: use K,V type parameters

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
     public void put(K key, V value) {
+        int index = findIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size++] = value;
+        } else {
+            throw new RuntimeException("Storage is full");
+        }
```
olekskov, line 28: In general the less `else` conditions we have the more easy to read code is
```suggestion
        int index = findIndex(key);
        if (index != -1) {
            values[index] = value;
            return;
        } 
        if (size < MAX_SIZE) {
            throw new RuntimeException("Storage is full");
        } 
        keys[size] = key;
        values[size++] = value;
```

olekskov (approved):
Good job, approved.
Comment below is for further consideration

PR #1448, opened 2024-02-26
feat: first-try

rel1c-hub:
Passed the test & build success.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -26,7 +26,8 @@ public class StorageImpl<K, V> implements Storage<K, V> {
             keyArr[size] = key;
             size++;
         } else {
-            throw new RuntimeException("Can't write to array! Index out of borders!");
+            throw new RuntimeException("Can't write to array! Index out of borders!"
+                    + "Max size is: " + MAX_SIZE);
         }
     }
 


src/main/java/core/basesyntax/Storage.java
```diff
@@ -1,6 +1,7 @@
 package core.basesyntax;
 
 public interface Storage<K, V> {
+
```
karma-o, line 4: we do not need that
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private static final int NOT_FOUND_INDEX = -1;
+    private int size = 0;
```
karma-o, line 9: 0 is the default value for primitive int fields
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+    private final K[] keyArr;
+    private final V[] valArr;
+
+    public StorageImpl() {
+        this.keyArr = (K[]) new Object[MAX_SIZE];
+        this.valArr = (V[]) new Object[MAX_SIZE];
```
karma-o, line 15: redundant because there are no other variables with such names in scope of that method
```suggestion
        keyArr = (K[]) new Object[MAX_SIZE];
        valArr = (V[]) new Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
+        if (index != NOT_FOUND_INDEX) {
+            valArr[index] = value;
+            return;
+        }
+        if (index < MAX_SIZE) {
+            valArr[size] = value;
+            keyArr[size] = key;
+            size++;
+        } else {
+            throw new RuntimeException("Can't write to array!");
```
karma-o, line 30: add the reason why it is not possible so that the user can understand what is wrong and fix that

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,56 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public int getKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keyArr[i])) {
```
karma-o, line 50: `Objects` class is not allowed to be used in this task

karma-o (changes requested):
Minor comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        if (index != NOT_FOUND_INDEX) {
+            valArr[index] = value;
+            return;
+        }
+        if (index < MAX_SIZE) {
+            valArr[size] = value;
+            keyArr[size] = key;
+            size++;
+        } else {
+            throw new RuntimeException("Can't write to array! Index out of borders!");
```
karma-o, line 29: we can be more specific than that by adding a max size to the message

karma-o (approved):
🔥

PR #1450, opened 2024-02-28
HW

RetKinf:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private Object[] keyArray = new Object[MAX_CAPACITY];
+    private Object[] valueArray = new Object[MAX_CAPACITY];
+    private int size = 0;
```
olekskov, line 9: 1. Better to properly paramtrize arrays during initialization in order not to type-cast in methods
2. Better to initialize variable in constructor
3. 0 is default int value, no need to explicitly set it
4. It's very rarely makes sense to include type into variable name
```suggestion
    private K[] keys;
    private V[] values;
    private int size;
    
    public Storage() {
        keys = (K[]) new Object[MAX_CAPACITY];
        values = (K[]) new Object[MAX_CAPACITY];
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        return (V) valueArray[count];
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    public int counter(K key) {
```
olekskov, line 37: 1. Method names should contain verbs for better code readability
2. Please keep public only official class logic. Details of realization shouldn't be public
```suggestion
    private int getIndex(K key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (key.equals(keyArray[i])) {
+                    count = i;
+                }
+            }
+        }
+        return count;
```
olekskov, line 52:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return i;
            }
        }
        return -1;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int count = counter(key);
+        if (count == -1) {
+            return (V) null;
+        }
+        return (V) valueArray[count];
```
olekskov, line 29: 1. variable `count` should be `index`
2. ternary operator instead of `if` here is more short and rather simple

olekskov (changes requested):

Vasul-Pankiv (approved):

PR #1457, opened 2024-03-14
Task performance jv-storage

ruslanPan:


artemZhuravskyi (changes requested):
image
please fix tests and check style

PR #1459, opened 2024-03-20
MateAcademy/Task/jv-storage - Bartosz Wójcik solution

banderan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys;
```
nowickimj, line 7: Arrays for `keys` and `values` may be generics:
```java
K[] keys = (K[]) new Object [MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (keyCondition(key, keys[i])) {
```
nowickimj, line 20: 1. Please align to boolean method naming convention (e.g. `isEmpty`, `hasChild`, `equals` etc.). For comparing keys of the same type, `keysEqual()` could be fine.
2. Since `keys[]` is a private field - you don't need to pass it as an arg - keep the number of args as low as possible.

How about `hasKeyOnIndex(K key, int index)` ?

nowickimj (changes requested):
Change required.

nowickimj (approved):

PR #1463, opened 2024-03-24
[Java Core/Generics/Storage] solution v1

ShastkivRuslan:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return storageSize;
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < storageSize; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
Rommelua, line 41:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

Rommelua (approved):

PR #1464, opened 2024-03-24
Implemented key-value storage

YEvhenKorobka:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
+    private Object[] values = new Object[MAX_SIZE];
```
Rommelua, line 8: image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+    private int size = 0;
```
Rommelua, line 9: 0 is a default value
```suggestion
    private int size;
```

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there.

https://github.com/YEvhenKorobka/jv-storage/blob/storage-implementation/checklist.md

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                if (keys[i] == null) {
+                    values[i] = value;
+                    return;
+                }
+            }
+        } else {
+            for (int i = 0; i < size; i++) {
+                if (keys[i] != null && keys[i].equals(key)) {
+                    values[i] = value;
+                    return;
```
kerrrusha, line 29: look, actions on 21-22 and 28-29 rows are the same, so this it code duplication. 
please, remove outer if-else block and combine conditions on 20 and 27 lines so you'll have only loop and 1 `if` block inside

kerrrusha (changes requested):

Rommelua (approved):

PR #1465, opened 2024-03-25
solution of the task

zahurskyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (MAX_SIZE <= size) {
+            throw new IllegalStateException("Storage is full");
+        }
```
MAMentorFX, line 20: We have to be able to change the value which already exists in the Storage. So better to put this validation after the for-loop

MAMentorFX (approved):

PR #1471, opened 2024-04-06
completed the task

citruscaviar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private final Pair<K, V>[] arrayOfPairs;
```
MAMentorFX, line 7:
```suggestion
    private final Pair<K, V>[] pairs;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    private int elementsAdded;
```
MAMentorFX, line 8:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        this.arrayOfPairs = (Pair<K, V>[]) new Pair[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (elementsAdded >= MAX_CAPACITY) {
+            throw new RuntimeException("The maximum capacity is 10");
+        }
```
MAMentorFX, line 19: It's possible to use put() without changing elementsAdded, by just changing value of existing key, so we want this logic to work if elements == max capacity

MAMentorFX (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
citruscaviar: I used those names deliberately to help myself understand the code better

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     public void put(K key, V value) {
+        if (size >= MAX_CAPACITY) {
+            throw new RuntimeException("The maximum capacity is 10");
+        }
+        for (Pair<K, V> element : pairs) {
+            if (element != null && equals(element.getKey(), key)) {
+                element.setValue(value);
+                return;
+            }
+        }
```
MAMentorFX, line 25:
```suggestion
        for (Pair<K, V> element : pairs) {
            if (element != null && equals(element.getKey(), key)) {
                element.setValue(value);
                return;
            }
        }
        if (size >= MAX_CAPACITY) {
            throw new RuntimeException("The maximum capacity is 10");
        }
```

MAMentorFX (changes requested):

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private final Pair<K, V>[] pairs;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        this.pairs = (Pair<K, V>[]) new Pair[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
```
MAMentorFX, line 16: No, put has three ways to use. 1) change existing value 2) if size >= throw 3) in other cases add to storage.
for now you check if size >= max, after you add even if storage is full. Do these functionalities in the order that I have sent you

MAMentorFX (changes requested):

Rommelua (approved):

PR #1472, opened 2024-04-07
the Storage task done

Svitlankaa:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS = 10;
+    private Object[] keys;
+    private Object[] values;
```
okuzan, line 8: why don't you use your class type parameters K, V, that's exactly what they are for

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (key != null && key.equals(keys[i]))) {
+                return (V) values[i];
+            }
+        }
         return null;
```
okuzan, line 40: extract logic of iterating and comparing keys, returning index of the match - to a separate method
return -1 for not found

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key != null && keys[i] != null && key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+            if (key == null && keys[i] == null) {
+                values[i] = value;
+                return;
+            }
+        }
```
okuzan, line 27: should be one clause
look up condition for Objects.equals - Сtrl and click the method to jump to the source (you can't use utility directly in this task)

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (consistIndex(i, key)) {
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
+    private boolean consistIndex(int i, K key) {
```
kozhukhovsky, line 40: All public methods come first, then all private ones.

kozhukhovsky (approved):

PR #1473, opened 2024-04-09
Implemented the Storage of the key-value type that is parameterized w…

Alona-Martynova:
…ith two types of data


src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,44 @@
...
+public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBERS = 10;
+    private Object[] keys;
+    private Object[] values;
+    private int count;
+
+    public StorageImpl() {
+        keys = new Object[MAX_NUMBERS];
+        values = new Object[MAX_NUMBERS];
+        count = 0;
```
kozhukhovsky, line 12:
```suggestion
```

0 is default value for int

kozhukhovsky (approved):

PR #1474, opened 2024-04-09
resolve jv-storage task

YouraSoliar:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_SIZE = 10;
+    private K[] keyStorage = (K[]) new Object[STORAGE_SIZE];
+    private V[] valueStorage = (V[]) new Object[STORAGE_SIZE];
```
okuzan, line 8: rename to just keys,values, init in the constructor (checklist issue)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     public V get(K key) {
+        for (int i = 0; i < STORAGE_SIZE; i++) {
+            if (keyStorage[i] == null && key == null && valueStorage[i] != null) {
+                return valueStorage[i];
+            }
+
+            if (keyStorage[i] != null) {
+                if (keyStorage[i].equals(key)) {
+                    return valueStorage[i];
+                }
```
okuzan, line 38: you need only one if in both get and put, with correct condition (look up Objects.equals() but remember you can't use it directly)
since this logic is repeated, consider extracting it to a private method (it will return index of a match, -1 if not found)

okuzan (changes requested):

src/main/java/core/basesyntax/KV.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class KV<K, V> {
```
Rommelua, line 3:
```suggestion
public class Pair<K, V> {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,60 @@
...
+    private KV<K, V>[] storage;
+
+    public StorageImpl() {
+        storage = (KV<K, V>[]) new KV<?, ?>[STORAGE_SIZE];
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = this.findIndex(key);
```
Rommelua, line 16:
```suggestion
        int index = findIndex(key);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,60 @@
...
+
+    @Override
+    public int size() {
+        int counter = 0;
+        for (int i = 0; i < STORAGE_SIZE; i++) {
+            if (storage[i] != null) {
+                counter++;
+            }
+        }
+        return counter;
```
Rommelua, line 58: Add `size `field and update it in the `put `method
```suggestion
        return size;
```

Rommelua (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,55 @@
...
+                ++size;
+                return;
+            }
+        }
+    }
+
+    private int findIndex(K key) {
+        for (int i = 0; i < STORAGE_SIZE; ++i) {
+            if (storage[i] != null) {
+                if ((storage[i].getKey() == null && key == null)
```
MAMentorFX, line 42: extract storage[i].getKey() to variable, three times used

MAMentorFX (approved):

PR #1475, opened 2024-04-09
implementation methods with Storage in StorageImpl class

SviatoslavTataryn:


okuzan (changes requested):
Please review checklist of common mistakes

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
+    private int size = 0;
```
kerrrusha, line 10: int variables are zero by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
```
kerrrusha, line 9: 1. try to use generic types
2. https://github.com/mate-academy/jv-storage/blob/master/checklist.md#all-non-constant-fields-should-be-initialized-in-the-constructor
```suggestion
    private final K[] keys = new (K[]) Object[MAX_SIZE];
    private final V[] values = new (V[]) Object[MAX_SIZE];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
     @Override
     public void put(K key, V value) {
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            values[index] = value;
+        } else if (size < MAX_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
```
kerrrusha, line 21: try to rewrite this code so it will inform user in case of size overflowing by throwing an exception that element can not be added because of size overflow

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public V get(K key) {
-        return null;
+        int index = findKeyIndex(key);
+        return index != -1 ? (V) values[index] : null;
```
kerrrusha, line 28: make -1 as a constant

kerrrusha (changes requested):

Vasul-Pankiv (approved):

PR #1480, opened 2024-04-15
Implemented logic to StorageImpl class

vovms:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int STORAGE_MAX_CAPACITY = 10;
+    private final Object[] keys;
```
Elena-Bruyako, line 8: It's better to use generics
```suggestion
    private final K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+    private final Object[] values;
```
```suggestion
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
     @Override
     public V get(K key) {
-        return null;
+        int index = getKeyIndex(key);
+        return index == -1
+                ? null
+                : (V) values[index];
+    }
+
+    private int getKeyIndex(K key) {
```
Elena-Bruyako, line 38: private methods should be placed after all public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+
+    public StorageImpl() {
+        keys = new Object[STORAGE_MAX_CAPACITY];
+        values = new Object[STORAGE_MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
```
Elena-Bruyako, line 20: Please add check when size == STORAGE_MAX_CAPACITY

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,51 @@
...
+        } else {
+            values[index] = value;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int index = getKeyIndex(key);
+        return index == -1
```
Elena-Bruyako, line 33: You can inline this operation

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
vovms: I thought about it, but such code doesn't compile. Error description: java: generic array creation

Screenshot 2024-04-16 092346

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+        keys = (K[]) new Object[STORAGE_MAX_CAPACITY];
+        values = (V[]) new Object[STORAGE_MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+        if (index == -1) {
+            if (size == STORAGE_MAX_CAPACITY) {
```
Elena-Bruyako, line 22: It should be the first check in method. 
Also, you can use return; if size is full

Elena-Bruyako (approved):
Approve, but check the comment

PR #1481, opened 2024-04-15
solution

trokhim03:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUM = 10;
```
Elena-Bruyako, line 6:
```suggestion
    private static final int MAX_SIZE = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private int size;
+    private K[] keys;
```
```suggestion
    private final K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private V[] values;
```
```suggestion
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+        values = (V[]) new Object[MAX_NUM];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (MAX_NUM < size) {
+            throw new RuntimeException("Storage is full!");
+        }
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 22: Please, use separate method to avoid duplicating

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            if ((key == null && keys[i] == null) || (key != null && key.equals(keys[i]))) {
```
Elena-Bruyako, line 23:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+            }
+        }
+        values[size] = value;
+        keys[size] = key;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 35: You need to get index of key and then 
`return index == -1 ? null : values[index];`

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
+    }
+
+    public void ensureCapacity() {
```
Elena-Bruyako, line 17:
```suggestion
    private void ensureCapacity() {
```
this method should be placed after all public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        if (MAX_SIZE < size) {
+            throw new RuntimeException("Storage is full!");
+        }
+    }
+
+    public int getIndex(K key, K[] keys) {
```
Elena-Bruyako, line 23:
```suggestion
    private int getIndex(K key, K[] keys) {
```

Elena-Bruyako (approved):
Approve, but pay attention to the private method and where they should be placed

PR #1482, opened 2024-04-16
add hw jv-storage

oleksandr-saulko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private Object[] keys;
```
Elena-Bruyako, line 7:
```suggestion
    private K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+    private Object[] values;
```
```suggestion
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+    public StorageImpl() {
+        keys = new Object[MAX_CAPACITY];
+        values = new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 20: To avoid several for iteration you can create private method to get index. Then check 
if index != -1 -> 
`values[index] = value; `

if index == -1; -> 
```
keys[size] = key;
values[size] = value;
size++;
```

```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+            values[size] = value;
+            size++;
+        } else {
+            System.out.println("Storage is full, cannot add more elements.");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 46:
```suggestion
       int index = indexOf(key);
       return index == -1 ? null : values[index];
```

Also, it's better to create constant for -1 value, like NO_INDEX for example

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done. Thx for advice, it's cool

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
+    }
+
+    private int indexOf(K key) {
```
okuzan, line 18: private methods (and class members in general) should go after public ones

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private static final int MAX_CAPACITY = 10;
+    private static final int NO_INDEX = -1;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
okuzan, line 15: 0 is default for int, this doesn't have any effect

okuzan (changes requested):
minor issues

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done

src/main/java/core/basesyntax/impl/StorageImpl.java
oleksandr-saulko: Done

Elena-Bruyako (approved):

PR #1484, opened 2024-04-17
First solution, basic tasks solved

noiseintheattic:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private final int maxStorage = 10;
```
nowickimj, line 7: This should be a constant (`private static final` with UPPERCASE name)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    private int currentSize = 0;
```
nowickimj, line 8: just `size`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    private int currentIndex = 0;
```
nowickimj, line 9: You don't need this value - it's just a duplicate of `size`

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        } else if (!isKeyExist(key)) {
+            valueArray[currentIndex] = value;
+            keyArray[currentIndex] = key;
+            currentSize++;
+            currentIndex++;
+        } else if (isKeyExist(key)) {
+            valueArray[currentIndex] = value;
+            keyArray[currentIndex] = key;
+            currentIndex++;
+        }
```
nowickimj, line 38: This seems a bit complicated for few reasons:
1. First of all, you repeat yourself:
```java
valueArray[currentIndex] = value;
keyArray[currentIndex] = key;
currentIndex++;
```
The code above executes despite the result of `isKeyExist()` - so why putting it in if/else?

2. You call `isKeyExist()` two times instead of getting value once and use it in if/else
3. `isKeyExist` is not very much of an English expression, `keyExists` would be better.

4. Got proposition for solution:
a) Use loop to iterate over keys
b) Use Objects.equals() to perform a null-safe comparison of keys
c) If key match, update/return value.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
     @Override
-    public int size() {
-        return -1;
+    public V get(K key) {
+        if (key == null) {
+            return valueForKeyNullCase;
+        }
+        V output = null;
+        for (int i = 0; i < keyArray.length; i++) {
+            if (key.equals(keyArray[i])) {
```
nowickimj, line 57: Same here - you don't need to store value for nullable key separately - just perform the proper comparison here. `Objects.equals()` will be helpful for nullsafe comparison.

nowickimj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+            keyArray[size] = key;
+            valueArray[size] = value;
+            size++;
+        }
+        for (int i = 0; i < keyArray.length; i++) {
+            if (Objects.equals(key, keyArray[i])) {
+                valueArray[i] = value;
+            }
+        }
+    }
```
nowickimj, line 26: 1. You don't need to use 2 loops - this should've been handled in single iteration.
2. The `size` indicates how many spaces are taken, so the last element you should iterate is `keyArray[size - 1]`
3. Think of the proper key check - you should cover both null and equality check in single `if` condition.
4. Think of what should happen when max size is reached - how about throwing an exception?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_STORAGE_VALUE = 10;
+    private int size = 0;
+
+    private final V[] valueArray = (V[]) new Object [MAX_STORAGE_VALUE];
```
nowickimj, line 11: `values`, try to avoid redundancy when naming stuff

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,52 @@
...
+    private final K[] keyArray = (K[]) new Object [MAX_STORAGE_VALUE];
```
nowickimj, line 12: `keys`, try to avoid redundancy when naming stuff

nowickimj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done. It took me a while, but I think  it's the right thing now.

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private static final int MAX_SIZE = 10;
+    private int size = 0;
+
+    private final V[] storageContents = (V[]) new Object [MAX_SIZE];
+    private final K[] storageNames = (K[]) new Object [MAX_SIZE];
```
nowickimj, line 12: As mentioned here: https://github.com/mate-academy/jv-storage/pull/1484#discussion_r1571959526
Keep it simple and consistent - here it's "contents" but inside methods those are "values".
As mentioned before, `keys` and `values` are enough.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+
     @Override
     public void put(K key, V value) {
+
```
nowickimj, line 16: Unnecessary new line.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, storageNames[i])) {
+                storageContents[i] = value;
+                size--;
```
nowickimj, line 20: why are you reducing size on putting new object?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+        try {
+            storageNames[size] = key;
+            storageContents[size] = value;
+            size++;
+        } catch (Exception e) {
+            if (size >= MAX_SIZE) {
+                throw new StorageSizeException("Can't put value to storage, "
+                        + "because storage is full");
+            }
+        }
```
nowickimj, line 33: Why using try/catch on code that won't throw any exception.

I think you should rethink your strategy:
You've just done iterating over the storage for indexes <0,size)
If you managed to insert object on existing key during iteration, you can exit method immediately from the loop with `return`. Size is not changes since you're overwriting the existing value.
But what, if you won't find expected object with loop?
My guess is: you must check whether the max size is reached. If it's exceeded, you should throw an exception. And if not? Just put the object as a next element and increase the size.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,53 @@
...
+
     }
 
     @Override
     public V get(K key) {
-        return null;
+        V output = null;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, storageNames[i])) {
+                output = storageContents[i];
```
nowickimj, line 42: No need for `output` variable - return object immediately here.

nowickimj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done. I'm sorry, I didn't understand Your comment correctly.

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done,

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: I missed the possibilty to exit loop by "return" statement.

src/main/java/core/basesyntax/impl/StorageImpl.java
noiseintheattic: ok, done

nowickimj (approved):

PR #1485, opened 2024-04-19
done generic class implementation

BelitskyiAlexandr:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[STORAGE_CAPACITY];
+        this. values = new Object[STORAGE_CAPACITY];
+        this.size = 0;
```
MentorMA, line 14:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            return getNullKey();
+        }
+        int index = indexOf(key);
+        if (index != -1) {
+            return (V) values[index];
```
MentorMA, line 69: Better cast in constructor once

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return this.size;
```
MentorMA, line 85:
```suggestion
        return size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[STORAGE_CAPACITY];
+        this. values = new Object[STORAGE_CAPACITY];
```
MentorMA, line 13:
```suggestion
        keys = new Object[STORAGE_CAPACITY];
        values = new Object[STORAGE_CAPACITY];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
+                return;
+            }
+        }
+        if (size < STORAGE_CAPACITY) {
+            keys[size] = null;
+            values[size] = value;
+            size++;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Storage is full");
+        }
```
MentorMA, line 59: Private methods goes after public

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
+        this. values = new Object[STORAGE_CAPACITY];
+        this.size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            putNullKey(value);
+            return;
+        }
```
MentorMA, line 22: Not sure that you need a separate method for null key, try to combine conditions using one block if

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
+        if (index != -1) {
+            values[index] = value;
+        } else {
+            if (size < STORAGE_CAPACITY) {
+                this.keys[size] = key;
+                this.values[size] = value;
+                size++;
+            } else {
+                throw new ArrayIndexOutOfBoundsException("Storage is full");
+            }
```
MentorMA, line 33: Use check storage capacity for full once

MentorMA (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: what's wrong with the size check? I don't understand your suggestion

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: why don't we use 'this'?

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: why do we need to find the index of the function parameter in the constructor?

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: why don't we initialize the size variable here?

okuzan: 0 is default for int, this line doesn't effect the execution flow :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
okuzan, line 8: The point of having class type parameters in this class is to avoid exactly this ;) Use K,V here instead of generic Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,85 @@
...
     }
 
     @Override
     public int size() {
+        return this.size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
okuzan, line 56: Please remove all null specific handling from your code, instead, you should use null safe equality check condition - a == b OR a IS NOT NULL AND a.equals(b)

okuzan (changes requested):
replace null handling with null save equal conditions

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: image
But how to fix this problem?

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: My apologies, I don't understand your comment at all. What do you want to say?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
Elena-Bruyako, line 55:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_CAPACITY];
+        values = (V[]) new Object[STORAGE_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
```
Elena-Bruyako, line 18: You can use this logic instead

```
if (size == STORAGE_CAPACITY) {
     return;
}
int index = indexOf(key);
if (index != -1) {
      values[index] = value;
} else if (index == -1) {
      keys[size] = key;
      values[size] = value;
      size++;
}
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+                size++;
+            } else {
+                throw new ArrayIndexOutOfBoundsException("Storage is full");
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
```
Elena-Bruyako, line 38:
```suggestion
        int index = indexOf(key);
        return index == -1 ? null : values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    private void putNullKey(V value) {
```
Elena-Bruyako, line 62: If you implement the logic above, you don't need this method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,84 @@
...
+        if (size < STORAGE_CAPACITY) {
+            keys[size] = null;
+            values[size] = value;
+            size++;
+        } else {
+            throw new ArrayIndexOutOfBoundsException("Storage is full");
+        }
+    }
+
+    private V getNullKey() {
```
Elena-Bruyako, line 78: If you implement the logic above, you don't need this method

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
BelitskyiAlexandr: `} else if (index == -1) {` - always true, isn`t it?

BelitskyiAlexandr:
> image But how to fix this problem? image Like this?

could you explain how to create a generic array? Because I'm not sure, I should do it like [that](https://www.baeldung.com/java-generic-array#creating-a-generic-array)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = indexOf(key);
+        if (index != -1) {
```
Elena-Bruyako, line 34: Minor comment

```suggestion
        return index == -1 ? null : values[index];
```

Let's create constant for -1 value

Elena-Bruyako (approved):
Approve, but left minor comment

src/main/java/core/basesyntax/impl/StorageImpl.java
okuzan: yes, like this

PR #1487, opened 2024-04-22
Solution

illiariaboy94:
My solution

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,47 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                return;
+            }
+        }
+        values[size] = value;
+        keys[size] = key;
+        size++;
+
+        if (MAX_SIZE <= size) {
+            throw new IllegalStateException("Storage is full");
+        }
+
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keys[i] == null)
+                    || key != null && key.equals(keys[i])) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 18: Let's replaced this code to separate method to avoid duplication
```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+
+        if (MAX_SIZE <= size) {
+            throw new IllegalStateException("Storage is full");
+        }
+
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 36: use indexOf method

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
```
Elena-Bruyako, line 7:
```suggestion
    private final K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private V[] values;
```
```suggestion
    private final V[] values;
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
illiariaboy94: ok

src/main/java/core/basesyntax/impl/StorageImpl.java
illiariaboy94: Thanks

Rommelua (approved):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
+    private int indexOf(K key) {
```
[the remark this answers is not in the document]
illiariaboy94: Ok

PR #1488, opened 2024-04-22
initial solution

MarcusPy:
Why is using Objects class (specifically its .equals method) considered a mistake according to the Common mistakes section? It would have allowed me to simplify this condition - keys[i] == key || (keys[i] != null && keys[i].equals(key)), if it weren't prohibited

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,58 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
+    private static final int MAX_SIZE = 10;
+    private final Object[] keys = new Object[MAX_SIZE];
+    private final Object[] values = new Object[MAX_SIZE];
+
+    private static Object[] removeNull(Object[] arr1, Object[] arr2) {
+        int newSize = 0;
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (arr1[i] != null || arr2[i] != null) {
+                arr1[newSize++] = arr1[i];
+            }
+        }
+
+        return Arrays.copyOf(arr1, newSize);
     }
 
-    @Override
-    public V get(K key) {
+    private Object indexOf(Object[] keys, Object key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                return i;
+            }
+        }
         return null;
     }
 
     @Override
+    public void put(K key, V value) {
+        for (int i = 0; i < MAX_SIZE; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
+                keys[i] = key;
+                values[i] = value;
+                break;
+            }
+        }
+    }
+
+    @Override
+    public V get(K key) {
+        Object[] trimmedKeys = removeNull(keys, values);
+        Object[] trimmedValues = removeNull(values, keys);
+        Object index = indexOf(trimmedKeys, key);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,58 @@
...
+    }
+
+    private Object indexOf(Object[] keys, Object key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
+                return i;
+            }
+        }
+        return null;
+    }
```
nowickimj, line 29: public methods should be listed first

nowickimj (changes requested):
Think of different approach to implement this structure.

MarcusPy:
After looking at my code today, I realised some of it was indeed redundant for this task. Hope this sleeker approach is better!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getKeyIdx(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == key || (keys[i] != null && keys[i].equals(key))) {
```
nowickimj, line 41: `Objects.equals()` handles null-safe comparison for you.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
     @Override
     public void put(K key, V value) {
+        int idx = getKeyIdx(key);
+        if (idx == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        } else {
+            values[idx] = value;
+        }
```
nowickimj, line 25: I see a problem with this solution:
1. Your size has reached `10`.
2. You're trying to insert a key that is not yet present in the storage.
3. `getKeyIdx()` returns `-1`.
4. You're overwriting the last object and set size to `11`.

What will happen if you try to add another object?

nowickimj (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        int idx = getKeyIdx(key);
+
+        if (size >= MAX_SIZE) {
+            throw new RuntimeException("Storage is full");
+        }
```
nowickimj, line 22: I see a problem here - you're throwing exception on reaching max size BEFORE checking, if key is present (`idx` different than `-1`). In such case, you might want to overwrite the existing value with new one.

src/main/java/core/basesyntax/impl/StorageImpl.java
nowickimj: Good call you saw that using `Objects` is forbidden here - didn't see that.
I guess it's due to make you understand of what's happening below instead of relying on util classes. Good work here!

nowickimj (changes requested):
One small change required.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        if (idx == -1) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+            if (size >= MAX_SIZE) {
+                throw new RuntimeException("Storage is full");
+            }
+        } else {
+            values[idx] = value;
+        }
```
nowickimj, line 28: 1. You should check the size limit BEFORE performing the operation. You shouldn't be able to add 11th element when max size is 10.
2. If you have a huge if body, try to come up with a reversed condition and use `return` in order to exit the method quickly:
```java
if(idx != -1) {
  values[idx] = value;
  return;
}
//rest of the code
```

nowickimj (changes requested):
Still one issue to resolve

nowickimj (approved):

PR #1494, opened 2024-04-26
create StorageImpl(),put,putNullKey,get and getNullKey() methods

AnMishkin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,76 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        if (key == null) {
+            putNullKey(value);
+            return;
+        }
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
Rommelua, line 24:
```suggestion
            if (key.equals(keys[i])) {
```

AnMishkin: why i do it? key may be null.....

Rommelua: If **key** is null we return on line 20
image

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,75 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (key == null) {
+            return getNullKey();
+        }
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key)) {
```
Rommelua, line 57:
```suggestion
            if (key.equals(keys[i])) {
```

Rommelua (changes requested):

kerrrusha (approved):

PR #1498, opened 2024-05-04
Implemented the Storage

svitlanagalicka:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < sizeOfStorage; i++) {
+            if ((keys[i] != null && keys[i].equals(key)) || keys[i] == key) {
+                return values[i];
+            }
+        }
```
okuzan, line 32: you have duplicated logic in both get and put methods, let's extract it to a new private method, that would return the index of a match, or -1 if not found

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private K[] keys = (K[])new Object[MAX_ELEMENTS_NUMBER];
+    private V[] values = (V[]) new Object[MAX_ELEMENTS_NUMBER];
+    private int sizeOfStorage;
```
```suggestion
    private int size;
```
for collections/storage type, size is understandable enough

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,38 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_NUMBER = 10;
+    private K[] keys = (K[])new Object[MAX_ELEMENTS_NUMBER];
+    private V[] values = (V[]) new Object[MAX_ELEMENTS_NUMBER];
```
okuzan, line 8: checklist issue, should be initialized in the constructor, and don't forget to hit Ctrl Alt L for correct formatting

okuzan (changes requested):
[Checklist](https://github.com/mate-academy/jv-storage/blob/master/checklist.md)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     @Override
     public V get(K key) {
+        int number = getKey(key);
+        if (number != -1) {
+            return values[number];
+        }
         return null;
     }
 
+    public int getKey(K key) {
```
sarakhmen, line 39: this method should be private. Also the name of this method isn't appropriate enough. If the method return an index of the key then the method should be name like 'getIndex'

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ELEMENTS_NUMBER];
+        values = (V[]) new Object[MAX_ELEMENTS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        int number = getKey(key);
```
sarakhmen, line 18:
```suggestion
        int index = getKey(key);
```

sarakhmen (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
Elena-Bruyako: Not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+        int index = getKey(key);
+        if (index != -1) {
+            return values[index];
+        }
         return null;
     }
 
+    public int getKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] != null && keys[i].equals(key)) || keys[i] == key) {
```
Elena-Bruyako, line 41:
```suggestion
            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
         return null;
     }
 
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] != null && keys[i].equals(key) || keys[i] == key) {
+                return i;
+            }
+        }
+        return -1;
```
okuzan, line 45: wouldn't hurt to make it a constant too (you use it two times, so if you decide to change it (but this is unlikely), you will only need to change one piece of code, not two)

okuzan (approved):
Good job

PR #1503, opened 2024-05-20
added solution of task

apanyutin:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_COLLECTION_SIZE = 10;
+    private Object[] keys = new Object[MAX_COLLECTION_SIZE];
+    private Object[] values = new Object[MAX_COLLECTION_SIZE];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key != null ? key.equals(keys[i]) : key == keys[i]) {
```
Vasul-Pankiv, line 13:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys(i)) {
```
As for me better to use such a check, it is easier to read, but yours is good too, so it's up to you which to choose

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,43 @@
...
     public int size() {
-        return -1;
+        int count = 0;
+
+        for (int i = 0; i < values.length; i++) {
+            if (values[i] != null) {
+                count++;
+            }
+        }
+        return count;
```
Vasul-Pankiv, line 43: what if you put null value into storage? Just crate an additional field "size" and increment it every time you add new item even if value is null

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
apanyutin: I decided to fix it another way, just added checking size for keys and values.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key != null ? key.equals(keys[i]) : keys[i] == null) {
+                return (V) values[i];
```
MentorMA, line 28: make cast in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        for (int i = 0; i < values.length; i++) {
+            if (values[i] != null) {
+                valuesAmount++;
+            }
+            if (keys[i] != null) {
+                keysAmount++;
+            }
+        }
+        return keysAmount >= valuesAmount ? keysAmount : valuesAmount;
     }
```
MentorMA, line 48:
```suggestion
return size
```

MentorMA: add size field for counting

MentorMA (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private Object[] keys = new Object[MAX_COLLECTION_SIZE];
+    private Object[] values = new Object[MAX_COLLECTION_SIZE];
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                values[i] = value;
+                break;
+            } else if (keys[i] == null && values[i] == null) {
```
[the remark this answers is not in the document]
apanyutin: don't agree, can be the situation when key=null, but value don't equal null.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private V[] values;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_COLLECTION_SIZE];
+        values = (V[]) new Object[MAX_COLLECTION_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keys.length; i++) {
```
Elena-Bruyako, line 18: It's better to create separate method to avoid duplication of code
```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

Elena-Bruyako: So your implementation will look like:
```
int index = indexOf(key);
if (index != -1) {
     values[index] = value;
} else if (index == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
}
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                values[i] = value;
+                size++;
+                break;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
Elena-Bruyako, line 36:
```suggestion
        int index = indexOf(key);
        return index == -1 ? null : values[index];
```

Elena-Bruyako (changes requested):

Vasul-Pankiv (approved):

PR #1508, opened 2024-05-25
commit storage

nplastun:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int currentSize;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        currentSize = 0;
```
```suggestion
```
0 is default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        } else {
+            throw new IllegalStateException("Storage is full");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentSize; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
+                return values[i];
```
okuzan, line 39: this logic is duplicated in `put` too, try to extract to a private helper method and use it in two places

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int currentSize;
```
```suggestion
    private int size;
```

okuzan (changes requested):
Nice code!

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
MentorMateAcademy, line 46:
```suggestion
            if (keys[i] ==key || keys[i] != null && keys[i].equals(key)) {
```

MentorMateAcademy: try to simplify

MentorMateAcademy (changes requested):

MentorMateAcademy (approved):

PR #1509, opened 2024-05-29
Implement class StorageImpl

YuraYurchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_INDEX = 10;
+    private int size;
+    private final Object[] key;
+    private final Object[] value;
+
+    public StorageImpl() {
+        key = new Object[MAX_ARRAY_INDEX];
+        value = new Object[MAX_ARRAY_INDEX];
+        this.size = 0;
```
Elena-Bruyako, line 14: int type = 0 by default, redundant initialization

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,44 @@
...
+            }
+        }
+        this.key[size] = key;
+        this.value[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 33: Create separate method to avoid duplication and use it here and in put method
```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private final V[] value;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ARRAY_INDEX];
+        value = (V[]) new Object[MAX_ARRAY_INDEX];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (indexOf(key) != -1) {
```
Elena-Bruyako, line 18: It's better to avoid calling method indexOf()twice
```suggestion
        int index = indexOf(key);
        if (index != -1) {
            values[index] = value;
        } else if (index == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            return;
+        }
+        this.keys[size] = key;
+        this.value[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        if (indexOf(key) != -1) {
```
Elena-Bruyako, line 29:
```suggestion
        int index = indexOf(key);
        return index == -1 ? null : values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_INDEX = 10;
+    private int size;
+    private final K[] keys;
+    private final V[] value;
```
```suggestion
    private final V[] values;
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ARRAY_INDEX];
+        values = (V[]) new Object[MAX_ARRAY_INDEX];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = indexOf(key);
+        if (index != -1) {
+            this.values[indexOf(key)] = value;
```
```suggestion
            values[index] = value;
```

Elena-Bruyako (approved):
Be attentive to the comments

PR #1510, opened 2024-05-30
make up the constructor initializing new Objects for key and value an…

odeineka:
…d its current size.

FindIndex compare input data with storage data to replace or add data to Storage and calculate its size

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,54 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int INITIAL_CAPACITY = 10;
+    private final K[] keys;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    private static final int INITIAL_CAPACITY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int currentSize;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        this.keys = (K[]) new Object[INITIAL_CAPACITY];
+        this.values = (V[]) new Object[INITIAL_CAPACITY];
+        this.currentSize = 0;
```
sarakhmen, line 15: currentSize is 0 by default

sarakhmen (changes requested):
Let's slightly improve your solution :)

Elena-Bruyako (approved):

PR #1511, opened 2024-06-05
Homework

Experywow:
Fields:
STORAGE_MAX_CAPACITY: A constant defining the maximum capacity of the storage (10 elements).
keys: An array for storing the keys.
values: An array for storing the values corresponding to the keys.
size: The current size of the storage.

Constructor:
Initializes the keys and values arrays with the specified capacity and sets the initial size (size) to 0.

Method put(K key, V value):
Adds a new key-value pair to the storage or replaces the value if the key already exists. First, it checks if the key already exists in the storage (considering the possible null value for the key). If the key is found, the value is updated.
If the key is not found, a new key-value pair is added if the storage size is less than the maximum capacity. If the storage is full, a RuntimeException is thrown.

Method get(K key):
Returns the value corresponding to the given key.
Iterates through the keys array to find the key (considering the possible null value for the key). If the key is found, the corresponding value from the values array is returned. If the key is not found, null is returned.

Method size():
Returns the current size of the storage (number of elements in the storage).

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,47 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_MAX_CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[STORAGE_MAX_CAPACITY];
+        values = new Object[STORAGE_MAX_CAPACITY];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (size < STORAGE_MAX_CAPACITY) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+
+        } else {
+            throw new RuntimeException("Storage is overfilled, maximum capacity is 10");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
+                return (V) values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_MAX_CAPACITY = 10;
+    private final Object[] keys;
```
Elena-Bruyako, line 7: Let's use generic type here
```suggestion
    private final K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private final Object[] values;
```
```suggestion
    private final V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = new Object[STORAGE_MAX_CAPACITY];
+        values = new Object[STORAGE_MAX_CAPACITY];
+        size = 0;
```
Elena-Bruyako, line 14: redundant, int = 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 19: Let's avoid duplication of code
Create indexOf method
```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
```suggestion
        int index = indexOf(key);
        if (index != -1) {
            values[index] = value;
        } else if (index == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            size++;
+
+        } else {
+            throw new RuntimeException("Storage is overfilled, maximum capacity is 10");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < keys.length; i++) {
```
Elena-Bruyako, line 37:
```suggestion
        int index = indexOf(key);
        return index == -1 ? null : values[index];
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,45 @@
...
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_MAX_CAPACITY];
+        values = (V[]) new Object[STORAGE_MAX_CAPACITY];
+    }
+
+    private int indexOf(K key) {
```
Elena-Bruyako, line 16: private method should be placed after public methods

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1512, opened 2024-06-10
Was implemented StorageImpl.java.Override methods - put, get and size…

bilberryyy:
…. Added static variable MAX_SIZE and ArrayLists


sarakhmen (approved):
Good job!

PR #1513, opened 2024-06-11
Implement storage in StorageImpl

smagles:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+                }
+            }
+        }
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                continue;
+            }
+            if (keys[i].equals(key)) {
+                return i;
+            }
```
MateuszRostek, line 63:
```suggestion
            if (keys[i] != null && keys[i].equals(key)) {
                return i;
            }
```

MateuszRostek (approved):
Good Job! See minor advice.

PR #1514, opened 2024-06-13
Implemented a key-value store parameterized by two data types.

Jacalius:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private static final int STORAGE_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+        size = 0;
```
sarakhmen, line 15: size is 0 by default

sarakhmen (changes requested):
let's slightly improve your solution :)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[STORAGE_SIZE];
+        values = (V[]) new Object[STORAGE_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 19: To avoid duplication let's create separate method

```
private int indexOf(K key) {
                return i;
            }
        }
        return -1;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+        if (size < STORAGE_SIZE) {
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 35:
```suggestion
        int index = indexOf(key);
        return index == -1 ? null : values[index];
```

Elena-Bruyako (changes requested):

sarakhmen (approved):
good job!

PR #1515, opened 2024-06-14
jv-storage

alwayswannajava:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean isReplaced = false;
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 22: please revisit common mistakes checklist and we will be able to continue the review 🤝

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,50 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
```
okuzan, line 7: don't leave empty lines after method/class definition

okuzan (changes requested):
Good job overall. Replace objects.equals with your own logic, extract common logic from put and get method to the private method which returns index of a key if it's found or -1 otherwise

Vasul-Pankiv (approved):

PR #1521, opened 2024-06-22
create class StoregeImpl that can put and get date from the array by …

ArtemHurievskyi:
…the key
easy task for 3 days, nice job Artem, nice job

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -1,19 +1,74 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 5;
+    private int size;
+    private K[] keyArray;
+    private V[] valueArray;
+
+    public StorageImpl() {
+        valueArray = (V[])new Object[ARRAY_LENGTH];
+        keyArray = (K[]) new Object[ARRAY_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (key == null && keyArray[i] == null && valueArray[i] == null) {
+                valueArray[i] = value;
+                size++;
+                return;
+            } else if ((key == null && keyArray[i] == null && valueArray[i] != null)
+                    || (key != null && key.equals(keyArray[i]))) {
+                valueArray[i] = value;
+                return;
+            }
+        }
+        if (size < ARRAY_LENGTH) {
+            keyArray[size] = key;
+            valueArray[size] = value;
+            size++;
+        } else {
+            throw new IllegalStateException("Storage is full");
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keyArray[i] == null) || (key != null && key.equals(keyArray[i]))) {
+                return valueArray[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int ARRAY_LENGTH = 5;
```
sarakhmen, line 6:
```suggestion
    private static final int ARRAY_LENGTH = 10;
```
please follow task description

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+    private int size;
+    private K[] keyArray;
+    private V[] valueArray;
```
sarakhmen, line 9:
```suggestion
    private K[] keys;
    private V[] values;
```
let's not use such a general names as array/list/queue/map etc in our namings

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        keyArray = (K[]) new Object[ARRAY_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < keyArray.length; i++) {
+            if (isNull(key, i)) {
+                valueArray[i] = value;
+                size++;
+                return;
```
sarakhmen, line 22: why did you need this null check here and separate logic for this case? what if the second element is null but 3-rd one isn't? the size will be incremented and the logic is weird

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,58 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((key == null && keyArray[i] == null) || (key != null && key.equals(keyArray[i]))) {
+                return valueArray[i];
+            }
+        }
```
sarakhmen, line 43: I'd recommend implementing additional private method getIndex/indexOf (whatever you want to name it) which will return the index of the provided key (or -1 in case key is absent in the keys array)

sarakhmen (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private K[] keys;
+    private V[] values;
+
+    public StorageImpl() {
+        values = (V[])new Object[ARRAY_LENGTH];
+        keys = (K[]) new Object[ARRAY_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
```
Elena-Bruyako, line 17:
```suggestion
    public void put(K key, V value) {
        if (size == ARRAY_LENGTH) {
                throw new ArrayIndexOutOfBoundsException("Storage is full");
            }
```

You should check first

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
```
Elena-Bruyako, line 45:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

Elena-Bruyako (changes requested):

sarakhmen (approved):
lgtm!

PR #1523, opened 2024-06-24
Implementation key-value

anna-halas:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_CAPACITY = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
Elena-Bruyako, line 14:
```suggestion
```
int type = 0 by default

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null && key == null
```
Elena-Bruyako, line 42:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1544, opened 2024-07-19
Implemented storage of key-value type

YevhenSrdk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ARRAY_SIZE = 10;
+    private int size = 0;
```
MentorMateAcademy, line 7:
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = findIndex(key);
+        if (index != -1) {
+            return (V) values[index];
```
MentorMateAcademy, line 35: better cast in constructor

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+                return i;
+            }
+        }
+        return -1;
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = findIndex(key);
+        if (index != -1) {
```
MentorMateAcademy, line 19:
```suggestion
for (int i = 0; i < size; i++) {
        if (key == keys[i] || key != null && key.equals(keys[i])
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key != null && key.equals(keys[i])) {
+                return i;
+            }
+            if (key == null && keys[i] == null) {
+                return i;
+            }
+        }
+        return -1;
+    }
```
MentorMateAcademy: you don`t need this method

MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
YevhenSrdk: But then how do I find the index of the key? I can iterate over the array in the put() method itself, but won't this worsen the readability of the code, and wouldn't it be more convenient to have a separate method that finds the index of the desired number?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_ARRAY_SIZE];
+        this.values = (V[]) new Object[MAX_ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        boolean keyExists = false;
```
Elena-Bruyako, line 18: use separate method indexOf to avoid duplication in put and get methods

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+    private V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_ARRAY_SIZE];
+        this.values = (V[]) new Object[MAX_ARRAY_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        if (indexOf(key) != -1) {
```
Elena-Bruyako, line 18: why do you need call indexOf method twice?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+            keys[size] = key;
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        if (indexOf(key) != -1) {
```
Elena-Bruyako, line 32: same comment

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1548, opened 2024-07-24
StorageImpl changes

Tiranose:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,45 @@
...
-    @Override
+    private static final int MAX_SIZE = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
okuzan, line 15: 0 is default for this data type (int)

okuzan (changes requested):
Checklist issue - we can't use Objects.equals in this task just yet, write implementation by yourself

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 17: To avoid duplication of code, create separate method `indexOf(K key)` and use it for put and get methods

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
     }
 
-    @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (keysEqual(keys[i], key)) {
```
Elena-Bruyako, line 42: Why do you need separate method here? 
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,56 @@
...
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || key != null && key.equals(keys[i])) {
+                return i;
+            }
+        }
         return -1;
     }
+
+    private boolean keysEqual(K key1, K key2) {
```
Elena-Bruyako, line 49: Do you still need this method?

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1552, opened 2024-07-30
commit task

Oleksii21th:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+                throw new RuntimeException("Storage is full");
+            }
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        int indexKey = findKeyIndex(key);
+        if (indexKey != -1) {
```
beatarjn, line 35: You can simplify the condition: return indexKey != -1 ? values[indexKey] : null;

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     }
 
     @Override
     public int size() {
+        return count;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < count; i++) {
+            if (keys[i].equals(key)) {
```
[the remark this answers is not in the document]
beatarjn: Pls apply comment from ai-mentor. This is the reason your tests fail.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS];
+        values = (V[]) new Object[MAX_ITEMS];
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexKey = findKeyIndex(key);
+        if (indexKey != -1) {
```
beatarjn, line 19: -1 is repeated several times and is a 'magic' number. Pls extract it to a constant with meaningful name.

beatarjn (changes requested):
Pls always make sure that your tests pass before submitting PR.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return count;
+    }
+
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < count; i++) {
+            if (key == null) {
```
beatarjn, line 50: Please refactor the loop - it could be simplified to one if as all of the if/else conditions return the same value.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,60 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS = 10;
+    private static final int KEY_NOT_FOUND = -1;
+    private K[] keys;
+    private V[] values;
+    private int count;
```
beatarjn, line 10: Maybe the better name would be arraySize or something similar

beatarjn (changes requested):

PR #1553, opened 2024-07-31
Added task solution

Lessorg:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,90 @@
...
+    private boolean hasKey(K key) {
+        for (Pair<K, V> pair: storageArray) {
+            if ((pair.getKey() != null && pair.getKey().equals(key)) || pair.getKey() == key) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private class Pair<K, V> {
```
okuzan, line 75: Better make it static if possible, we won't need getters setters either

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,90 @@
...
     @Override
     public void put(K key, V value) {
+        if (hasKey(key)) {
+            replacePair(new Pair<>(key, value));
+        } else {
+            Pair<K, V>[] newStorageArray = getNewStorageArray();
+            newStorageArray[newStorageArray.length - 1] = new Pair<>(key, value);
+            setStorageArray(newStorageArray);
+        }
     }
```
okuzan, line 21:
```suggestion
            replacePair(new Pair<>(key, value));
            return;
        }
        Pair<K, V>[] newStorageArray = getNewStorageArray();
        newStorageArray[newStorageArray.length - 1] = new Pair<>(key, value);
        setStorageArray(newStorageArray);
      }
    }
```

okuzan (changes requested):
Good job so far! Make inner class static

okuzan (approved):
Well done

PR #1566, opened 2024-08-15
first solution

Rzarcik97:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private StorageArray[] storageArray;
```
MateuszRostek, line 7: Let's do it like this:

```java
    private static final int MAX_STORAGE_ELEMENTS = 10;

    private final K[] keys;
    private final V[] values;
    private int size;
```

MateuszRostek: When you make this changes a lot of small changes will be needed in this class, please refactor it and re-request review then.

MateuszRostek (changes requested):
Don't create two separate classes for this solution. It could and should be done with only one.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+            if (isEqualCheck(key,i)) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
```
beatarjn, line 17: As suggested previously add a field  
private int size;
and operate on it so that in this method you will return simply size.
It will represent the actual size of the array, no need to iterate and check if value exist in a index each time. It is time  consuming.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_ELEMENTS = 10;
+
```
beatarjn, line 7: Remove empty space

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+    private final K[] keys;
+    private final V[] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_STORAGE_ELEMENTS];
+        this.values = (V[]) new Object[MAX_STORAGE_ELEMENTS];
+    }
+
+    public boolean isEqualCheck(K key, int i) {
```
beatarjn, line 16: This method is a helper method for that class so should be private. Also better name would be something like isKeyEqual()

beatarjn (changes requested):

MateuszRostek (approved):
Good job!

PR #1571, opened 2024-08-20
Implemented the Storage of the key-value type

tbasy:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS = 10;
+    private K[] keys;
+    private V[] values;
+    private K key;
+    private V value;
+    private int sizeStorage = 0;
```
okuzan, line 11:
```suggestion
    private int size;
```
0 is default for int, let's keep name simple

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_ELEMENTS];
+        this.values = (V[]) new Object[MAX_ELEMENTS];
```
okuzan, line 15: [where do we need ` this` ?](https://stackoverflow.com/a/2411283)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+
+    public V getValue() {
+        return value;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < sizeStorage; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
```
okuzan, line 30: Look up code of Objects.equals, and use this code (but not Objects.equals directly - you need to be able to write this yourself)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < sizeStorage; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    return values[i];
+                }
+            }
```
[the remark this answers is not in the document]
okuzan: agree with bot - Again, consider extracting the logic for null key check and value retrieval into a private method to avoid code duplication, in line with the 'Don't create repeating code' checklist item.
Create a method that would return index of a matched key, -1 if not found

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS = 10;
+    private K[] keys;
+    private V[] values;
+    private K key;
+    private V value;
```
Elena-Bruyako, line 10:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
Elena-Bruyako: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
Elena-Bruyako: not fixed

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+        this.values = (V[]) new Object[MAX_ELEMENTS];
+    }
+
+    public K getKey() {
+        return key;
+    }
+
+    public V getValue() {
+        return value;
+    }
```
Elena-Bruyako, line 24:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,64 @@
...
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < sizeStorage; i++) {
```
Elena-Bruyako, line 28: create indexOf method to avoid duplication of code

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
tbasy: I accidentally sent an old commit, a poorly made push

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        values = (V[]) new Object[MAX_ELEMENTS];
+    }
+
+    public K getKey() {
+        return null;
+    }
+
+    public V getValue() {
+        return null;
+    }
```
okuzan, line 22: ? remove these two

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+        }
     }
 
     @Override
     public V get(K key) {
+        int index = indexOf(key);
+        if (index != -1) {
+            return values[index];
+        }
         return null;
```
okuzan, line 44: not a mistake, but try to write in one line with a ternary operator, as an extra task

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null) {
+                return i;
+            }
+            if (key != null && key.equals(keys[i])) {
+                return i;
+            }
```
okuzan, line 59: body in both `if`s are the same - it means you need to merge conditions into one

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == null && keys[i] == null || key != null && key.equals(keys[i])) {
```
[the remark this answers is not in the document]
tbasy: Мені цей вираз IJ підкреслює. Але я виправила згідно зауважень 
'key == keys[i] || key != null && key.equals(keys[i])' replaceable by 'Objects.equals()' expression

Elena-Bruyako (approved):

PR #1573, opened 2024-08-23
Implemented "StorageImpl" class, added auxiliary class "Pair".

f3ops3:


sarakhmen (approved):
interesting solution :)

PR #1574, opened 2024-08-24
solution added

xotekijke:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
```
Elena-Bruyako, line 55:
```suggestion
    }
    
    private int indexOf(K key) {
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return i;
            }
        }
        return NO_INDEX;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        for (int i = 0; i < size; i++) {
+            if (key == null) {
+                if (keys[i] == null) {
+                    return (V) values[i];
+                }
+            } else if (keys[i] != null && keys[i].equals(key)) {
+                return (V) values[i];
+            }
+        }
         return null;
```
Elena-Bruyako, line 49:
```suggestion
        int index = indexOf(key);
        return index == NO_INDEX ? null : values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 18:
```suggestion
        int index = indexOf(key);
        if (index != NO_INDEX) {
            values[index] = value;
        } else if (index == NO_INDEX) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
+    private static final int MAX_SIZE = 10;
+    private static final int NO_INDEX = -1;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_SIZE];
+        values = (V[]) new Object[MAX_SIZE];
+        size = 0;
```
Elena-Bruyako, line 15:
```suggestion
```

int type = 0 by default, so you don't need to initialise it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (key == keys[i] || (key != null && key.equals(keys[i]))) {
```
Elena-Bruyako, line 45:
```suggestion
            if (key == keys[i] || key != null && key.equals(keys[i])) {
```

Elena-Bruyako (changes requested):

PR #1581, opened 2024-09-08
Solution

vsupruniuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_ELEMENTS_LENGTH = 10;
+    public static final int NOT_FOUND_ELEM_IDX = -1;
```
liliia-ponomarenko, line 10: make these constants private

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
...
+
+    private final Integer[] keys = new Integer[MAX_ELEMENTS_LENGTH];
+    private V[] values;
```
liliia-ponomarenko, line 13: All non-constant fields should be initialized in the constructor. Why do you initialize values array in the separate method?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
...
+        this.values = (V[]) Array.newInstance(
+                valuesClass,
+                MAX_ELEMENTS_LENGTH
+        );
+    }
+
+    private int getKeyIndex(K key) {
+        for (int i = 0; i < this.keys.length; i += 1) {
+            if (keys[i] != null
+                    && keys[i] == (key != null ? key.hashCode() : 0)
```
liliia-ponomarenko, line 64: will this check be correct if key == null? 
null and 0 it's not the same

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
...
 import core.basesyntax.Storage;
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_ELEMENTS_LENGTH = 10;
+    public static final int NOT_FOUND_ELEM_IDX = -1;
+
+    private final Integer[] keys = new Integer[MAX_ELEMENTS_LENGTH];
```
liliia-ponomarenko, line 12: Let's parametrize keys it doesn't have to be Integer

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,72 @@
...
     }
 
     @Override
     public int size() {
-        return -1;
+        return Arrays
+                .stream(this.keys)
+                .filter(Objects::nonNull)
+                .toArray()
+                .length;
```
liliia-ponomarenko, line 50: Simplify this method by introducing a size class field. (This principle is included in the work of all collections)

liliia-ponomarenko (changes requested):
Good job! Let's fix your solution ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ELEMENTS_LENGTH = 10;
+    private static final int NOT_FOUND_ELEM_IDX = -1;
+
+    private final K[] keys;
+    private final V[] values;
+    private int size = 0;
```
liliia-ponomarenko, line 11:
```suggestion
    private int size;
```
0 is default value int type

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,62 @@
...
+        this.keys = (K[]) new Object[MAX_ELEMENTS_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        int existingKeyIndex = this.getKeyIndex(key);
+
+        if (existingKeyIndex != NOT_FOUND_ELEM_IDX) {
+            if (this.values[existingKeyIndex] == null) {
+                this.size += 1;
```
liliia-ponomarenko, line 25:
```suggestion
                this.size++;
```
Also you can move it after if-else block

liliia-ponomarenko (approved):
Good job! A few minor comments ;)

PR #1591, opened 2024-09-20
First solution of the task, with two arrays T nad V, that are impleme…

B4rt1omiej:
…nts and chacked by for loop.

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,46 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] == key || (keys[i] == null && values[i] == null)) {
+                keys[i] = key;
+                if (values[i] == null) {
+                    size++;
+                }
+                values[i] = value;
+                return;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] != null && keys[i].equals(key) || key == keys[i]) {
+                return values[i];
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        return size;
     }
 }


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] == key || (keys[i] == null && values[i] == null)) {
```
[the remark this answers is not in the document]
B4rt1omiej: but in checklist is wrote "Don't use class Objects"

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                keys[i] = key;
+                if (values[i] == null) {
+                    size++;
+                }
+                values[i] = value;
+                return;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
```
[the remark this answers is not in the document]
B4rt1omiej: but in checklist is wrote "Don't use class Objects"

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                values[i] = value;
+                return;
+            }
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
+            if (keys[i] != null && keys[i].equals(key) || key == keys[i]) {
```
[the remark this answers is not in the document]
B4rt1omiej: but in checklist is wrote "Don't use class Objects"

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+                }
+                values[i] = value;
+                return;
+            }
+            if (keys[i] != null && keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
+        }
     }
```
slade13, line 32: You should handle the case when the storage is full (size has reached MAX_ITEMS_NUMBER). Currently, if one tries to put more elements than MAX_ITEMS_NUMBER, it will result in an ArrayIndexOutOfBoundsException.

src/main/java/core/basesyntax/impl/StorageImpl.java
slade13, line 19: This condition can result in an incorrect match if there are null keys in the array. You should check first if the key is null, and if it is not, then use the equals method.

src/main/java/core/basesyntax/impl/StorageImpl.java
slade13, line 37: The same null-safety problem as above. First check if the key is null and only then use equals method.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
```
slade13, line 36: The same issue with the loop boundaries as above. It should iterate from 0 to size.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private int size;
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
```
slade13, line 18: The loop should iterate from 0 to size, not MAX_ITEMS_NUMBER. The max size is the limit, but you may have less actual elements in the storage.

slade13 (changes requested):
Few changes are required, which may lead to some unexpected behavior of your program, but overall solution looks very nice!

B4rt1omiej:
i don't know what you ment in "null-safety problem" as i try chnage it as you suguest all my code get in mess and didn't work, so thats why i didn't changed anything there due to my lack of experience in this becouse i don't know how i can make it work better.

B4rt1omiej:
i don't know what you ment in "null-safety problem" as i try chnage it as you suguest all my code get in mess and didn't work, so thats why i didn't changed anything there due to my lack of experience in this becouse i don't know how i can make it work better.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_ITEMS_NUMBER];
+        values = (V[]) new Object[MAX_ITEMS_NUMBER];
+    }
+
     @Override
     public void put(K key, V value) {
+        try {
+            for (int i = 0; i < MAX_ITEMS_NUMBER; i++) {
```
MateuszRostek, line 19:
```suggestion
            for (int i = 0; i < size; i++) {
```

MateuszRostek (approved):
Please remove these `try` blocks.

PR #1598, opened 2024-10-03
Implement the Storage of the key-value type

OlehRidchenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
+    private static final int MAX_LENGTH_OF_ARRAY = 10;
+    private final K[] keys;
+    private final V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_LENGTH_OF_ARRAY];
+        values = (V[]) new Object[MAX_LENGTH_OF_ARRAY];
+        size = 0;
```
okuzan, line 16: 0 is default value anyway

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,57 @@
...
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexOf(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 51: checklist issue, you need to write impl yourself, can't use objects.equals

okuzan (changes requested):

PR #1611, opened 2024-10-07
done

justVovan4ik:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keys;
+    private final Object[] values;
+    private int size = 0;
```
okuzan, line 9: 0 is default for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEMS_NUMBER = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
okuzan, line 8: use your type parameters K,V

okuzan (changes requested):

PR #1614, opened 2024-10-12
implemented the 'Storage' of the 'key-value' type

opdhk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Arrays;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int STORAGE_DEFAULT_CAPACITY = 10;
+    private int size = 0;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+            }
+        }
+        if (size == STORAGE_DEFAULT_CAPACITY - 1) {
+            increaseStorageSize();
+        }
+        items[size] = new Pair<>(key, value);
+        size++;
+    }
+
+    private boolean areBothKeysNull(K firstKey, K secondKey) {
```
Elena-Bruyako, line 34: private methods should be placed after all public methods

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,63 @@
...
+        return secondKey == null && firstKey == null;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 40: there is duplication of code, please refactor

Elena-Bruyako (changes requested):

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -14,31 +15,25 @@ public StorageImpl() {
...
-                return;
+        if (findIndexOfItemByKey(key) == INDEX_OF_ABSENT_KEY) {
+            try {
+                items[size] = new Pair<>(key, value);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                increaseStorageSize();
+                items[size] = new Pair<>(key, value);
+            } finally {
+                size++;
             }
```
sarakhmen, line 26: do you really need this excessive try catch block here? I guess you can simply check the size instead

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,50 @@
...
+            }
+            if (items[i].getKey() != null && items[i].getKey().equals(key)) {
+                items[i] = new Pair<>(key, value);
+                return;
+            }
+        }
+        if (size < STORAGE_DEFAULT_CAPACITY - 1) {
+            items[size] = new Pair<>(key, value);
+            size++;
+        }
```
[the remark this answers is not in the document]
sarakhmen: findIndexOfItemByKey is executed twice, let's optimize it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -14,31 +15,25 @@ public StorageImpl() {
...
     public V get(K key) {
-        for (int i = 0; i < size; i++) {
-            if (areKeysEqual(key, items[i].getKey()) || areBothKeysNull(items[i].getKey(), key)) {
-                return items[i].getValue();
-            }
-        }
-        return null;
+        return findIndexOfItemByKey(key) == -1
+                ? null
+                : items[findIndexOfItemByKey(key)].getValue();
```
sarakhmen, line 36: same issue

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -55,6 +50,17 @@ private boolean areKeysEqual(K firstKey, K secondKey) {
...
     private void increaseStorageSize() {
-        items = Arrays.copyOf(items, items.length * 2);
+        Pair<K, V>[] doubleSizedArray = new Pair[items.length * ARRAY_SIZE_MULTIPLICATOR];
+        System.arraycopy(items, 0, doubleSizedArray, 0, items.length);
+        items = doubleSizedArray;
+    }
+
+    private int findIndexOfItemByKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if (areKeysEqual(key, items[i].getKey()) || areBothKeysNull(items[i].getKey(), key)) {
```
sarakhmen, line 60: if first and second key == null then they are equal ;) Please refactor your code

sarakhmen (changes requested):
let's improve your solution :)

src/main/java/core/basesyntax/impl/Pair.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+        if (pair == null || !pair.getClass().equals(Pair.class)) {
+            return false;
+        }
+        Pair<K,V> current = (Pair<K,V>) pair;
+        if (getKey() != null ? !getKey().equals(current.getKey()) : current.getKey() != null) {
+            return false;
+        }
+        if (getValue() != null ? !getValue().equals(current.getValue()) :
+                current.getValue() != null) {
```
ahoienko, line 33: Code duplication

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,59 @@
...
 
     @Override
     public int size() {
-        return -1;
+        return size;
+    }
+
+    private boolean areKeysEqual(K firstKey, K secondKey) {
+        return secondKey != null && secondKey.equals(firstKey) || firstKey == secondKey;
+    }
```
ahoienko, line 45: Could be moved to Pair class

ahoienko (changes requested):
Nice, left few last minor comments

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+        int indexOfItemByKey = findIndexOfItemByKey(key);
+        if (indexOfItemByKey == INDEX_OF_ABSENT_KEY) {
+            if (size == items.length) {
+                increaseStorageSize();
+            }
+            items[size] = new Pair<>(key, value);
+            size++;
+        } else {
+            items[indexOfItemByKey] = new Pair<>(key, value);
+        }
```
```suggestion
        if (indexOfItemByKey != INDEX_OF_ABSENT_KEY) {
             items[indexOfItemByKey] = new Pair<>(key, value);
             return;
        } 
        if (size == items.length) {
           increaseStorageSize();
        }
        items[size] = new Pair<>(key, value);
        size++;
```

okuzan: you have two levels of nested code (which is harder to read), you can do it in only one. If you have  if-else, if clause is big, and else clause is just one line, like here - invert condition, [return early](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement), and unnest else clause, removing else (we don't need else after returning from a function)

okuzan (changes requested):
Almost done

okuzan (approved):

PR #1618, opened 2024-10-15
my storage

VovaPryt:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Object[] keys = new Object[MAX_SIZE];
```
Elena-Bruyako, line 7: it's better to initialize in constructor

```suggestion
    private K[] keys;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private Object[] values = new Object[MAX_SIZE];
```
Elena-Bruyako, line 8: same here
```suggestion
    private V[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+    private int size = 0;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,48 @@
...
+        int index = findKeyIndex(key);
+        if (index != -1) {
+            return (V) values[index];
+        }
         return null;
     }
 
+    private int findKeyIndex(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
Elena-Bruyako, line 39:
```suggestion
            if (keys[i] == key || keys[i] != null && keys[i].equals(key)) {
```

Elena-Bruyako (changes requested):

PR #1621, opened 2024-10-18
Implement the Storage

DariaBakal:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,64 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_CAPACITY = 10;
+    private final Object[] keyStorage;
```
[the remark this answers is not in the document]
okuzan: Issue: The checklist specifies not to use class Objects. Consider using a different approach to store keys and values without using Object arrays.
use K,V

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_CAPACITY = 10;
+    private final Object[] keyStorage;
+    private final Object[] valueStorage;
+    private int storageCapacity;
+    private V nullKeyStorage;
```
okuzan, line 10: don't handle null key separately, if you have written correct comparison logic, it should work with both null and not null

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+    private boolean hasNullKey;
+
+    public StorageImpl() {
+        keyStorage = new Object[MAX_STORAGE_CAPACITY];
+        valueStorage = new Object[MAX_STORAGE_CAPACITY];
+        storageCapacity = 0;
+        hasNullKey = false;
```
okuzan, line 17:
```suggestion
```
those were default values

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
     @Override
     public V get(K key) {
+        if (key == null) {
+            return hasNullKey ? nullKeyStorage : null;
+        }
+        for (int i = 0; i < keyStorage.length; i++) {
+            if (keyStorage[i] != null && keyStorage[i].equals(key)) {
+                return (V) valueStorage[i];
+            }
+        }
```
okuzan, line 54: try to extract this logic to a private method and reuse in both public methods 
return -1 if not found

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,61 @@
...
+                valueStorage[i] = value;
+                return;
+            }
+        }
+        if (storageCapacity < MAX_STORAGE_CAPACITY) {
+            keyStorage[storageCapacity] = key;
+            valueStorage[storageCapacity] = value;
+            storageCapacity++;
+        } else {
+            System.out.println("Out of storage capacity");
```
okuzan, line 41: don't handle this case, out of scope of the task

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
     }
 
     @Override
     public int size() {
+        return storageCapacity;
+    }
+
+    private int getIndexOfValue(K[] keyStorage, K key) {
+        for (int i = 0; i < keyStorage.length; i++) {
+            if ((keyStorage[i] == key) || (keyStorage[i] != null && keyStorage[i].equals(key))) {
```
Elena-Bruyako, line 49:
```suggestion
            if (keyStorage[i] == key || keyStorage[i] != null && keyStorage[i].equals(key)) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+        indexOfValue = getIndexOfValue(keyStorage, key);
+        return indexOfValue >= 0 ? valueStorage[indexOfValue] : null;
     }
 
     @Override
     public int size() {
+        return storageCapacity;
+    }
+
+    private int getIndexOfValue(K[] keyStorage, K key) {
```
Elena-Bruyako, line 47:
```suggestion
    private int getIndexOfValue(K key) {
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_STORAGE_CAPACITY = 10;
+    private final K[] keyStorage;
+    private final V[] valueStorage;
+    private int storageCapacity;
```
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private int indexOfValue;
+    private boolean hasNullKey;
```
Elena-Bruyako, line 11:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+            keyStorage[storageCapacity] = key;
+            valueStorage[storageCapacity] = value;
+            storageCapacity++;
+        }
     }
 
     @Override
     public V get(K key) {
-        return null;
+        indexOfValue = getIndexOfValue(keyStorage, key);
```
Elena-Bruyako, line 38:
```suggestion
        int indexOfValue = getIndexOfValue(keyStorage, key);
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private boolean hasNullKey;
+
+    public StorageImpl() {
+        keyStorage = (K[]) new Object[MAX_STORAGE_CAPACITY];
+        valueStorage = (V[]) new Object[MAX_STORAGE_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        indexOfValue = getIndexOfValue(keyStorage, key);
```
```suggestion
        int indexOfValue = getIndexOfValue(keyStorage, key);
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
DariaBakal: @Elena-Bruyako what logic should I implement for the null key if I delete the private boolean hasNullKey? tests fail without it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    public StorageImpl() {
+        keyStorage = (K[]) new Object[MAX_STORAGE_CAPACITY];
+        valueStorage = (V[]) new Object[MAX_STORAGE_CAPACITY];
+    }
+
     @Override
     public void put(K key, V value) {
+        int indexOfValue = getIndexOfValue(key);
+        if (indexOfValue >= 0) {
```
Elena-Bruyako, line 20: You can just use 
```suggestion
        int index = indexOf(key);
        if (index != -1) {
            values[index] = value;
        } else if (index == -1) {
            keys[size] = key;
            values[size] = value;
            size++;
        }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+        int indexOfValue = getIndexOfValue(key);
+        return indexOfValue >= 0 ? valueStorage[indexOfValue] : null;
     }
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int getIndexOfValue(K key) {
```
Elena-Bruyako, line 46: Are you sure that method is getting index of value?

Elena-Bruyako (changes requested):

PR #1638, opened 2024-11-13
fixed impl

LizaBalshv:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAXIMUM_CAPACITY = 10;
+    private final Object[] keys;
+    private final Object[] values;
```
okuzan, line 8: use your K, V class type parameter, that's why we have them in the first place - to avoid using object

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    private int size = 0;
```
okuzan, line 9: 0 is default value for int

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+            values[size] = value;
+            size++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null)
+                    || (keys[i] != null && keys[i].equals(key))) {
```
okuzan, line 37: extract repeated logic from put and get into a private method that searches for a match and returns its index and -1 otherwise

okuzan (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,49 @@
...
 
     @Override
     public int size() {
+        return size;
+    }
+
+    private int indexKey(K key) {
+        for (int i = 0; i < size; i++) {
+            if ((keys[i] == null && key == null)
+                    || (keys[i] != null && keys[i].equals(key))) {
```
liliia-ponomarenko, line 45:
```suggestion
            if (keys[i] == null && key == null
                    || keys[i] != null && keys[i].equals(key)) {
```
redundant brackets

liliia-ponomarenko (approved):
Awesome 😎

PR #1642, opened 2024-11-19
i want to give some explain

Sasha1256:


liliia-ponomarenko (changes requested):
Unfortunately, CI checks failed, let's fix them first. Use informative messages for comments and PR.

PR #1651, opened 2024-12-05
initial commit

Tetiana-Hrytsenko:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
     public void put(K key, V value) {
+        for (int i = 0; i < currentSize; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+        if (currentSize >= MAX_STORAGE_SIZE) {
+            throw new IllegalStateException("Storage is full");
+        } else {
```
okuzan, line 28: no need to use else after returning from the function (including throwing)

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,49 @@
...
+            keys[currentSize] = key;
+            values[currentSize] = value;
+            currentSize++;
+        }
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < currentSize; i++) {
+            if (Objects.equals(keys[i], key)) {
```
okuzan, line 38: Common mistake checklist - don't use class Objects in your solution, write your own function for that

okuzan (changes requested):

Elena-Bruyako (approved):

PR #1677, opened 2025-01-19
done

xMaXD:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,57 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int defaultCapacity = 10;
```
```suggestion
    private static final int DEFAULT_CAPACITY = 10;
```
follow constant name conversation.

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
xMaXD: Abbreviation in name 'DEFAULT_CAPACITY' must contain no more than '2' consecutive capital letters. [AbbreviationAsWordInName]
Member name 'DEFAULT_CAPACITY' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]

xMaXD: Maven not allow me to make it like this "DEFAULT_CAPACITY"

liliia-ponomarenko (approved):
Great!

PR #1682, opened 2025-01-31
first solution

anpiet:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys;
+    private Object[] values;
+    private int index;
+
+    public StorageImpl() {
+        int maxArrSize = 10;
```
slade13, line 12: Replace this local variable with constant field: `private static final int MAX_SIZE = 10;`.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys;
+    private Object[] values;
+    private int index;
```
slade13, line 9: Rename this variable, so the name will better suit it's purpose.
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+        for (int i = 0; i < index; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
+
+        keys[index] = key;
+        values[index] = value;
+        index++;
```
slade13, line 30: Add check to ensure that maximum capacity is not exceeded.
```suggestion
        if (size < MAX_SIZE) {
            keys[size] = key;
            values[size] = value;
            size++;
        } else {
            throw new IllegalStateException("Storage is full. Maximum capacity is " + MAX_SIZE);
        }
```

slade13 (approved):
Good job on implementing this task, please review my comments.

PR #1688, opened 2025-02-09
done

mizuo-san:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,97 @@
...
+        }
+
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
+                tempValues[i] = value;
+                flag = true;
+                keys = tempKeys;
+                values = tempValues;
+                return;
+            }
```
liliia-ponomarenko, line 44: move the logic searching  key to the private method and reuse it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,97 @@
...
+            return;
+        }
+
+        for (int i = 0; i < keys.length; i++) {
+            tempKeys[i] = keys[i];
+            tempValues[i] = values[i];
+        }
+
+        for (int i = 0; i < keys.length; i++) {
+            if ((keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key))) {
```
liliia-ponomarenko, line 38:
```suggestion
            if (keys[i] == null && key == null || keys[i] != null && keys[i].equals(key)) {
```
redundant brackets. Read about logic operators' priority

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,97 @@
...
+            }
+        }
         return null;
     }
 
     @Override
     public int size() {
-        return -1;
+        if (flag) {
+            return keys.length - 1;
```
liliia-ponomarenko, line 67: just return size value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,97 @@
...
+            StorageImpl current = (StorageImpl) pair;
+            return (values == null ? current.values == null : Arrays.equals(values, current.values))
+                    && (keys == null ? current.keys == null : Arrays.equals(keys, current.keys));
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 17;
```
liliia-ponomarenko, line 92: Do you really need these methods?

mizuo-san: tbh not but as i read in one of modules - better to redefine this methods

src/main/java/core/basesyntax/impl/StorageImpl.java
mizuo-san: thank you

PR #1701, opened 2025-02-19
v1

PiotrHic:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,31 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.HashMap;
+import java.util.Map;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private Map<K,V> map = new HashMap<>();
```
slade13, line 9: The implementation uses a HashMap to store key-value pairs, which is not suitable for the task requirement of using one or two arrays to store a maximum of 10 elements. Consider using two arrays, one for keys and one for values, to meet the task requirements.

slade13 (changes requested):
You're not allowed to use any List, Set or Map, this was noted inside the task description: https://github.com/mate-academy/jv-storage/blob/master/checklist.md#dont-use-any-kind-of-list-set-or-map

slade13 (changes requested):
You're not allowed to use any List, Set or Map, this was noted inside the task description: https://github.com/mate-academy/jv-storage/blob/master/checklist.md#dont-use-any-kind-of-list-set-or-map

PR #1722, opened 2025-03-14
implement class StorageImpl<K, V>

klv-liliya:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_NUMBER_OF_ELEMENTS = 10;
+    private K[] key;
+    private V[] value;
+    private int size = 0;
```
Elena-Bruyako, line 9: int type = 0 by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+            if (this.key[i] == key
+                    || (this.key[i] != null
+                    && key != null
+                    && this.key[i].equals(key))) {
+                this.value[i] = value;
+                return;
+            }
+        }
+        this.key[size] = key;
+        this.value[size] = value;
```
Elena-Bruyako, line 28:
```suggestion
        for (int i = 0; i < size; i++) {
            if (key == keys[i]
                    || keys[i] != null && keys[i].equals(key)) {
                value[i] = value;
                return;
            }
        }
        key[size] = key;
        value[size] = value;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (this.key[i] == key
```
Elena-Bruyako, line 35: same comment here

Elena-Bruyako (changes requested):

PR #1747, opened 2025-05-16
implement StorageImpl with fixed-size array and support for null keys

Mwxyz2022:
- Used Object[][] array to store key-value pairs (max 10 items)
- Implemented put(), get(), and size() methods
- Added logic to update value if key already exists
- get() method supports null keys and uses @SuppressWarnings to handle cast
- Added helper method getItemIndex() to find key index in storage

### DIFF
--- a/src/main/java/core/basesyntax/impl/StorageImpl.java
+++ b/src/main/java/core/basesyntax/impl/StorageImpl.java
@@ -3,17 +3,56 @@ package core.basesyntax.impl;
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
-    @Override
-    public void put(K key, V value) {
+    private static final int MAX_ITEM = 10;
+    private final Object[][] storage;
+
+    public StorageImpl() {
+        this.storage = new Object[MAX_ITEM][];
     }
 
     @Override
+    public void put(K key, V value) {
+        int itemIndex = getItemIndex(key);
+        if (itemIndex == -1) {
+            int nextEmptyIndex = size();
+            storage[nextEmptyIndex] = new Object[]{key, value};
+            return;
+        }
+
+        storage[itemIndex][1] = value;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
     public V get(K key) {
-        return null;
+        int itemIndex = getItemIndex(key);
+        if (itemIndex == -1) {
+            return null;
+        }
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEM = 10;
+    private final Object[][] storage;
+    private int countFillItems = 0;
```
Elena-Bruyako:
```suggestion
    private int size = 0;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
 
     @Override
     public int size() {
+        return countFillItems;
+    }
+
+    private int getItemIndex(K key) {
+        for (int i = 0; i < countFillItems; i++) {
+            if ((key == null && storage[i][0] == null)
+                    || (key != null && key.equals(storage[i][0]))) {
```
Elena-Bruyako:
```suggestion
            if (key == storage[i][0]
                    || key != null && key.equals(storage[i][0])) {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_ITEM = 10;
+    private final Object[][] storage;
+    private int size = 0;
```
liliia-ponomarenko, line 8:
```suggestion
    private int size;
```
default value

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+
+    public StorageImpl() {
+        this.storage = new Object[MAX_ITEM][2];
```
liliia-ponomarenko, line 11: two-dimensional array is not really a clear solution. Better have separate arrays or create an object that represents the product. but in general it is work. so you can try refactor it if you want.

liliia-ponomarenko (approved):
good job!

PR #1751, opened 2025-05-26
Test passed

Lentezz:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private int capacity = 10;
+    private final Node<K, V>[] table = new Node[capacity];
+    private int currentIndex = 0;
```
liliia-ponomarenko:
```suggestion
    private int currentIndex;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
+            currentIndex++;
+        } else {
+            for (Node node : table) {
+                if ((node != null && node.getKey() != null
+                        && node.getKey().equals(key))
+                        || (node != null && key == null
+                        && node.getKey() == null)) {
+                    node.setValue(value);
+                }
+            }
```
liliia-ponomarenko: You have similar logic in the get method. Let’s  move it to the private method and reuse

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
     public void put(K key, V value) {
+        if (this.get(key) == null && size() < capacity - 1) {
+            table[currentIndex] = new Node(key, value);
+            currentIndex++;
+        } else {
+            for (Node node : table) {
+                if ((node != null && node.getKey() != null
+                        && node.getKey().equals(key))
+                        || (node != null && key == null
+                        && node.getKey() == null)) {
```
liliia-ponomarenko: too complicated conditional. Think about how to simplify it

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
...
 
     @Override
     public int size() {
-        return -1;
+        int size = 0;
+        for (Node node : table) {
+            if (node != null) {
+                size++;
+            }
+        }
```
liliia-ponomarenko:
```suggestion
```
Make size class-level variable

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,66 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+
+    private int capacity = 10;
```
liliia-ponomarenko: It should constant field INITIAL_CAPACITY

liliia-ponomarenko (changes requested):

Elena-Bruyako (approved):

PR #1767, opened 2025-07-09
Ready

DmitryShmerchuk:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    public static final int MAX_CAPACITY = 10;
```
liliia-ponomarenko:
```suggestion
   private static final int MAX_CAPACITY = 10;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+    private K[] keys;
+    private V[] values;
+    private int size;
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_CAPACITY];
+        values = (V[]) new Object[MAX_CAPACITY];
+        size = 0;
```
liliia-ponomarenko:
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                values[i] = value;
+                return;
+            }
+        }
```
liliia-ponomarenko: Don't use class Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(keys[i], key)) {
+                return values[i];
+            }
+        }
```
liliia-ponomarenko: I see similar code in the method put. Let's move it to the private method

liliia-ponomarenko (changes requested):
Good job! A few commnets ;)

src/main/java/core/basesyntax/impl/StorageImpl.java
DmitryShmerchuk: I did it that way at first:
if (key == null ? keys[i] == null : key.equals(keys[i]))
but IntelliJ itself suggested changing it to Objects.equals() 😅

Elena-Bruyako (approved):

PR #1820, opened 2025-10-02
StorageImpl implemented

tufned:


tufned:
I am not using Objects class, thus I do not violet checklist "Don't use class Objects". I am using only class Object, not Objects. Object and Objects are two different classes

PR #1829, opened 2025-10-31
implement the StorageImpl class

isher2k1:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+    private int index;
+
+    public StorageImpl() {
+        Object[] genericKeys = new Object[ARRAY_SIZE];
+        Object[] genericValues = new Object[ARRAY_SIZE];
+
+        this.keys = (K[]) genericKeys;
+        this.values = (V[]) genericValues;
+        this.index = 0;
+    }
```
sarakhmen: index will be 0 by default so no need to explicitly initialize it here

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,52 @@
...
+
+    private boolean isKey(int i, K key) {
+        return (keys[i] == null && key == null) || (keys[i] != null && keys[i].equals(key));
+    }
```
sarakhmen: method naming actually is not clear. Also private methods should be placed after public ones

sarakhmen (changes requested):
let's slightly improve your solution :)

sarakhmen (approved):

PR #1848, opened 2026-01-03
Implement Storage solution

b1gsal:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+        }
+        keys[size] = key;
+        values[size] = value;
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
Elena-Bruyako: Don't use class Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,44 @@
...
+
+    public StorageImpl() {
+        keys = (K[]) new Object[ARRAY_LENGTH];
+        values = (V[]) new Object[ARRAY_LENGTH];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key, keys[i])) {
```
Elena-Bruyako: Don't use class Objects

Elena-Bruyako (changes requested):
see comments

sarakhmen (approved):

PR #1855, opened 2026-01-26
storage task implementation

qreqit:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private Object[] keys = new Object[10];
```
fetis-off: try to avoid using magic numbers in your code

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,47 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                if (key == null) {
+                    values[i] = value;
+                    return;
+                }
+            } else if (keys[i].equals(key)) {
+                values[i] = value;
+                return;
+            }
```
fetis-off:
```suggestion
```

Don't create repeating code (you have this logic in get method)
If the logic of your code repeats - move it to the separate private method for example 
findKey().

fetis-off (changes requested):
Great! Left a few comments so let's improve your solution

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static int COUNT_OF_ELEMENTS = 10;
+    private static int NO_SUCH_INDEX = -1;
+    private Object[] keys = new Object[COUNT_OF_ELEMENTS];
+    private Object[] values = new Object[COUNT_OF_ELEMENTS];
```
Elena-Bruyako: [All non-constant fields should be initialized in the constructor](https://github.com/mate-academy/jv-storage/blob/master/checklist.md#all-non-constant-fields-should-be-initialized-in-the-constructor)

```suggestion
    private Object[] keys;
    private Object[] values;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    private int size = 0;
```
Elena-Bruyako: int type = 0 by default
```suggestion
    private int size;
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+        for (int i = 0; i < size; i++) {
+            if (keys[i] == null) {
+                if (key == null) {
+                    return i;
+                }
+            } else if (keys[i].equals(key)) {
+                return i;
+            }
+        }
+        return NO_SUCH_INDEX;
```
Elena-Bruyako, line 51:
```suggestion
    private int getKeyIndex(K key) {
        for (int i = 0; i < size; i++) {
            if (key == keys[i] || key != null && key.equals(keys[i])) {
                return i;
            }
        }
        return NO_SUCH_INDEX;
    }
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
     @Override
     public V get(K key) {
-        return null;
+        int index = getKeyIndex(key);
+
+        if (index == -1) {
+            return null;
+        }
+
+        return (V) values[index];
```
Elena-Bruyako:
```suggestion
        int index = getKeyIndex(key);
        return index == NO_SUCH_INDEX ? null : values[index];
```

Elena-Bruyako (changes requested):
Let's improve your solution

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,51 @@
...
+    public StorageImpl() {
+        keys = new Object[COUNT_OF_ELEMENTS];
+        values = new Object[COUNT_OF_ELEMENTS];
+    }
+
     @Override
     public void put(K key, V value) {
+        int index = getKeyIndex(key);
+
+        if (index != -1) {
```
fetis-off:
```suggestion
        if (index != NO_SUCH_INDEX) {
```

fetis-off (changes requested):
Great work! Left one comment 
And also please rename your pull request properly e.g "storage task implementation" or something like that to avoid misunderstanding

fetis-off (approved):
Good one!

PR #1873, opened 2026-02-13
Solution implementation

tylerj231:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,54 @@
...
+    private int size = 0;
+    private final K [] keys;
+    private final V [] values;
+
+    public StorageImpl() {
+        this.keys = (K[]) new Object[MAX_CAPACITY];
+        this.values = (V[]) new Object[MAX_CAPACITY];
+    }
+
+    private int findIndexOf(K key) {
```
Elena-Bruyako, line 16: private method should be placed after all public methods

Elena-Bruyako (changes requested):
Good job
One comment

PR #1875, opened 2026-02-19
logic for StorageImpl class created

ReduxR:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    private static final int MAX_ARRAY_SIZE = 10;
+
+    private final Object[] keys;
+    private final Object[] values;
+    private int size;
+
+    public StorageImpl() {
+        this.keys = new Object[MAX_ARRAY_SIZE];
+        this.values = new Object[MAX_ARRAY_SIZE];
+        this.size = 0;
```
Elena-Bruyako: int type = 0 by default
```suggestion
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,53 @@
...
+    }
+
+    private boolean keysEqual(Object a, Object b) {
```
Elena-Bruyako: rename a and b, bad naming

private method should be placed after all public methods

Elena-Bruyako (changes requested):
Good job! See comments

fetis-off (approved):
Good one!

PR #1896, opened 2026-03-25
Storage - first solution

mate-student26:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
 package core.basesyntax.impl;
 
 import core.basesyntax.Storage;
+import java.util.Objects;
```
slade13: Your implementation uses java.util.Objects, this directly violates the task checklist requirement:

> Don't use class Objects

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -1,19 +1,48 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_SIZE];
+        values = (V[]) new Object[MAX_STORAGE_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            if (Objects.equals(key,keys[i])) {
```
slade13: There should be a space after the comma:
```suggestion
            if (Objects.equals(key, keys[i])) {
```

slade13 (changes requested):
The implementation is generally correct, but it uses `Objects.equals()`, which is explicitly forbidden in this task. Please replace this comparison with manual null-safe key comparison logic. After that, the solution should be fine.

slade13 (changes requested):
Your code does not build now: https://github.com/mate-academy/jv-storage/actions/runs/23595932059/job/68712557766?pr=1896

> Error:  Failures: 
> Error:    StorageImplTest.addTwoElementsWithSameKey:140 With two elements added with the same key, the value should be rewritten expected:<[Four]> but was:<[Two]>
> Error:    StorageImplTest.getElementWhenKeyIsObject:118 Test failed! Method get(key) should return correct value expected:<Element 2> but was:<null>
> Error:  Errors: 
> Error:    StorageImplTest.addTwoElementsWithNullKey:155 » NullPointer Key or value cannot be null.
> Error:    StorageImplTest.getElementWithKeyNull:82 » NullPointer Key or value cannot be null.
> [INFO] 
> Error:  Tests run: 8, Failures: 2, Errors: 2, Skipped: 0

The implementation has two functional issues. First, it rejects `null` keys, while this task expects null key support. Second, keys are compared with `==` instead of logical equality, so equal objects are not treated as the same key. Because of that, retrieving and replacing values by equal object keys does not work correctly.

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,46 @@
...
+    public StorageImpl() {
+        keys = (K[]) new Object[MAX_STORAGE_SIZE];
+        values = (V[]) new Object[MAX_STORAGE_SIZE];
+        size = 0;
+    }
+
     @Override
     public void put(K key, V value) {
+
+        for (int i = 0; i < size; i++) {
```
mateuszwojtkowiak, line 20: You have duplicated key‑searching logic in both `put()` and `get()`.  
This violates the “Don’t repeat yourself” rule from the checklist.

Please extract this repeated logic into a private helper method, for example:

private int indexOfKey(K key) { ... }

Then call this method from both `put()` and `get()` to avoid code duplication.

mateuszwojtkowiak (approved):
Great job! The only thing that could be improved is repeated logic in two methods. The task is accepted, but please read my comment and create a helper method to avoid duplication.

PR #1910, opened 2026-04-30
Maksym Gorodnychyi

MaksymGorodnychyi:


src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
 import core.basesyntax.Storage;
 
 public class StorageImpl<K, V> implements Storage<K, V> {
+    private static final int MAX_SIZE = 10;
+    private Entry<K, V>[] entries;
+    private int size;
+
+    private static class Entry<K, V> {
```
Elena-Bruyako, line 10: inner class should be placed at the end of outer class

src/main/java/core/basesyntax/StorageImpl.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+public class StorageImpl<K, V> implements Storage<K, V> {
```
Elena-Bruyako, line 3: Why there is 2 StorageImpl classes?

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+                return;
+            }
+        }
+        entries[size] = new Entry<>(key, value);
+        size++;
     }
 
     @Override
     public V get(K key) {
+        for (int i = 0; i < size; i++) {
```
Elena-Bruyako, line 42: create 
```
private int indexOf(K key) {
                return i;
            }
        }
        return NO_INDEX;
    }
```

```suggestion
        int index = indexOf(key);
        return index == NO_INDEX ? null : values[index];
```

src/main/java/core/basesyntax/impl/StorageImpl.java
```diff
@@ -3,17 +3,55 @@
...
+
+    @SuppressWarnings("unchecked")
+    public StorageImpl() {
+        entries = new Entry[MAX_SIZE];
+    }
+
     @Override
     public void put(K key, V value) {
+        for (int i = 0; i < size; i++) {
+            boolean keysAreEqual = entries[i].key == null
```
Elena-Bruyako, line 28: think how to modify method according to my previous comment

Elena-Bruyako (changes requested):
see comments

